PR #39, opened 2016-08-20
Fix crash in yaYUL on OSX

jimlawton:


rburkey2005:
Jim, I'm afraid you've outsmarted yourself.  Or maybe it's just me.  
I've already started to think that commits are safe because travis will 
detect all build errors, but of course that's not true, and this commit 
demonstrates that.

On my Linux box, at least, your commit actually breaks the top-level 
Makefile.  (The syntax "else if defined(...)" isn't allowed, I guess, in 
GNU make.)  I couldn't see at first why the travis build didn't catch 
it, but then I realized that the travis build must just do something 
like "make -C yaYUL ; make missions", and so perhaps the broken part of 
the top-level Makefile isn't encountered.

At any rate, I fixed the Makefile in a later commit, though I had a lot 
of problem with merging and what-not, due to unfamiliarity with git, so 
the comment attached to my commit got messed up and doesn't reflect the 
fixed makefile.

Does travis provide any unique environment variables that we could use 
in the top-level makefile to selectively disable certain targets within it?

jimlawton:
Interesting. I tested it locally in both OSX and in my Docker Linux image
and both built before I submitted it to Travis, which also built it. And
Travis is building all the non-gui stuff.

I've gradually been starting to wonder about the versions of tools you're
using? Because if I try "-Wall -Werror" it doesn't build for me at all on
Linux or Mac. Lots and lots of errors...

Gnu Make has supported the "else if defined" syntax for a very long time,
as far as I remember.

Maybe you could me know what versions of make and gcc you're using? I can
probably construct a Docker image with those versions.

jimlawton:
The versions the Docker image is using:

   - GNU Make 4.1 Built for x86_64-pc-linux-gnu
   - gcc (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609

The local OSX build is using GNU Make 3.81, which is fairly old.

I just checked the Make history, and the 'else <condition>' feature was
introduced in Make 3.81, which was released in April 2006.

rburkey2005:
So maybe it /was/ just me. :-)   Well, all of my recent builds on x86 
have been on Linux Mint 17.3 and LMDE 1 (Linux Mint Debian Edition), and 
I suppose those are pretty conservative distributions, so I do have 
older tools:

  gcc 4.8
  gmake 3.8 and 3.81

I'd never given it much thought before, but the strategy for having code 
that the most number of people could actually compile would seem to be 
for us to use the latest version of gcc (which is likely to flag the 
most problems) but a really early version of gmake (which is likely to 
have the least features and therefore bomb the easiest)?  But is it 
really worth it, I wonder, considering the relative rarity with which 
this kind of problem is likely to come up?

And apparently it's

  if defined

that gmake didn't recognize.  The else had nothing to do with it.

What happens if a travis build fails?  I assume you just get a 
notification, and travis doesn't have the necessary permissions to undo 
the commit.

jimlawton:
Gcc 4.8 is reasonably recent. Make 3.81 should work fine too. I don't know
about Make 3.8 though, is that 3.80? Because that probably does not support
else conditionals.

rburkey2005:
I think it just said 3.8, but it doesn't really matter in terms of this 
particular discussion because the computer on which I noticed and 
verified the problem was the one using 3.81.

jimlawton:
Just saw the fix you made to my complete mess of one of the strcpy fixes.
😲 Sorry!

I'm wondering though how my local builds and the Travis build managed to
build everything with that blooper in there... Perhaps I *have* outsmarted
myself!

rburkey2005:
Huh?  I thought we just decide it was my old version of gmake, and not 
anything you messed up.

Oh, I know what you mean.  I think you had added something like 
s[length]-0 and then immediately fixed it to s[length]=0.  Somehow when 
I pulled it, it changed back to s[length]-0, so I had to fix it again to 
s[length]=0.  But you had already fixed it, I think. Weird, yes, but not 
your fault.  I blame git.

rburkey2005:
You aren't talking about wxWidgets errors by any chance, are you?  I get 
a couple of hundred of those too, simply because that's what happens 
with wxWidgets, and there's nothing I can do about it that I know of.

Issue #30, opened 2015-09-14
Luminary Autopilot bug


spacex15:
when doing a v49 maneuver with a pitch angle greater than 180 deg
(v49E, V25E,+00000E,+33000E,+00000E,PRO,PRO) for example

the autopilot is executing the correct maneuver but will not maintain the attitude correctly

it slowly drifts toward zero pitch attitude (very slowly, several minutes)

telemetry is showing that variable CDUYD value is decreasing slowly towards zero instead of not moving.

code not working is in P-AXIS_RCS_AUTOPILOT
1STOTWOS	CCS	 A
		AD	ONE
		TC	Q
		CS	A
		TC	Q

this function, instead of storing the CDUYD variable without changing it , is  substracting one from the variable before returning to the caller

it seems there is an implementation problem in the MSU operand when calculating on CDU counters greater than 16384

indy91:
I have stumbled upon the same issue. Any insight in how to fix it?

rburkey2005:
I don't have any real insight into it in the short term.

My fear is that it's a low-level bug in the CPU simulator, and that I my not personally have the expertise to fix it right now.  Part of the problem is that the CPU validation software we have right now is based on _descriptions_ of tests, from a testing document that may have been obsolete with respect to Block 2.

What I'm hoping in the longer term is this:  There is a possibility that in the relatively near future we might have a listing of a program called AURORA which, I'm told, is the _last_ version of the Luminary software to actually contain the official Block 2 version of the testing code.  Once we have that, we can obviously run those tests, and may be able to find and fix specific low-level problems in the CPU.

Technically, the problem is that the behaviors of the CPU's instructions were supposed to be dependent on microcode that the AGC developers called "pulse sequences", but I find those pulse sequences very confusing and got nowhere trying to use those as a guide to implementation, instead basing the behavior on the documented behavior of the instructions.  Some people _don't_ seem to find the pulse sequences confusing, and if there were a programmer who understood the pulse sequences better than I do, it might be possible to track down problems like this prior to having AURORA in hand.

The bottom line is that while I'll still think about it, this problem may surpass my meager programming abilities until AURORA becomes available.

rburkey2005:
[...]

Issue #9, opened 2015-04-07
Add support for gdb command line options


GoogleCodeExporter:
```
Add support for gdb command line options will allow for better integration 
with 3rd party debug GUI's.

launching the emulator could take a binary directly as the default input 
instead of using the --core option in the current version.

It should be possible to launch the tool like:

yaAGC Colossu249.bin

to start debugging. And use an option like --nodebug to launch just the 
simulation without debugging.



```

Original issue reported on code.google.com by `ohommes@gmail.com` on 21 Dec 2008 at 4:52

PR #41, opened 2016-08-21
yaAGC: Fix possible bug in MSU implementation, and Validation to match

thewonderidiot:
This is a potential fix for #30, changing the behavior of the MSU instruction in certain situations (thanks to  @spacex15 for identifying this as an MSU error).

The underlying assumption of this PR is that the following check in E-2065 is incorrect: 
msu_error

Running Validation in my hardware simulator and looking [this MSU instruction](https://github.com/rburkey2005/virtualagc/blob/c56c9cdce8ab963441a333ce728d832e85a6b229/Validation/SmallyMSUCHK.agc#L45), I see the following waveform:
msu_waveform

If you match the control pulses under the highlighted MSU0 sequence line up with the timepulses at the top, you can see that the control pulses that appear during each timestep are what Hugh Blair-Smith defined in AGC4 Memo #9:
msu_memo

And the result in the **A** register at the end of T12 (vertical bar) is +1 instead of the E-2065-suggested +0.

Breaking down what's going on by timestep:
**T01**: RL10BB WS reads the lowest 10 bits of the current instruction word (ie, the address) and writes them to the S register, which controls which memory address will be read in the coming memory cycle. For us, 00001 goes into S register since we're addressing L

**T02**: RSC WG reads the contents from a flip-flop register and writes it into the memory buffer register G, if a flip-flop is being addressed. In our case it is, so the contents of L (177777) are copied into G

**T05**: RG WB copies the contents of the G register to the B register. B is special because it's the only way to invert a word -- this is done by reading the fake "C register" which is the complement of B. By this point the memory cycle is done and the contents of G reflect the contents of the memory address being read. Since we were addressing a flip flop register, no memory cycle occurred and T02 took care of populating G. So now, B has received the contents of L (177777).

**T06**: RC WY CI A2X copies the complement of B (and therefore the complement of L) to one of the inputs of the adder, copies the contents of A to the other input, and introduces a carry bit to effect a +1. A is currently +0, and the complement of L is +0, so the adder now works on calculating 0 + 0 + 1

**T07**: RUS WA TSGN copies the 15-bit sum from the adder into the A register, sign-extending bit 15 into bit 16. It then tests the sign of this value, setting branch register 1 to the value bit 16. We just got a +1 from the adder, so A is set to +1 and branch register 1 is set to 0.

**T08**: RZ WS ST2 reads the address of the next instruction from Z and writes it to the addressing register S, so the upcoming read cycle will read from that address. It also advances the current instruction stage to Stage 2, which performs instruction loading and advancement.

**T09**: RB WG takes the original operand and puts it back into the memory buffer register G. If it was read from erasable memory (which is destructive), this restores it. We read from a flip-flop, which wasn't destructive, so this doesn't matter for us.

**T10**: RA WY MONEX is only executed if branch register 1 was set to 1 during T07. For us it wasn't so these pulses don't occur. If they did, they would copy A to one adder input and -1 to the other, to do A - 1

**T11**: RUS WA copies result of T10 (and the final result of the MSU) into the A register. If T10 didn't occur, this just harmlessly gets the earlier result from T07 again. For us, +1 is again copied from the adder to A.

I'm pretty confident that the behavior described above, as seen in the simulator, is correct.

To recreate this in yaAGC, I changed the subtraction to adding the complement plus one, and then looking at the sign bits to determine whether or not the further subtraction of 1 is necessary. This more closely mirrors the hardware implementation of the instruction. The old yaAGC implementation of MSU would have considered the difference to be negative instead of positive, and gone ahead with the subtraction of 1, resulting in +0.

@indy91 or @spacex15, can one of you test this in NASSP to see if it fixes the issue? I've never used Orbiter before.


indy91:
Yes, that seems to fix the issue. No problems holding attitude at pitch angles above 180° anymore. Awesome! Any idea why there never was any issue with this in the CSM? The Digital Autopilot is way different in the LM, but I'm sure the MSU instruction is used in the CM AGCs, too...

rburkey2005:
So a White Knight _did_ come along and save me from using my brain.  Terrific!  I took the one additional step of comparing AGC4 Memo #9 to the readable plain-text version that Hugh Blair-Smith had prepared for us (https://www.ibiblio.org/apollo/Documents/AgcPulsesAndSequences.txt), and there were no corrections in it that seemed relevant.

indy91:
Yes, I was sure this problem would bother us for a little bit longer, haha.

thewonderidiot:
@indy91 Thanks a lot for testing! I can't answer your question, unfortunately, as my knowledge drops off quite abruptly above the hardware.

PR #42, opened 2016-08-22
yaAGC: Overflow-correct the L register at the end of agc_engine()

thewonderidiot:
Based on discussion at bottom of #30.

There's two pieces of evidence that overflow-correcting any value read from L is the correct behavior. First, on the last of the 4 bit module pages, there's some backplane trickery involving bits 15 and 16 of L:
l_overflow_correction

During the RL control pulse, the value stored in L15 is hard grounded, and instead the value from L16 is taken across the backplane and onto the bit 15 line of the write bus.

Smally corroborates this in his description of the D--LCHK test:
"This subroutine was  written to check that the overflow bit disappeared when a word went into and out of the L register and to make sure that the Q register was capable of holding 16  bits."

In yaAGC, per Ron's suggestion, the overflow correction is done not at read time but at the end of every call to agc_engine(). The behavior is transparent to software running on the AGC, so this difference doesn't matter.

For Validation, the LXCH checks have been updated to make sure that values read from L are overflow-corrected. All other tests that used L to temporarily store an overflow condition have been changed to use Q instead.


thewonderidiot:
You know what, this got me thinking about interrupt inhibition when overflow exists. I missed a change on [this line](https://github.com/rburkey2005/virtualagc/blob/a5fa74588c4cbe35534437eb65d5f1e246973ffc/yaAGC/agc_engine.c#L1828): 
```C
      // Handle 
      if (DebuggerInterruptMasks[0] && !State->InIsr && State->AllowInterrupt
	  && !State->ExtraCode && State->IndexValue == 0 && !State->PendFlag
	  && !Overflow && ValueOverflowed (c(RegL)) == AGC_P0
	  && ValueOverflowed (c(RegQ)) == AGC_P0 &&
	  //ProgramCounter > 060 && 
	  Instruction != 3 && Instruction != 4 && Instruction != 6) 
```
With this, L no longer needs to be checked if it stores overflow. But what about Q?

The circuit that inhibits interrupts is at the top of Inout VII, sheet 1:
ovnhrp

It operates by checking to see if A15 and A16 differ each time the NISQ (load New Instruction into the SQ register) pulse happens. The result is latched until the next NISQ (unless this happened during cycle 3 of MP). The flip flop drives the line OVNHRP, which stands for "OVerflow iNHibit inteRuPts". OVNHRP directly prevents interrupts from happening.

It, quite apparently, only checks the A register. Not only that, but Q15 and Q16 aren't even considered special enough to get their own nets; they're never accessed internally by the hardware, only put onto the write bus by request of an RQ control pulse.

We know beyond a shadow of a doubt that Q can store an overflow condition in its 16 bits, thanks to Smally. So, then, presumably that means that overflows in Q do *not* prevent interrupts. QRUPT is just a standard erasable memory location too, so overflow in Q would be lost by an interrupt routine that backs Q up to QRUPT.

So, something's not adding up for me. Maybe there were rules dictating that Q should not be used to store overflow conditions in the flight programs?

rburkey2005:
My guess is that if I thought the inhibition was necessary, it may have 
been because of the loss of the bit in QRUPT, as you say, rather than 
anything I had read in the documentation.  Unfortunately, I no longer 
recall.

Ultimately it probably doesn't make much difference, because eventually 
(quickly, most likely), the state of Q is going to change, and the 
inhibited interrupt will fire then anyway.  In the simulation, which 
isn't accurate timewise on a cycle by cycle basis, but rather gangs 
together a bunch of instructions in short bursts of time, it wouldn't be 
possible to even notice that the interrupt had fired late, unless one 
was single-stepping it through the debugger.

But you may be right about the rules.  I'm afraid that's the kind of 
thing that would be tribal knowledge rather than written down, and it 
might be an answer that could come only from an AGC developer.

rburkey2005:
It makes me feel particularly good that all of the corrections to the Validation suite were in the tests I wrote from scratch, rather than those that I think of as coming from Smally.  I'll merge this after I close this comment.

thewonderidiot:
@rburkey: Your [Assembly Language Manual](http://www.ibiblio.org/apollo/assembly_language_manual.html) page needs to be updated to reflect this one now that it's been merged. The description of the L register says it's 16 bits and can store overflow. :)

rburkey2005:
The manual has been updated.

PR #47, opened 2016-09-01
Updated handling of TIME6 and T6RUPT. Hopefully fixes #46.

### LINKED ISSUE #46: Possible incorrect TIME6 behavior.
This is related to NASSP for Orbiter.

We've been testing LM powered descent operations, and have been getting consistent 1103 alarms (unused CCS branch, causing a P00DOO) during DAP usage. The V05N08 display pinpoints the problematic CCS in the T6RUPT handler routines (fixed bank 017, address 02032), and comparing the AGC assembly listing information with the C code for handling timers, it seems possible to me that the TIME6 counter is improperly handled.

According to code comments for the Luminary 99 assembly code (page 1403), it stipulates that TIME6 should at all times be greater than +0, triggering the interrupt after it is loaded with -0, 67777 indicating counter disabled. Otherwise, the CCS instruction transfers control to the abort handler, causing the 1103.
screenshot 79
screenshot 80

EDIT: I originally thought it was based on an increasing counter/PINC, but I did not know about the 1+(-1)=-0 detail of the adder. Thanks to thewonderidiot for clearing that up.

thewonderidiot:
In theory this should make handling of TIME6 much more accurate, and fix #46. Rationale for the changes:

t6on
T6P, the counter pulse that generates a counter request on TIME6, can only occur if T6ON is set via CH13 bit 15. If not set, TIME6 cannot increment.

t6on_clear
T6ON, in turn, is cleared when T6RPT is triggered.

t6rpt
T6RPT requests are generated when a ZOUT pulse is emitted for counter 31, AKA TIME6 (the other two inputs on that gate are counter addressing)

And lastly, a quick look at the sim, on the DINC *before* ZOUT occurs and T6RUPT is generated:
dinc_cycle

The control pulses generated by the sim match what is described by Hugh Blair-Smith in his Memo#9:
```
DINC

 1. xx  RSCT   WS
 2. xx  RSC    WG
 5. xx  RG     WY     TSGN   TMZ    TPZG
 6. 00  MONEX  POUT
 6. 10  PONEX  MOUT
 6. x1         ZOUT
 7. xx  RU     WSC    WG     WOVR
 8. xx  RB     WS
```
I won't go through every step (unless requested), but the important bits are:
**T01.** RSCT WS reads the counter cells to determine which counter is being addressed. This is why the DINC line doesn't go high until T02; the computer doesn't know it's doing a DINC yet
**T05.** The current counter value is put into one input of the adder, and tested for polarity and zero-ness. We read a positive 1, so no tests pass and the branch registers remain 0.
**T06.** MONEX puts -1 into the other adder input, and generates a POUT. Note that POUT lasts longer than all the other control pulses -- that's because it's apparently timed with the CDU strobe, rather than one of the 12 fundamental timepulses.
**T07.** Reads the results of 1 + (-1) out of the adder and puts them into G. This is where the vertical cursor is; the result is -0.

According to the sim, TIME6 increments 0.3125ms after TIME1 and TIME3 do. That's smaller than VirtualAGC is even ticking the clocks right now, so I doubt the delta is going to matter for simulation purposes. It might, though, so it's worth noting.

@eddievhfan1984, can you give this a whirl and see if it fixes your program alarms?

e: By the way, I tested this on a very hacked up version of Validation with a smattering of print statements in yaAGC to verify that the behavior of yaAGC is now more or less what the sim is doing.

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -245,6 +245,11 @@
  *				resumption in Windows.
  *		07/17/16 RSB	Commented out a variable that wasn't being
  *				used but was generating compiler warnings.
+ *		08/31/16 MAS	Corrected implementation of DINC and TIME6.
+ *				DINC now uses AddSP16() to do its math. T6RUPT
+ *				can only be triggered by a ZOUT from DINC. TIME6
+ *				only counts when enabled, and is disabled upon
+ *				triggering T6RUPT.
  *
  * The technical documentation for the Apollo Guidance & Navigation (G&N) system,
  * or more particularly for the Apollo Guidance Computer (AGC) may be found at
@@ -1023,63 +1028,36 @@ CounterMCDU (int16_t * Counter)
 int
 CounterDINC (agc_t *State, int CounterNum, int16_t * Counter)
 {
-  //static int CountTIME6 = 0;
   int RetVal = 0;
-  //int IsTIME6, FlushTIME6;
   int16_t i;
-  //IsTIME6 = (Counter == &c (RegTIME6));
-  //FlushTIME6 = 0;
   i = *Counter;
   if (i == AGC_P0 || i == AGC_M0)	// Zero?
     {
       // Emit a ZOUT.
       if (CounterNum != 0)
 	ChannelOutput (State, 0x80 | CounterNum, 017);
+
+      RetVal = 1;
     }
   else if (040000 & i)			// Negative?
     {
-      i++;
-      if (i == AGC_M0)
-	{
-	  RetVal = -1;
-	  //if (IsTIME6)
-	  //  FlushTIME6 = 1;
-	}
-      //else if (IsTIME6)
-      //  {
-      //    CountTIME6--;
-      //    if (CountTIME6 <= -160)
-      //      FlushTIME6 = 1;
-      //  }
+      i = AddSP16(SignExtend(i), SignExtend(AGC_P1)) & 077777;
+
       // Emit a MOUT.
       if (CounterNum != 0)
 	ChannelOutput (State, 0x80 | CounterNum, 016);
     }
   else					// Positive?
     {
-      i--;
-      if (i == AGC_P0)
-	{
-	  RetVal = 1;
-	  //if (IsTIME6)
-	  //  FlushTIME6 = 1;
-	}
-      //else if (IsTIME6)
-      //  {
-      //    CountTIME6++;
-      //    if (CountTIME6 >= 160)
-      //      FlushTIME6 = 1;
-      //  }
+      i = AddSP16(SignExtend(i), SignExtend(AGC_M1)) & 077777;
+
       // Emit a POUT.
       if (CounterNum != 0)
 	ChannelOutput (State, 0x80 | CounterNum, 015);
     }
+
   *Counter = i;
-  //if (FlushTIME6 && CountTIME6)
-  //  {
-  //    ChannelOutput (State, 0165, CountTIME6);
-  //    CountTIME6 = 0;
-  //  }
+
   return (RetVal);
 }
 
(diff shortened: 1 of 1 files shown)


rburkey2005:
Thanks, Mike! I'm committing it right now, in the expectation that it will be okay.  I'm surprised to see that I had at least put in a note saying that I didn't know whether TIME6 was supposed to fire on an interrupt or not, because frankly it all looks like gobbledy-gook to me right now!  :-)

indy91:
We are currently trying this update. The 1103 alarms are gone and there are no more P00DOO in P63 or at other times when the DAP is used. The LGC is still not too happy, but that seems to be mostly some problems with the landing radar and other problems not caused by the Virtual AGC. If we think that any of the still occuring program alarms (1210, 401...) could be caused by the Virtual AGC, then we will let you know. Thanks for this update though, it works great! Even if the landing doesn't work yet, at least we can fly an P70 or P71 abort now without the 1103 alarms and P00DOO. :)

rburkey2005:
Speaking of aborts, I'm curious ... does the LM sim use our AEA 
simulator as well as the AGC simulator?

indy91:
The AEA isn't running yet, but the AGS flight programs are already distributed with NASSP and there is a simulated and working DEDA. All it would take to at least have the AEA run is a few lines of code so that it does its cycles per second and some basic I/O with the DEDA. Adding the components of the Abort Sensor Assembly could then be done step by step. I am more of an astrodynamics and orbital mechanics specialist, so even this basic I/O is a little bit too complicated for me. But the LM is ready to at least have the AEA start up and I am very interested in working out the prelaunch erasable load for the various missions. Luckily it seems to be flexible enough to support an Earth orbital mission like Apollo 9 even with the Apollo 11 AGS version. Hopefully someone more qualified than me does this initial work on the AGS soon so I can get to work and make it operational.

rburkey2005:
I'm kind of sorry I asked, since I'm sure it will have lots of the kind 
of issues you guys have experienced with the AGC, but without Mike being 
in quite as advantageous position for me to exploit him to route out all 
the problems for me. Fortunately, it's a much simpler system than the 
AGC, and so we may be luckier with it.  Thanks for the info!  -- Ron

indy91:
We really only had any problem that could be caused by the AGC engine since we started to work on the LM again. The CMC has always worked great. Well, within the (luckily only a few) limitations of Orbiter. 

And we are of course in the unique position to find all these new issues! I would doubt anyone else got P63 of the LGC to work with the Virtual AGC before. The ascent is already working great: https://www.youtube.com/watch?v=bT7JGyOae6U

thewonderidiot:
> We are currently trying this update. The 1103 alarms are gone and there are no more P00DOO in P63 or at other times when the DAP is used. The LGC is still not too happy, but that seems to be mostly some problems with the landing radar and other problems not caused by the Virtual AGC. If we think that any of the still occuring program alarms (1210, 401...) could be caused by the Virtual AGC, then we will let you know. Thanks for this update though, it works great! Even if the landing doesn't work yet, at least we can fly an P70 or P71 abort now without the 1103 alarms and P00DOO. :)

Excellent! I'll probe a bit into those alarms on our side too to see if there's anything that immediately stands out to me. And yeah, you guys are awesome for testing -- for both VirtualAGC and my hardware sim as I research into the problems! I'd never actually switched my TIME6 timer on before yesterday. And there's not a bat's chance in hell I could get any of these programs actually running on anything yet. :)

> ...but without Mike being in quite as advantageous position...

Maaaaan, "quite" is an understatement. I've personally digitized all the AGC schematics.... but as far as the AGS goes, well, I know what "AGS" stands for! Hahaha.

rburkey2005:
Perhaps I understated, but it's dangerous to say what other people can't do if they put their mind to it. :-)
--Ron


indy91:
While this change fixed #46 for the LGC, it seems to have done some bad things to the CMC. It now permanently has a RESTART and STBY light on. I had to manually copy over the changes to the agc_engine.c to our modified NASSP version of it, but I think I've done it correctly. Were there any other changes to the file in the last few weeks that we are missing and #47 is depending on? It's of course possible that the NASSP modifications are the cause of this, but if you have any quick ideas...

PR #48, opened 2016-09-05
#46: Correct clearing of CH13's TIME6 enable bit on T6RUPT

thewonderidiot:
Fixes a problem with #46. Channel 13 should be written with the masked contents of channel 13, *not* of TIME6...

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -1639,7 +1639,7 @@ agc_engine (agc_t * State)
             {
 	      State->InterruptRequests[1] = 1;
               // Triggering a T6RUPT disables T6 by clearing the CH13 bit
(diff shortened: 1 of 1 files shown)


rburkey2005:
@indy91 Should I conclude that this is fixed now, or that you still need some testing time on both the LGC and CMC?  I'm sure it's okay, but then I was sure of that last time too, so I'm not going to merge it until the last t is crossed.  :-)

indy91:
Yes, #48 solved all the issues with the CMC and LGC that would caused by writing bad numbers in output channel 13.

rburkey2005:
Super!  Thanks all!

thewonderidiot:
Yup, sorry about the bug!

PR #49, opened 2016-09-07
Change TIME5 to be phased 5ms after TIME3, and TIME4 to be phased 7.5ms after TIME3

thewonderidiot:
The NASSP guys are still having program alarms from multiple things attempting to use the radars at once. I took a look at timer phasing to see if things might be running at the wrong relative times, and discovered that TIME4 and TIME5 both have the wrong phasing.

First, I switched from PINCing SCALER1 and SCALER2 to just straight incrementing them. Those channels are direct views into the [flip-flops in the center of every clock divider in the Scaler module](http://klabs.org/history/ech/agc_schematics/logic/a01-1.jpg), and don't need CPU time to change states.

For timer phasing, counter increment requests are generated by the following circuit in Inout IV, page 2:
counter_timing

TIME1 and TIME3 both increment every F10B pulse, whereas TIME5 increments every F10A pulse. F10 is the 100Hz frequency, and the A and B pulses are separated by half of the frequency's period (ie 5ms).

TIME4, on the other hand, is triggered by F9B, the B pulse of the 200Hz frequency. The A and B pulses of each frequency align in time with the A pulse of the frequency above it; so F10A and F10B are both coincident with every other F9A. F9B, being offset by 2.5ms from F9A (half the 5ms period), is therefore offset by 2.5ms from every F10 pulse (and therefore from T1 and T3 incrementing). In order to keep TIME4 from going double speed, the F9B is also filtered based on FS10, the square wave version of F10. In picture form:
timer_wave

The TIME4 offset is particularly interesting, because Savage & Drake pretty clearly state in E-2052:

> Time 3 and TIME 4 are phased to be 5 ms apart in pulsing. As long as RUPT processing does not exceed 4 ms, they will not interfere with each other. 

And seemingly just about all the other documentation out there that refers to timer phasing repeats that. However, one of the tests that Smally describes in E-2065 is this:

> The only non-programmable instruction that is checked is  PINC, which is checked in the RUPTCHK subroutine. The fact that TIME3 interrupts 2 1/2 milliseconds after TIME4 assures the proper functioning of all the pulses in this instruction. It is not possible to check the pulses in the other non-programmable instructions.

So I'm pretty sure the behavior described above is right, despite what all the rest of the literature says!

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -1592,13 +1592,16 @@ agc_engine (agc_t * State)
   // This can only iterate once, but I use 'while' just in case.
   while (ScalerCounter >= SCALER_OVERFLOW)
     {
-      // First, update SCALER1 and SCALER2.
+      // First, update SCALER1 and SCALER2. These are direct views into
+      // the clock dividers in the Scaler module, and so don't take CPU
+      // time to 'increment'
       ScalerCounter -= SCALER_OVERFLOW;
-      if (CounterPINC (&State->InputChannel[ChanSCALER1]))
-	{
-	  State->ExtraDelay++;
-	  CounterPINC (&State->InputChannel[ChanSCALER2]);
-	}
+      State->InputChannel[ChanSCALER1]++;
+      if (State->InputChannel[ChanSCALER1] == 037777)
+        {
+          State->InputChannel[ChanSCALER1] = 0;
+          State->InputChannel[ChanSCALER2] = (State->InputChannel[ChanSCALER2] + 1) & 037777;
+        }
       // Check whether there was a pulse into bit 5 of SCALER1.
       // If so, the 10 ms. timers TIME1 and TIME3 are updated.
       // Recall that the registers are in AGC integer format,
@@ -1614,15 +1617,16 @@ agc_engine (agc_t * State)
 	  State->ExtraDelay++;
 	  if (CounterPINC (&c(RegTIME3)))
 	    State->InterruptRequests[3] = 1;
-	  // I have very little data about what TIME5 is supposed to do.
-	  // From the table on p. 1-64 of Savage & Drake, I assume
-	  // it works just like TIME3.
+	}
+      // TIME5 is the same as TIME3, but 5 ms. out of phase.
+      if (010 == (017 & State->InputChannel[ChanSCALER1]))
+	{
 	  State->ExtraDelay++;
 	  if (CounterPINC (&c(RegTIME5)))
 	    State->InterruptRequests[2] = 1;
 	}
-      // TIME4 is the same as TIME3, but 5 ms. out of phase.
-      if (010 == (017 & State->InputChannel[ChanSCALER1]))
+      // TIME4 is the same as TIME3, but 7.5ms out of phase
+      if (014 == (017 & State->InputChannel[ChanSCALER1]))
 	{
 	  State->ExtraDelay++;
 	  if (CounterPINC (&c(RegTIME4)))


rburkey2005:
Cool!  But just as a matter of policy, it has occurred to me that for problems originating from NASSP, it probably makes sense to always let them check the fixes before merging them.  Does that sound reasonable?  Of course, by making it a "policy", that probably insures this will be the last problem they encounter anyway.

thewonderidiot:
Absolutely, and totally agreed. I think in this case though the best we can do is make sure it doesn't _break_ anything for them, since this came up from my own diggings, and we have absolutely no idea whether or not it will fix the alarms they're getting. :)

rburkey2005:
Exactly.  Even if you're absolutely right, it's best to make sure that being right doesn't break anything.  By the way, in a weird coincidence, I happen to be working on the Block 1 simulator right now, and am in the fortunate position of having a different Block 1 simulator written many years ago by somebody else to compare mine to ... and one of the first problems I discovered is that all the timers in that old sim run 10X as fast as they're supposed to.  Either that, or else all the Block 1 documentation is wrong.  :-)

thewonderidiot:
Ha, interesting! That's a pretty major difference. I'm a bit surprised he didn't run into trouble with ISRs taking too long with that. I guess he wrote all of his own AGC code based on descriptions, though, didn't he?

It's been a long while since I've spent quality time with R-393, but I did actually refer to its descriptions of the Scaler frequencies in the table at the end to make sure they aligned with the hardware sim results while trying to figure out why the hell my TIME4 was "wrong". So I don't think the Scaler module itself, at least, changed much from Block 1 to Block 2. Does Block 1 have the same suite of timers?

indy91:
Doesn't seem to break anything, doesn't seem to fix anything. Still the 1210 alarm when the LGC is first looking for landing radar data (I guess) and the DSKY blanks about once per guidance cycle. Late in P63 the frequency of the DSKY blanks increases. Then in early P64 it first doesn't have new program alarms (probably while the LGC waits for the LR to move to position 2) and then enough program alarms and DSKY blanks that you barely get to see the registers. This happens in all landing programs, P64-P67.

rburkey2005:
On 09/06/2016 10:09 PM, Mike Stewart wrote:
>
> Ha, interesting! That's a pretty major difference. I'm a bit surprised 
> he didn't run into trouble with ISRs taking too long with that. I 
> guess he wrote all of his own AGC code based on descriptions, though, 
> didn't he?
>
Yes.  It's a command-pulse simulator, as opposed to a simulator at the 
level of the assembly-language instructions.  It was written as a 
proving-ground for the block 1 hardware sim he was developing, and there 
was nothing like a simulated DSKY.  R-393 has a number of discrepancies 
with reality that (from my perspective) ended up biting him, but he 
produced an amazingly accurate simulation in most respects.  Frankly, I 
suspect that he cranked the timer rate up just to make the interrupts 
happen sooner during debugging, and then forget to slow them down again 
afterward, since he wasn't really interested in the software sim as such.
>
> It's been a long while since I've spent quality time with R-393, but I 
> did actually refer to its descriptions of the Scaler frequencies in 
> the table at the end to make sure they aligned with the hardware sim 
> results while trying to figure out why the hell my TIME4 was "wrong". 
> So I don't think the Scaler module itself, at least, changed much from 
> Block 1 to Block 2. Does Block 1 have the same suite of timers?
>
TIME1-TIME4.  1, 3, and 4 purportedly clock at 10 ms intervals, while 
TIME2 clocks from the overflow in TIME1.  TIME3 and 4 have their own 
rupts.  So according to the way I remember that I /thought/ they were 
supposed to work in block 2, they seem more-or-less the same.

thewonderidiot:
Darn! Oh well, shall we go ahead and merge this then, since nothing is broken?

rburkey2005:
Okay.  Done.

rburkey2005:
I opened a new issue for the original problem.

thewonderidiot:
Thanks Ron!

PR #53, opened 2016-09-09
yaAGC: Added a special case to DV (Dividend = 0, Divisor != 0)

### LINKED ISSUE #52: Program alarm error from LGC during descent and ascent.
There is still a program alarm happening during descent and also ascent. It's a 401 and doesn't seem to do any harm and other than the 1210 alarm it also doesn't lock you out of the DSKY. Here the alarm data:

V05 N08
R1: 3757
R2: 60066

No attitude in gimbal lock is commanded at all, so a 401 alarm really shouldn't happen. This seems more like a Virtual AGC problem than the 1210 alarm #50 , which was pretty much 99% a NASSP problem.

thewonderidiot:
This changes agc_engine to behave similarly to the hardware sim in this particular case of divide. This has the potential to fix #52, as in isolated test cases, a requested 77777 gimbal angle triggers 401 alarms in yaAGC, but not in the hardware sim.

I've got no detailed analysis to suggest that this should be the case this time around... just the statement that this is the apparent current behavior of the hardware sim (which very well could be wrong. I don't trust my DV anywhere near enough to say with certainty that it is wired correctly). The divide case makes plentiful use of WHOMP, which isn't documented in AGC4 Memo *#*9 and makes control pulse analysis even more of a pain than in non-WHOMP-using DVs.

Somewhat even more frighteningly, this PR contains no changes to Validation, because no changes to Validation were required.... this results in a measurable behavior change for yaAGC but Validation can't tell the difference. Even though the DV checks are constructed straight from Smally. Yikes.

@indy91 or @eddievhfan1984, would you mind trying this out to see if the 401s are impacted at all? Thanks!

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -250,6 +250,10 @@
  *				can only be triggered by a ZOUT from DINC. TIME6
  *				only counts when enabled, and is disabled upon
  *				triggering T6RUPT.
+ *		09/08/16 MAS	Added a special case for DV -- when the dividend
+ *				is 0 and the divisor is not, the quotient and
+ *				remainder are both 0 with the sign matching the
+ *				dividend.
  *
  * The technical documentation for the Apollo Guidance & Navigation (G&N) system,
  * or more particularly for the Apollo Guidance Computer (AGC) may be found at
@@ -2464,6 +2468,21 @@ agc_engine (agc_t * State)
 		  c (RegL) = SignExtend (*WhereWord);
 		}
 	      c (RegA) = SignExtend (Operand16);
+	    }
+	  else if (AbsA == 0 && AbsL == 0 && AbsK != 0)
+	    {
+	      // The divisor is 0 but the dividend is not. The quotient and
+	      // the remainder both receive 0 with the sign matching the dividend
+	      if (Dividend == 0)
+	        {
+	          c (RegA) = AGC_P0;
+	          c (RegL) = AGC_P0;
+	        }
+	      else
+	        {
+	          c (RegA) = SignExtend (AGC_M0);
+	          c (RegL) = SignExtend (AGC_M0);
+	        }
 	    }
 	  else
 	    {
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -2471,7 +2471,7 @@ agc_engine (agc_t * State)
 	    }
 	  else if (AbsA == 0 && AbsL == 0 && AbsK != 0)
 	    {
-	      // The divisor is 0 but the dividend is not. The quotient and
+	      // The dividend is 0 but the divisor is not. The quotient and
 	      // the remainder both receive 0 with the sign matching the dividend
 	      if (Dividend == 0)
 	        {


indy91:
I did a bunch of ascents and descents, no more 401 alarms. Hopefully this did the trick, 2 descents and 3 ascents should be enough to test a program alarm that doesn't happen each time.

rburkey2005:
It kind of makes sense to have DV return a goofy value in the case of an illegal divide, since there's no other built-in facility to trap an error like that.  On the other hand, I'm tempted to say that we should test the fix more, by having yaAGC log some kind of message (containing the program counter) every time a divide-by-zero occurs, and then ask for some more tests on Colossus and Luminary to see where these occur.  If we examined the logs and the divide-by-zero occurs _only_ at the point triggering the 401, then the change is safe.

On the other hand ... there's no actual guarantee that a log like that would find all possible divide-by-zero errors anyway, so maybe it wouldn't really prove the fix was safe after all.

On the gripping hand (:-)), am I just being too paranoid?  Should I just merge the pull request and forget about it?  Opinions?

indy91:
Well, last time a yaAGC change overwrote an output channel with nonsensical values, so, more testing is always good. ;)

thewonderidiot:
@rburkey2005 , it's the other way around, this is the zero-divided-by-something case. Which is probably a lot more common than division by 0. :)

I'd be okay with more testing. Although sleeping on it and the fact that the 401s are gone really makes me think this is right. It's the only way that I can see that LIMITSUB will behave correctly. That routine limits the input to -14344..14344 (in this case) preserving magnitude of the input. Without this special division case, the main divide logic just loses track of the dividend's sign.

In lieu of testing this more, we could also just try asking Hugh about this case. He wrote LIMITSUB.

thewonderidiot:
@indy91, to get the logging Ron suggested, put this:
```
fprintf(State->out_file, "*** %02o,%05o Division of %05o,%05o by %05o\n", CurrentFB, ProgramCounter, OverflowCorrected(Accumulator) & 077777, c(RegL) & 077777, *WhereWord);
```
on the line immediately following the comment
```
// The divisor is 0 but the dividend is not. The quotient and
// the remainder both receive 0 with the sign matching the dividend
```
It'll go out to your usual AGC log file.

indy91:
I'll try this during P12 and look how often it occurs.

rburkey2005:
Hugh explains a lot about what happens with double precision values when 
the signs differ in R-393.  I wouldn't be surprised if he covered this 
case, but since I simply cut and pasted the MP and DV stuff from yaAGC 
into the Block 1 sim, I didn't actually read any of the theory.

thewonderidiot:
Hmm.... I hadn't looked at R-393, since DV underwent a lot of change between Block 1 and Block 2. It looks like it specifies that for Block 1, LP gets not the remainder but some number of unspecified magnitude that matches the sign of the quotient? While Q gets the remainder, which is always negative. Does that sound right? I'm concerned that they're internally so different that while in the normal sense they're probably numerically equivalent, in these corner cases they might not always get the same answers.

indy91:
Here is the log file for a complete ascent, only logging the relevant parameters. It's very long, so, have fun.

[ProjectApollo AGC.txt](https://github.com/rburkey2005/virtualagc/files/464394/ProjectApollo.AGC.txt)

thewonderidiot:
Thanks @indy91! Well, it looks like at least all the divisions of 00000,77777 by something are by our friend LIMITSUB. The regular divisions of 77777,77777 happen all over the place, and those get different numerical results now too. I'll look at more cases of it tonight.

thewonderidiot:
Also noticed that I flubbed that comment, and fixed it.

rburkey2005:
I don't know.  As I say, I haven't read it yet ... however, to the 
extent I've tested it so far, the Block 2 MP from yaAGC matches what 
John says the Block 1 control pulses are producing, so I'm certainly 
hoping that DV does so as well.

thewonderidiot:
I've had a thought that should help us build confidence in this in changes like it once we have a real version of self-check. If the tests are sufficient to prove the circuitry is behaving as expected (at least to the developers' satisfaction), then we can build confidence in the hardware sim once it passes, and safely assume that its outputs in all cases, including corner cases like this, are correct. We can then use those as a source of truth to fill out these cases in yaAGC. Just gotta get that self-check code. :)

rburkey2005:
> I've had a thought that should help us build confidence in this in changes like it once we have a real version of self-check. ...

As I'm sure you know, Jim has set it up so that commits to the main branch actually test the buildability of all the programs, like yaAGC and yaYUL, as well as the correctness of yaYUL's assembly of all the AGC source files, and if the test fails the merge is rejected.  There's no reason that the validity check on yaAGC you mention couldn't be added to that.  (Actually, it's just a matter of adding it to the Makefile normally, since all the test does is to run the Makefile.)

Of course, as you say, having that code would be a good start.

thewonderidiot:
Automating that check is a great idea! We'd have to figure out the best way to do it... I'll have to think on it.

So, there's exactly three other places throughout the ascent that divided -0 by some number (always in the form 77777,77777). They are:

* [16,2306 in the DIVIDER subroutine of P-AXIS RCS AUTOPILOT](http://www.ibiblio.org/apollo/listings/Luminary099/P-AXIS_RCS_AUTOPILOT.agc.html#44495649444552). It's unclear to me if any direct behavioral differences happen here -- it looks like if anything the numbers calculated will just be 1 different.
* [30,3465 in CDUWXFR of FINDCDUW--GUIDAP INTERFACE](http://www.ibiblio.org/apollo/listings/Luminary099/FINDCDUW--GUIDAP_INTERFACE.agc.html#43445557584652). This one passes its remainder immediately to LIMITSUB. That makes me think it's probably fine here for the same reason that I think LIMITSUB is fine.
* [30,3351 a ways after DELGMBLP of FINDCDUW--GUIDAP INTERFACE](http://www.ibiblio.org/apollo/listings/Luminary099/FINDCDUW--GUIDAP_INTERFACE.agc.html#44454C474D424C50). All the math following this results in -0 instead of +0, which then it finds its way into LIMITSUB once again.

So... that's what I've got. What say you, Ron?

rburkey2005:
My completely-wimpy cop-out is that that's an excellent guide for where @eddievhfan1984 or @indy91 need to concentrate their testing.  :-)

Though I gather that if you saw this in the logs, it could only be because they already had gone through those areas and found nothing bogus.  Which reveals that the whole logging idea was silly, because the log couldn't possibly show any analogous behavior.  QED.

In the absence of any other ideas, I've gone ahead and merged the pull request.

thewonderidiot:
Cool, thanks! If I find myself needing to kill time this weekend, I might do a control pulse analysis just to be sure. I've been needing to study WHOMP anyways.

thewonderidiot:
Alright, so I did a control pulse analysis of the 00000,77777 case.

Starting Conditions
-------------------
A = 000000
L = 177777
K = 14344
WHOMP = 0

Ground rules of WHOMP
----------------------
WHOMP is set by CLXC.
WHOMP is cleared at the start of the following timepulses:
	DV1: T01
	DV3: T04 T07 T10 T01
	DV7: T04 T07 T10 T01
	DV6: T04 T07 T10 T01
	DV4: T04
It's also cleared by GOJAM and NISQ.
	
From R-700:
One of the requirements for the control pulse sequence
for Block II division is for the removal of carry propagations in the adder. The
carry anticipation circuits do not expedite this action, and logic is therefore furnished
to suppress every fourth existing carry at the appropriate time, so that this action
can be accomplished within one pulse time as can addition.

I believe WHOMP to be the logic described here.

DV0
----
1. RA WB TSGN TMZ
	B = 000000
	TSGN: A is positive so BR1 = 0
	TMZ: A is not -0, so BR2 = 0
2. 0x RC WA TMZ DVST
    A = 177777
	TMZ: A is now -0 so BR2 = 1
	DVST: Next divide stage calculated
3. RU WB STAGE
	B = Z, by total happenstance. Nothing has been put into the adder, and Z was the last thing calculated
	STAGE: Stage advanced to DV1

DV1
----
4. x1 RL WB TSGN
	B = 177777
	TSGN: L is negative, so BR1 = 1
5. 1x RC WY B15X Z16
	Y = 000000
	X = 040000
	Z = 1xxxxx  The final sign is stashed here, where it can't hurt anything and nothing else will touch it
6. RU WL TOV
	L = 040000
	TOV: detects a + overflow in L, so BR1,BR2 = 01
7. RG RSC WB TSGN
	B = 014344
	RSC does nothing, we're hitting erasable mem.
	TSGN: B is positive, so BR1 = 0
8. x1 RA WY
	X = 000000
	Y = 177777
9. 0x RB WA
	A = 014344
10. RU WB
	B = 177777
11. RL WYD
	X = 000000
	Y = 000000 Only bit 15 of L is set, and WYD makes no use of WL15
12. RU WL
	L = 000000
1. L2GD RB WYD A2X PIFL
	G = 000000 MCRO is a multiply thing, so remains unset throughout the divide
	Y = 177777 PIFL is active but L is 0, so Y1 = WL16
	X = 014344
2. 0x RG WL TSGU DVST CLXC
	L = 000000
	TSGU: Sum (014343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
	DVST: Next stage is calculated
3. RU WB STAGE
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
	STAGE: Stage is advanced to DV3

DV3
----
4. L2GD RB WYD A2X PIFL
	G = 000000
	Y = 177777 L is still 0, so PIFL doesn't prevent Y1 from being set
	X = 014344
	WHOMP = 0
5. 0x RG WL TSGU CLXC
	L = 000000
	TSGU: Sum (14343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
6. RU WB
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
7. L2GD RB WYD A2X PIFL
	G = 000000
	Y = 177777 L is still 0, so PIFL doesn't prevent Y1 from being set
	X = 014344
	WHOMP = 0
8. 0x RG WL TSGU CLXC
	L = 000000
	TSGU: Sum (14343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
9. RU WB
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
10. L2GD RB WYD A2X PIFL
	G = 000000
	Y = 177777 L is still 0, so PIFL doesn't prevent Y1 from being set
	X = 014344
	WHOMP = 0
11. 0x RG WL TSGU CLXC
	L = 000000
	TSGU: Sum (14343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
12. RU WB
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
1. L2GD RB WYD A2X PIFL
	G = 000000
	Y = 177777 L is still 0, so PIFL doesn't prevent Y1 from being set
	X = 014344
	WHOMP = 0
2. 0x RG WL TSGU DVST CLXC
	L = 000000
	TSGU: Sum (14343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
	DVST: Next stage is calculated
3. RU WB STAGE
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
	STAGE: Stage is advanced to DV7

------------------------------------------------------------------------------------------
DV7 and DV6 are the same as DV3, as pointed out by Hugh in his readable version of Memo#9.
It's also readily apparent that DV3 did absolutely nothing, so I'm going to skip both of
them. The control pulses from the sim for these two also exactly match DV3. The register
values going into DV4 exactly match those going into DV7.
-------------------------------------------------------------------------------------------

DV4
----
3. RU WB STAGE
	Same timepulse as at the end of DV6. I don't know why it's listed here as well.
4. L2GD RB WYD A2X PIFL
	G = 000000
	Y = 177777 L is still 0, so PIFL doesn't prevent Y1 from being set
	X = 014344
	WHOMP = 0
5. 0x RG WB WA TSGU CLXC
	B = 000000
	A = 000000
	TSGU: Sum (14343) is positive, so BR1 = 0
	CLXC: BR1 = 0 from TSGU, so X is cleared. Additionally, WHOMP = 1.
	X = 000000
6. RZ TOV
	TOV: Z16 is 1, Z15 is 0. Negative overflow, so BR1,BR2 = 10
7. 1x RC WA
	A = 177777
8. RZ WS ST2 TSGN RSTSTG
	S = next instruction addr. It's not wide enough to be bothered by Z16 being 1.
	ST2: Next intruction sequence will be STD2
	TSGN: Z16 is 1 still, so BR1 = 1
9. RU WB WL
	B = 177777 X was zero, so carry suppression from WHOMP doesn't matter.
	L = 177777 ^^

So at the end of all that, WHOMP didn't actually matter for any additions, and we get the results
A = 177777
L = 177777

*Now* I'm confident in these results.

Here's the complete waveform for all of DV:
div_study

jimlawton:
Mike, you write the best bug reports I have ever seen!

thewonderidiot:
Thanks! :)

rburkey2005:
He probably knows that I'm too lazy to think about the issues, so I'll 
be impressed enough by the obvious care that went into it to just 
automatically merge the pull requests. :-)

-- Ron

PR #54, opened 2016-09-11
Correct (again) signs when dividing 0 by a nonzero number

thewonderidiot:
Welp... the control pulse study I did for #53 ended up being useful. Because despite showing the changes were correct in all the cases studied on that PR... it made me realize that they were wrong in some others that I hadn't looked at. Namely, when the dividend is 77777,00000, and when the divisor is negative.

Since I'm now pretty confident in the sim's correctness when dividing 0 by things, I put together a little table of all combinations of +0s, -0s, and divisors:

00000,00000 / p = 00000,00000
77777,00000 / p = 00000,00000
00000,77777 / p = 77777,77777
77777,77777 / p = 77777,77777

00000,00000 / n = 77777,00000
77777,00000 / n = 77777,00000
00000,77777 / n = 00000,77777
77777,77777 / n = 00000,77777

There's two things to take away from that:
1. L never changes, so I should just leave it alone
2. The sign of A follows standard rules. It's positive if the signs of L and the divisor match, and negative if they don't.

This changes the behavior to match those two points. @indy91, you're probably going to want to pull this one over.

Also, this has annoyed me enough that I'm going to put together Validation cases for all eight of the combinations listed above. I have a feeling that even after we have a complete self-check, Validation will remain useful as a bastion for corner cases and other differences not directly caught by it.

### DIFF
--- a/Validation/ValidateDV.agc
+++ b/Validation/ValidateDV.agc
@@ -20,6 +20,136 @@
 # Purpose:	This is the part of the Validation program that validates
 #		just the DV instruction.
 # Mod history:	07/05/04 RSB.	Began.
+# 		09/11/16 MAS.	Added eight checks for 0 divided by a number
 		
 		INCR	ERRNUM
 		
+		# Check some corner corner cases discovered by NASSP
+
+		INCR	ERRSUB		#1
+		CA	TWO
+		TS	TEMPI
+		CA	ZEROES		# Check +0+0 / +2 = +0, +0
+		TS	L
+		EXTEND
+		DV	TEMPI
+		CCS	A		# Check that A is +0
+		TCF	DVERR
+		TCF	DV1LCHK
+		TCF	DVERR
+		TCF	DVERR
+DV1LCHK		CCS	L 		# Check that L is +0
+		TCF	DVERR
+		TCF	DV2
+		TCF	DVERR
+		TCF	DVERR
+
+DV2		INCR	ERRSUB		#2
+		CA	NEGZERO		# Check that -0+0 / +2 = +0, +0
+		EXTEND
+		DV	TEMPI
+		CCS	A		# Check that A is +0
+		TCF	DVERR
+		TCF	DV2LCHK
+		TCF	DVERR
+		TCF	DVERR
+DV2LCHK		CCS	L 		# Check that L is +0
+		TCF	DVERR
+		TCF	DV3
+		TCF	DVERR
+		TCF	DVERR
+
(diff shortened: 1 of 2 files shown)


rburkey2005:
It's stuff like this that makes me hate 1's complement, which sounds so 
appealing in the abstract sense. I'm barely recovering from learning 
that +x + -x = -0, which at least makes some degree of sense.  -- Ron

indy91:
Looks good. The 401 alarms have not returned. So far I have noticed no difference.

rburkey2005:
On 09/11/2016 04:31 AM, indy91 wrote:
>
> Looks good. The 401 alarms have not returned. So far I have noticed no 
> difference.
>
>
Frankly, it seems like more of a mystery to me as to why it should ever 
have made a difference even in the cases where you did notice them.  Too 
much tribal knowledge among the original AGC developers, I suppose.   Or 
much more willingness on their part than on mine to dig into the CPU's 
schematics to figure out how the software was supposed to work. :-)

indy91:
Looking from a not-so-low level angle at the 401 alarms, I'm pretty sure they were happening in this part of the routine calculating the commanded gimbal angles (GSOP 5.3-170): 
401alarm

The limiting function did something so that Q and CDUZ were not identical anymore (like 00000 != 77777 or so) and then the 401 alarm was triggered. Does that make sense?

rburkey2005:
On 09/11/2016 05:50 AM, indy91 wrote:
>
> Looking from a not-so-low level angle at the 401 alarms, I'm pretty 
> sure they were happening in this part of the routine calculating the 
> commanded gimbal angles (GSOP 5.3-170):
> 401alarm 
> <https://cloud.githubusercontent.com/assets/13571607/18416773/d5db17dc-781d-11e6-95c0-dec45a478657.png>
>
> The limiting function did something so that Q and CDUZ were not 
> identical anymore (like 00000 != 77777 or so) and then the 401 alarm 
> was triggered. Does that make sense?
>
>
If you're responding to my comment, then sure.  I wasn't saying the 
issue or the explanation or the fix were wrong, though, just that the 
original developers were in error not to look for that condition.  
Although if the check involves a counter, perhaps not. It's still a 
pain, though. :-)

thewonderidiot:
Yeah, no kidding... I still question my sanity when operating around 0. And I'm glad to hear there's no apparent changes in NASSP -- out of that whole log file there were only 2 or 3 divisions that would have been affected.

Anyways, here's all the validation checks. Both the hardware sim and yaAGC pass all of them.

Issue #26, opened 2015-04-07
Inconsistent cycle count


GoogleCodeExporter:
```
Experimenting with AGC API I've bumped into a strange phenomenon. Running the 
agc_engine() step by step, and printing the registers before calling the 
function, I got the following result:
PendDelay:0 A:0 L:0 Z: 4000 BB:0
PendDelay:0 A:0 L:0 Z: 4001 BB:0
PendDelay:1 A:0 L:0 Z: 4001 BB:0
PendDelay:0 A:0 L:0 Z: 4001 BB:0
PendDelay:0 A:12063 L:0 Z: 4002 BB:0
PendDelay:1 A:12063 L:0 Z: 4002 BB:0
PendDelay:0 A:12063 L:0 Z: 4002 BB:0
PendDelay:0 A:0 L:0 Z: 4003 BB:12003
PendDelay:0 A:0 L:0 Z: 2563 BB:12003

The ROM image was Artemis072, so the instructions are the following:

INHINT                                         
CAF      GOBB                                  
XCH      BBANK 
TCF      GOPROG
...

The INHINT and TCF consumes 1 machine cycle correctly, but the XCH and CAF 
consumes 3 machine cycles which contradicts to the language manual. It says 
that both of these two instructions needs 2 MCT to finish. I changed line 
agc_engince.c:1891   
    State->PendDelay = i; 
to
    State->PendDelay = i-1;
with the intention to decrease the machine cycle by 1 for every multi-MCT 
instuctions. This modification doesn't have any bad effect on the scheduled 
processes, I checked V16N36 and it worked correctly.
```

Original issue reported on code.google.com by `gergo.ga...@gmail.com` on 16 Nov 2010 at 10:43

thewonderidiot:
@indy91, @eddievhfan1984, I've been kind of wondering if this old issue is  related to the 1201s and 1202s. It would certainly lead to increases processing time, although I would have guessed the impact to be much more severe if it was right for all instructions. The relevant line looks like it has since move to 1888.

thewonderidiot:
I'm pretty sure this issue is correct. What happens is this:

#### First call to agc_engine()
```
// Add delay for multi-MCT instructions.  Works for all instructions 
// except EDRUPT, BZF, and BZMF.  For those, an extra cycle is added
// AFTER executing the instruction -- not because it's more logically
// correct, just because it's easier.
if (!State->PendFlag)
  {
    int i;
    i = QuarterCode >> 10;
[...]

PR #56, opened 2016-09-11
#26: Correct cycle count for multi-MCT instructions

thewonderidiot:
This is Gergo "[Massive Camel](https://code.google.com/archive/p/virtualagc/issues/26)"'s fix for #26. I posted a description of what's happening on the issue.

@indy91, this will almost certainly make the 1201s and 1202s go away -- the total CPU load is going to be greatly reduced. I gotta say, it's pretty friggin awesome that this ended up mattering for something, though. Especially since it resulted in those particular alarms!

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -254,6 +254,8 @@
  *				is 0 and the divisor is not, the quotient and
  *				remainder are both 0 with the sign matching the
  *				dividend.
+ *		09/11/16 MAS	Applied Gergo's fix for multi-MCT instructions
+ *				taking a cycle longer than they should.
  *
  * The technical documentation for the Apollo Guidance & Navigation (G&N) system,
  * or more particularly for the Apollo Guidance Computer (AGC) may be found at
@@ -1885,7 +1887,7 @@ agc_engine (agc_t * State)
       if (i)
 	{
 	  State->PendFlag = 1;
-	  State->PendDelay = i;
+	  State->PendDelay = i-1;
 	  return (0);
 	}
     }


indy91:
I'll try this asap. To be fair though, I'm sceptical. This seems too good to be true, but I of course hope it is right. During some rendezvous calculations (P35) the CMC behaves as fast as I read in a technical debriefing. So after knowing that I was pretty convinced that the Virtual AGC is as fast during complex operations as it should be. What kind of instructions does this apply to?

thewonderidiot:
It applies to most of them, really, excepting TC, TCF, BZF, and the unprogrammed sequences. Just about every other instruction requires two memory accesses -- one to load its argument, and one to load the next instruction. It's possible that in cases like the above, when the CPU isn't taxed to the max like during landing, the only observable effect would be that you spend less time running the dummy job. But I don't know much about how those calculations are done, so I could be wrong.

indy91:
First run through the descent. No program alarms. LPD time remaining in P64 is still always displaying 0 or 99 (max). Probably an unrelated issue to the program alarms. I'll check the pad loads for it again. I did some LPD commands, not sure if they did anything though. I added LR velocity today and the LM is sometimes randomly banking, which almost looks like a reaction to LPD even when I haven't commanded anything. I'll try a few more extreme commands during the 2nd attempt. So far it looks good. Is there any way to analyse the CPU load and compare them to the expected values to confirm that this change is correct?

thewonderidiot:
I don't think so. I've tested the changes here to make sure the instructions take the expected number of calls to agc_engine() now.

It's definitely a great thought for the future though. Thanks to the real 1201s and 1202s, we know the expected CPU load is about 85% during the descent. I'm sure we can come up with a way to measure that, but I don't think we'll be able to do anything useful with it until all of the AGC's interfaces are complete and working, and we believe the I/O load to be correct (what remains here? are you guys already to this point?). We'll also have to account for counter handling differences... like I'm pretty sure the radar stuff is currently stuffed into RNRAD instead of using the cycle-stealing SHINC and SHANC.

indy91:
LPD is not working yet. I'll have to look into that. I doubt it's a Virtual AGC problem, although I have the feeling that I saw numbers other than 0 and 99 for the time remaining before...

But no 1201 or 1202 alarms anymore. The COMP ACTY light even sometimes has times to be off now!

indy91:
I wasn't sure where to post this. Here a quick demonstration video of the lunar landing with NASSP, made possibe by this and other fixes to the AGC engine: https://www.youtube.com/watch?v=sHaS6sYJsMg I have done such a video whenever a new feature was working in NASSP. So here is the one for the lunar landing, I thought you all would enjoy to watch it!

rburkey2005:
Perfectly fine!  Though I'm sure more people would see it on the Virtual AGC discussion list.  I took the liberty of putting an embedded youtube link to it near the top of the Virtual AGC website, though at present only at https://rburkey2005.github.io/virtualagc, since I'm temporarily not pushing changes all the way to http://www.ibibio.org/apollo until some problems are resolved there.

indy91:
Where can one find this discussion list? You can use my real name if you want, which is Niklas Beug. Doesn't have to be the very creative user name I used for Youtube 10 years ago :D

jimlawton:
Superb!

rburkey2005:
The list I was referring to was this one:  http://lists.ibiblio.org/mailman/listinfo/apollo.

It's not very big, and hasn't been very active, but using it does save me the trouble of reannouncing, which I probably might have done.  :-)  There has been a problem for the last six months, though, which I haven't been able to resolve yet, in that when you try to subscribe you may get a "captcha error" or something even though there's no captcha to be seen.  If that happens, let me know directly at rburkey2005@earthlink.net what name and email address you want to use, and I'll subscribe you manually.

And yes, I would prefer to refer to people by name whenever possible, so that credit goes to whom it is due, so I'll make that change.

thewonderidiot:
Sweet, thanks for linking the video! :smile: That's awesome!

PR #61, opened 2016-09-20
Initial Aurora tree

jimlawton:


jimlawton:
Check: 
 $ cd Aurora12 && make MAIN.agc.bin
succeeds, i.e. MAIN.agc contains valid includes.

PR #113, opened 2016-09-22
Update KEYRUPT_UPRUPT.agc

ehdorrii:


thewonderidiot:
For future submissions, would you mind using "## Page" instead of "## PAGE"? We try to reserve the all-caps for things in the original listings, and add lowercase for things we add in ourselves.

I'll make the changes to this one. Other than that, this looks great! I'm merging it in. Thanks!

PR #114, opened 2016-09-22
Transcribed pages 368-569. Resolves #83.

### LINKED ISSUE #83: Transcribe CONTROLLER AND METER ROUTINES (pp. 368-369)
Transcribe the scans for the CONTROLLER AND METER ROUTINES log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/CONTROLLER_AND_METER_ROUTINES.agc). This section ranges from pages 368 to 369.

This section has no corresponding log section in Luminary 099. It will have to be written from scratch.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Aurora12/). The original quality scans are available at [the Internet Archive page for Aurora](https://archive.org/details/aurora00dapg). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, *not* those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

nadalle:
After manual transcription, did a yaYUL --format.

### DIFF
--- a/Aurora12/CONTROLLER_AND_METER_ROUTINES.agc
+++ b/Aurora12/CONTROLLER_AND_METER_ROUTINES.agc
@@ -25,3 +25,78 @@
 #       https://www.ibiblio.org/apollo.  
 # The original high-quality digital images are available at archive.org:
 #       https://archive.org/details/aurora00dapg
+
+## Page 368
+
+                SETLOC          ENDEXTVS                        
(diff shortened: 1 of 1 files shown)


thewonderidiot:
Looks great, thanks!

PR #119, opened 2016-09-23
Transcribed pages 126-138, EXECUTIVE.agc

### LINKED ISSUE #69: Transcribe EXECUTIVE (pp. 126-138)
Transcribe the scans for the EXECUTIVE log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/EXECUTIVE.agc). This section ranges from pages 126 to 138.

[This section has a corresponding log section in Luminary 099](https://github.com/rburkey2005/virtualagc/blob/master/Luminary099/EXECUTIVE.agc). The two seem very similar. I recommend starting with the Luminary code, and looking for differences in the Aurora scans.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Aurora12/). The original quality scans are available at [the Internet Archive page for Aurora](https://archive.org/details/aurora00dapg). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, *not* those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

nadalle:
Resolves #69

Used the corresponding file from Lumiary099 as a guide, but the changes were
extensive enough that it wasn't very helpful.

### DIFF
--- a/Aurora12/EXECUTIVE.agc
+++ b/Aurora12/EXECUTIVE.agc
@@ -25,3 +25,476 @@
 #       https://www.ibiblio.org/apollo.  
 # The original high-quality digital images are available at archive.org:
 #       https://archive.org/details/aurora00dapg
+
+## PAGE 126
+
+                SETLOC  ENDSUBSF
+
+# TO ENTER A JOB REQUEST REQUIRING NO VAC AREA:
+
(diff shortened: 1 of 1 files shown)


thewonderidiot:
Looks great, thanks!

Issue #89, opened 2016-09-21
Transcribe IMU PERFORMANCE TESTS 1 (pp. 444-474)


thewonderidiot:
Transcribe the scans for the IMU PERFORMANCE TESTS 1 log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/IMU_PERFORMANCE_TESTS_1.agc). This section ranges from pages 444 to 474.

This section, along with the other two IMU PERFORANCE TESTS sections, have two corresponding log sections in Luminary 099: [IMU PERFORMANCE TEST 2](https://github.com/rburkey2005/virtualagc/blob/master/Luminary099/IMU_PERFORMANCE_TEST_2.agc) and [IMU PERFORMANCE TESTS 4](https://github.com/rburkey2005/virtualagc/blob/master/Luminary099/IMU_PERFORMANCE_TESTS_4.agc). It looks like the tests are somewhat jumbled between them, so while you may find some code in Luminary, a lot will probably need to be written from scratch.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Aurora12/). The original quality scans are available at [the Internet Archive page for Aurora](https://archive.org/details/aurora00dapg). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, *not* those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

thewonderidiot:
@7ofNine will be working on this one.

PR #123, opened 2016-09-25
LIST_PROCESSING_INTERPRETER.agc finalized

7ofNine:
how about task #89 IMU PERFORMANCE TESTS 1 as next one?
I might make sense to assign the whole IMU set to me in one shot i.e. Task #89 - 92


thewonderidiot:
Thanks! That sounds good to me, @7ofNine, I'll mark those three as yours.

thewonderidiot:
Actually @7ofNine, it looks like @ohommes started working on IMU PERFORMANCE TESTS 3, so I'll just assign the first two to you.

PR #125, opened 2016-09-27
Add Apollo 7 Rendezvous core dump

indy91:
Haven't worked with pull requests before, so I hope this works out all right.

This has been much, much easier than I anticipated. Originally I thought I had to convert our scenario files with the format "EMEM1234 56701" to the fomat used for core dumps in yaAGC. But while the scenario files for Orbiter have a different format, there still is the possibility to create core dumps accepted by yaAGC. 

All I had to do was bring the function MakeCoreDump() in agc_engine_init.c up-to-date. There were some changes to the saving format in 2009 which were never done in NASSP.

This file is my initial test of a NASSP supplied core dump for the standalone Virtual AGC. It was created in my Apollo 7 rendezvous practice scenario. I have tried the core dump in the Virtual AGC and it was able to calculate a rendezvous maneuver with Program 34 (Terminal Phase Initiation Prethrust) the same way it would in NASSP.


indy91:
Yeah, they did a short maneuver to get away from their S-IVB and then 24h later they re-rendezvoused with it. The first part of this re-rendezvous was ground calculated, because the AGC didn't have the full suite of rendezvous programs yet, and then the TPI and midcourse correction maneuvers were calculated onboard. The core dump happened just after the last ground calculated maneuver. They would then track the S-IVB optically with the sextant (no rendezvous radar or VHF ranging system yet) and calculate and perform the rendezvous as they would during the lunar missions. We have a nice rendezvous procedures that details everything the Apollo 7 astronauts had to do step by step, but Sundisk is quite different from Colossus, so we can only partly use it.

I'll write this and the necessary inputs for P34 in a readme file that will be named identically to the core dump.

PR #128, opened 2016-09-28
Readme for Apollo 7 rendezvous core dump

indy91:
Here my first attempt at a readme for the Apollo 7 rendezvous core dump. I found it a little bit difficult how much knowledge to expect from the reader. I guess the most important part are the step-by-step instructions for the DSKY inputs.


rburkey2005:
Fantastic! I love it. It's more than I expected, but everything I would have wished for.  Thanks!

Would you, by any chance, like to "announce" this on the discussion list as a new thing that might be expanded in the future if people like it?  I would, but our internet has been off for a day, and it was your idea in the first place.

Issue #96, opened 2016-09-21
Transcribe Q,R-AXES REACTION CONTROL SYSTEM AUTOPILOT (pp. 562-593)


thewonderidiot:
Transcribe the scans for the Q,R-AXES REACTION CONTROL SYSTEM AUTOPILOT log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/Q%2CR-AXES_REACTION_CONTROL_SYSTEM_AUTOPILOT.agc). This section ranges from pages 562 to 593.

[This section has a corresponding log section in Luminary 099](https://github.com/rburkey2005/virtualagc/blob/master/Luminary099/Q_R-AXIS_RCS_AUTOPILOT.agc). They appear to be quite different though. I recommend starting from the Luminary code, but a lot of this may need to be written from scratch.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Aurora12/). The original quality scans are available at [the Internet Archive page for Aurora](https://archive.org/details/aurora00dapg). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, *not* those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

thewonderidiot:
@7ofNine will be taking this.

PR #130, opened 2016-09-28
Task90 imu performance tests 475 499

7ofNine:
How about task #96 Transcribe Q,R-AXES REACTION CONTROL SYSTEM AUTOPILOT
as next one


thewonderidiot:
Sounds good to me!

PR #131, opened 2016-09-28
Apollo 9, 11 and 15 launch scenarios

indy91:
Here three launch scenarios for three different AGC versions (Colossus249, Comanche055 and Artemis072). Very easy to create, I start up the launch scenario in Orbiter, push in the computer circuit breakers and create the core dump.

Our launch scenarios in Orbiter all start with the Backup Crew Prelaunch Checks, so the clock will show about T-4 hours. Starting Program 01 per normal checklist doesn't work in the standalone Virtual AGC because it returns a 210 (IMU not operating) program alarm, but that was expected of course.


rburkey2005:
Swell, thanks!

PR #133, opened 2016-09-29
Apollo 8 Program 37 procedures and Apollo 11 P63 Ignition Algorithm core dumps and readmes

indy91:
Here two more scenarios, then I am out of ideas for now.

The Apollo 11 PDI one is for showcasing that the ignition algorithm is working and how far you can progress into Program 63 of the LGC in Virtual AGC. If you bypass all the stuff the LGC wants from you (Landing Radar on! Every PGNS switch in auto!), then it works pretty much until ignition. Then it doesn't measure any accelerations (oh wonder!) and will start the engine fail routine.

With the Apollo 8 Program 37 core dump I went a little bit overboard with my descriptions, but I think it was worth it. This is not only a demonstration that the AGC can do something, but it includes a workaround that had to be done on Apollo 8 and requires a few very quick DSKY inputs while the AGC is already calculating something. The whole thing is a long procedure and maybe the readme is not very well written at a few points, but I hope it shows how a problem with the Apollo 8 AGC (wasn't fixed until Apollo 10) was overcome.


rburkey2005:
Super!  Don't strain yourself for ideas.  I view this (and everything in the project) as something for the long term and not just for today, so let the ideas come to you at their own pace.

indy91:
All the scenarios so far didn't take a lot of creativity, I knew pretty quickly that they would work in the standalone Virtual AGC. I asked for ideas or requests in the discussion group, so if nothing comes up in there, then I am finished for now.

PR #138, opened 2016-09-30
correct duplicate entries

7ofNine:
File should also be properly formatted according to given formatting rules.
It contains tab instead of blanks and fields are incorrectly aligned


thewonderidiot:
Nice catches, thanks!

jimlawton:
We've been using tabs for code entry since the beginning of the project. If Ron decides to change the way formatting is done globally with yaYUL that's fine, but right now tabs are fine.

7ofNine:
I know, but then this should not be existing https://github.com/rburkey2005/virtualagc/wiki/Aurora-12-Transcription . See specifically the section “Formatting”.

Which rule is supposed to be used if that is the one that has been given directly in connection with the transcription?

jimlawton:
That's fine. Use it for the modules you're transcribing. 

This is the first time we've had a written down style guide, so there are bound to be some teething problems. I've been doing these transcriptions for a very long time, so it may take me a little more time to adjust :-).

jimlawton:
OK, maybe it's time to bite the bullet and run `yaYUL --format`. I've been putting this off for too long.

rburkey2005:
My personal executive summary would be this:  Your time is precious, and if files have been created with all tabs rather than spaces, I doubt it's worth your effort to track them down and manually fix them.

But here's the long answer I'd give.

I fear that Jim is right.  At least for my part in it, I didn't intend these "rules" to be a straight-jacket, and the rules are new enough that there are bound to be some problems with them at first.  They're more of a guideline for somebody has no clue as to what's going on than anything else, but they're also nice in terms of maintaining consistency from one file to the next, so as to avoid disorientation when you have to look at more than one file. 

The iron-clad rule you don't want to break is that opcodes do begin at column 16 or the 2nd tab-stop, and not at a column < 16 or > 16, because the assembler does use that information under some circumstances, at least for formatting purposes.

Tabs vs. spaces are the knottiest problem, in that when they work properly you can't tell by looking which were used.  But when they don't work ... ick!  Ultimately, `yaYUL --format`, or something like it, can convert everything to some standard format, which at present happens to be spaces rather than tabs.  So that's the reasoning, and it's certainly disputable.

jimlawton:
I ran `yayul --format` and checked in the reformatted version. Looks good I think. I was afraid it would mess with comment formatting, but it looks like it preserves comments.

BTW Ron, while we're on the subject of `holy wars`, I opened a ticket about line-endings (CRLF vs LF). I don't think the transcription page says anything about it. As long as users leave the Git settings at default then it will be fine. The only problem is if someone changes their settings on Windows to preserve CRLFs on commit, then we could end up with a mess. We can always run `dos2unix` on everything later I suppose...

7ofNine:
I didn’t mean to start the never ending war of formatting. It would be preferable to have a consistent “rule”. I personally think that tabs should never be used. For me they are a reminder of pre-historic times when punching blanks into cards was actually work and storing blanks on tapes (paper and/or magnetic) was a big waste of precious space. Visually they can not be distinguished on the screen but when I change the settings for tabs  the original intended formatting disappears. Good or bad might be a matter of opinion but I prefer to see the original as intended by the author.

 

With respect to CRLF: I am working on Windows and leaving the settings in Git at the default is what worked best for me. Even in a professional environment with non-homogenous local work environments, that is what we used . That reads as: check-out to the local format check-in to the format it came from. I don’t see why one would ever use anything else. The server determines what format it uses and this remains transparent to the client.

 

Conclusion: I will continue doing what I was doing before ;-)

Hartmuth

rburkey2005:
I didn't know that GitHub magic was involved.  So I guess it's a non-issue.

And yes, `yaYUL --format` does try to preserve comment spacing, though there's not much effort in doing that other than converting tabs to spaces.  But there's not much it can do if somebody was using tabs set to something other than 8.  Probably if a comment appears without a # it wouldn't be aligned right, either ... though maybe it would.

PR #139, opened 2016-09-30
Fix unrecognized opcodes and some messed up comments

7ofNine:
At this momnt only undefined EB/FB values and locations remain

### DIFF
--- a/Aurora12/AOSTASK.agc
+++ b/Aurora12/AOSTASK.agc
@@ -143,7 +143,7 @@ COEFFA1         CAE     ITEMP2          # FORM COEFFA(AOSQ)
                 TS      AXISCNTR
                 
 ALLAXES         INDEX   AXISCNTR
-                CSS     AOSQ            # FORM ABSOLUTE VALUE OF AOS
+                CCS     AOSQ            # FORM ABSOLUTE VALUE OF AOS
                 AD      ONE
                 TCF      +2
                 AD      ONE
@@ -179,7 +179,7 @@ ACCMINVL        INDEX   AXISCNTR
 AXISLOOP        INDEX   AXISCNTR        # SET UP 1/NJTSX TO 1/2JTSX FOR URGENCY
                 TS      1/NJTSQ         # (DOES NOT COUNT FOR U,V-AXES)
                 
-                CSS     AXISCNTR        # DECREMENT LOOP COUNTER AND TEST FOR END
+                CCS     AXISCNTR        # DECREMENT LOOP COUNTER AND TEST FOR END
                 TCF      +2
                 TCF     NEXTCALL        # FINISHED LOOP
                 
--- a/Aurora12/CONTROLLER_AND_METER_ROUTINES.agc
+++ b/Aurora12/CONTROLLER_AND_METER_ROUTINES.agc
@@ -94,7 +94,7 @@ XAXOVRD         CA              RHCY                            # YAW CHANNEL ON
 
 NORATE          CAF             ZERO                            # SET RATE COMMANDS TO ZERO
                 TS              PCOM                            
-                RS              RCOM                            
+                TS              RCOM                            
                 TS              YCOM                            
                 TC              TASKOVER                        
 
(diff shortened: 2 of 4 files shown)


thewonderidiot:
Great, thanks!

7ofNine:
BTW if somebody wants it I can occasionally run my home cooked JaYUL assembler (yes it is in Java) over the preliminary results and check for more of these things. It is a one pass assembler and doesn’t spit out the thousands of fatal errors one sees from the original yaYUL, i.e. one can actually find them easily.

rburkey2005:
Trying to keep secrets from us, eh?  Sounds cools, so you should give it a shot.  Why did you create it?

jimlawton:
Hey everyone does one at some point Ron :-).

Since it's confession time... I did one too, in Python naturally. 
https://github.com/jimlawton/pyagc

rburkey2005:
Really, the nerve!  Why am I always the last to know?  Hiding it in plain view like that was very clever, because I never would have thought to look there.

jimlawton:
:-)

I had great intentions to add Block I support, but never got round to it. It was a great learning experience though. Getting the link stage to produce correct output nearly drove me crazy. 

I used it last to try out different schemes when messing with SBANK= strategies, it's much easier to prototype stuff in Python.

7ofNine:
Just a little exercise in brain activity and how one could do it in a one pass. It’s pretty straight forward with some call-backs to fill in the holes.

It’s not ready for prime time,  e.g. it doesn’t create html output or a binary output but it creates the binsource type of output which than enables an easy compare with the provided binsource files.

 

I am actually thinking of a Java version of the  emulator/simulator with even a more detailed internal handling like B, X, Y, U registers and better timing. Not capable to be embedded but easily runnable on most machines. 

The beauty of retirement one can spend time on things one has fun with.

rburkey2005:
Well, I think that's great, and all these different approaches certainly have different merits.  I particularly like the idea of being able to assemble in a single pass, which I was sure was possible, but didn't want to mess with.  I'm as guilty as anyone else of giving up on perfecting the system or trying a different approach once I had something that worked for me, and it sounds as though there are some advantages in what both of you were doing.

I'm not sure that a more-detailed handling of B,X,Y,etc. is really feasible, though, unless you abandon the concept of a basic-instruction simulator and go directly to having a command-pulse simulator instead.  That's what John Pultorak did with his Block 1 system --- and I mean in the Block 1 software simulator he wrote, not just in his Block 1 hardware simulator --- but working at that level is a very curious thing.  You'd think that it's just a question of dividing up every basic instruction into 12 or 24 or whatever command pulses, but in some ways it's not really like that.  It's not as easy to divide up sets of command pulses into basic instructions as you might think, and if you want to think at the level of basic instructions rather than command pulses, it becomes very tricky and confusing to work with unless you give a lot of thought to how to do so, which John doesn't seem to have.  I sort of retrofitted that into his simulator, and it only worked tolerably well.  On the plus side, though, you do bypass some of the problems which Mike has lately been fixing, so belatedly, in yaAGC.  To give you an example, the B and Z registers for an instruction aren't set up during the instruction cycle for the basic instruction cycle they correspond to, but some time (and not always at the same time) during the preceding basic instruction.

thewonderidiot:
I was at one point also working on assembler, citing the fact that the binaries I needed to run on the hardware needed to have different bank ordering and parity bits. Then I realized I could be much, much lazier by just adding those things to yaYUL :)

Block 2 is interesting in other ways that really threw me off when implementing my memory module -- the memory cycles actually straddle the MCT boundary. Memory accesses for the next instruction are already in progress in the last couple of timesteps of the one preceding it.

@7ofNine, if you do get around to writing such an emulator, feel free to hit me up with any questions you run into. Certain instructions do things that aren't documented anywhere, even in AGC4 Memo *#*9. For example, DV triggers some carry-suppression circuitry that is controlled by one of the control pulses (CLXC) and various timepulses. It also (if I recall correctly) sets bit 12 of the S register to 1 for the duration of the instruction. I think there are some others like that, but can't think of them off the top of my head.

rburkey2005:
I'm afraid you're not actually very good at being lazy. The lazy solution that immediately occurred to me upon receiving your patch was Why didn't he just post process the rope file rather than messing with yaYUL? That's a nice, zero learning curve approach that takes probably 15 minutes to implement.  :-)
I already have similar post processors like that myself in the Tools directory.

thewonderidiot:
Hrrrrm. That's a good point. Especially since I've now written exactly that for the Aurora binary. >_>

rburkey2005:
Well, yes, I was wondering about that, but I assumed there was something I was overlooking.


thewonderidiot:
Would you prefer if I took that bit out of yaYUL and added the rope converter to Tools instead? I'm quite sure I'm the only user of it, so it's not like anybody would miss it.

rburkey2005:
No, why bother? It's minding its own business.

Issue #137, opened 2016-09-30
Add ALGA alarms to yaAGC


thewonderidiot:
We've decided to add in the Night Watchman alarm, which goes off if the software goes for too long without reading from address 67 (NEWJOB). That's only one of the ALGA alarms, all of which trigger restarts through the same mechanism. The other ALGA alarms are:

- Parity Alarm -- All fixed and erasable reads must have the correct parity.
- Rupt Lock -- An interrupt of any type must happen frequently enough, but the computer can't spend too long servicing rupts. This one I have seen, and Validation would need to be changed to enable interrupts.
- TC Trap -- The TC or TCF instructions must be executed frequently enough, but the computer can't spend too long executing only TCs and TCFs.

We probably want to leave the parity alarm out of yaAGC, since its presence doesn't add anything other than binary compatibility to the hardware. Nothing's ever going to end up with incorrect parity in an emulator.

The others are probably worth adding at some point, especially if Aurora's going to want to trigger some of them. For the time being, though, just the Night Watchman should be enough to prevent most lockups.

rburkey2005:
The yaAGC doesn't support parity anyway.


Sent on my Virgin Mobile Samsung Galaxy S® 5

-------- Original message --------

thewonderidiot:
Yeah, that's what I was trying to say in not enough words. We'd have to add parity support to yaAGC for the parity alarm to be happy, but there's little point in doing so because there's then no way for it to become unhappy. And it would break backwards compatibility for all existing binaries.

thewonderidiot:
As part of this, I think I'll add the capability of the CH77 Restart Monitor too, so we can take a look at that channel to see when any of these alarms happen, and which.

rburkey2005:
Sure, why not?


Sent on my Virgin Mobile Samsung Galaxy S® 5

-------- Original message --------

thewonderidiot:
So, according to R-700v3 the behavior of the Night Watchman is:
"Occurs if computer software should fail to access address 67 at least once every 0.64 second."

The actual behavior, looking at the circuit, is roughly in line with this, but slightly different. More precisely, that should read:
"Occurs if address 67 is not read at least once within every other 0.64 second window".

 In other words, the timing pays no attention to exactly _when_ in the window the read occurs, and it's possible for the alarm to take up to 1.28s to trigger, if the last read was placed right in the beginning of the 0.64 second window that's not checked.

thewonderidiot:
[...]

PR #140, opened 2016-09-30
Night Watchman emulation in yaAGC

thewonderidiot:
This adds emulation for the first of the ALGA alarms (#137), the Night Watchman. It should be very accurate to the behavior of the hardware. Once every 1.28s, it starts looking for accesses to NEWJOB (address 67). If that doesn't happen within 0.64s, a reset is triggered. The reset takes 2 MCTs to get to address 4000, going through the sequences GOJAM and TC 4000. CH77 bits are accrued as alarms are looked at, and then emitted all at the end if any alarms were tripped.

The alarms will always occur and will always update CH77, but their ability to cause a reset can be inhibited with the command-line option --inhibit-alarms. This functionality is exactly analogous to the AGC Monitor signal NHALGA. It's plumbed through in much the same way as DebugDsky.

I also noticed that counter interrupts had the same bug as multi-MCT instructions --i.e., they'd take one more MCT than they were supposed to. I fixed that by decrementing ExtraDelay before returning at the end of the counter/alarm update block.

With this, Aurora 12 no longer gets stuck in the "00 lockup" (see #105) -- the Night Watchman successfully detects the infinite loop, causes a reset, and after the reset everything works fine. I've tested --inhibit-alarms this way, and with alarms inhibited, the 00 lockup remains inescapable.

There's required changes to Validation to keep it from angering the watchman when running without alarms inhibited. I'll handle those before merging this in.

@rburkey2005, how does this look to you?

### DIFF
--- a/yaAGC/agc_cli.c
+++ b/yaAGC/agc_cli.c
@@ -34,6 +34,7 @@
   Reference:	http://www.ibiblio.org/apollo
   Mods:         11/30/08 OH.	Began rework
                 08/04/16 OH     Fixed the GPL statement and old user-id
+                09/30/16 MAS    Added the --inhibit-alarms option
  */
 
 #include <string.h>
@@ -89,6 +90,8 @@ static void CliShowUsage(void)
 "                  messages being received from the DEDA, but never to\n"
 "                  send any.  That lets \"yaAGC --debug-deda --deda-quiet\"\n"
 "                  to be used alongside yaAGS without conflict.\n"
+"--inhibit-alarms  Prevents the simulated hardware alarms (Night Watchman\n"
+"                  Rupt Lock, and TC Trap) from causing resets.\n"
 "--cfg=file        The name of a configuration file.  Presently, the\n"
 "                  configuration files is used only for --debug-dsky\n"
 "                  mode.  It would typically be the same configuration\n"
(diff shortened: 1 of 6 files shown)


indy91:
It certainly does work now without my circuit breaker tricks in NASSP. It doesn't even seem to be bothered by the Guidance Control switch setting. It gets extremely unresponsive when I set the PGNS Mode Control switch to Auto though. Probably something unexpected with the DAP. Is there any DAP test in Aurora12?

rburkey2005:
Looks like wonderful magic that I didn't have to think about.  :-)  But seriously, it's just the kind of thing I wouldn't have thought about, or would have put off, and since no complaints were ever triggered by not having it, I never would have thought about fixing it later, so I'm not surprised that I didn't add it myself.  Great work!

I can't see any reason not to go with it, and I'll merge it and close the issue myself.

thewonderidiot:
Thanks Ron! :)

@indy91, if there is a DAP test, I don't know about it (but that's not to say it's not there). It'd almost certainly be hiding somewhere in the listing after the DIGITAL AUTOPILOT ERASABLE section.

There's been a lot of mystery surrounding why this listing is Aurora 12 and not Aurora 88+, since Aurora had already been at least at revision 88 for 6 months, and old versions weren't kept around for anywhere near that long. Our current best theory for this is that this isn't mainline Aurora; it's a "fork" from the DAP Group, who simply didn't change the name of the program. I could well be wrong, but I don't believe any DAP was in the mainline Aurora at all.

We asked Jim Kernan, who led up LM rope development and was Aurora's rope mother, and he said that the DAP Group didn't work directly on Sunburst. Instead, they went off and did their own thing while the LEM Group wrote Sunburst. Only once the DAP was ready was it merged in. So the theory is that this is actually the twelfth revision of the LM DAP... ever... tacked onto the back of whatever revision of Aurora was available at the time. Very early development, either way, so it's not super surprising that it's not the most stable thing. :)

PR #142, opened 2016-09-30
yaAGC: Added emulation of TC Trap and Rupt Lock alarms

thewonderidiot:
Pretty straightforward, now that the framework is there. Aurora now trips the Rupt Lock rather than the Night Watchman, and thus breaks out of the 00 lockup just as quickly as the hardware sim. Finishes #137, I think, since we probably don't want to model any of the other possible alarms.


rburkey2005:
Any reason to wait for NASSP on this one?

thewonderidiot:
I don't think so. The flight software (or at least Luminary 099) executes fine, unaffected by any of this.

PR #147, opened 2016-10-01
yaAGC: CCS and interrupt fixes for RUPTCHK

thewonderidiot:
There were two things causing RUPTCHK to fail -- CCS was considering overflow condition *after* it did its diminished absolute value thing, rather than before, causing it to take the wrong branch.

The other problem was with interrupt handling. Most importantly, when the AGC takes an interrupt, Z has already been incremented _past_ the address the interrupt will be returning to by the previous instruction. RUPTCHK was using the value of ZRUPT inside the TIME3 ISR to make sure the interrupt interrupted exactly the right place, so that was off by one.

I've also changed the interrupt selection logic. It was previously a round-robin sort of thing, but the behavior of the hardware was much simpler. The interrupt vectors starting at 4000 were ordered in terms of their priority, and the highest priority interrupt requested was always taken. So for us it's a simple matter of iterating through the request list until we find a request.

Lastly, and least importantly, I tacked an extra MCT on to interrupt handling since it takes two MCTs to fill both ZRUPT and BRUPT.

With these changes, yaAGC completely passes all of SELFCHK. Yay!

@indy91, it would be great if you could confirm that this doesn't terribly break anything for you before we merge this in! :)

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -1934,7 +1934,7 @@ agc_engine (agc_t * State)
 	  // request. There's two extra MCTs associated with taking an
 	  // interrupt -- one each for filling ZRUPT and BRUPT.
 	  // Search for the next interrupt request.
-	  for (i = 1; i < NUM_INTERRUPT_TYPES; i++)
+	  for (i = 1; i <= NUM_INTERRUPT_TYPES; i++)
 	    {
 	      if (State->InterruptRequests[i] && DebuggerInterruptMasks[i])
 		{
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -264,6 +264,12 @@
  *				warnings (and hence errors) for me.  (The
  *				"goto AllDone" jumped over it, leaving
  *				ExecutedTC potentially uninitialized.)
+ *		10/01/16 MAS	Added a corner case to one of DV's corner cases
+ *				(0 / 0), made CCS consider overflow before the
+ *				diminished absolute value calculation, changed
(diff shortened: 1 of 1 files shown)


indy91:
I've implemented the last few commits step by step into our AGC engine, so it took a little bit to check everything. This commit seems to break LPD commands. Everything else is working fine. Maybe my bad hack for the HANDRUPT isn't working with this?

Also, near the last line of this commit, there is a little bit of a difference between the NASSP and Virtual AGC agc_engine.c. We have:

```
// Remove ifdef because Luminary131 LM Autopilot code is using that feature
//#ifdef ALLOW_BSUB
	  State->SubstituteInstruction = 1;
//#endif
```

Our old default LGC version was Luminary131, but now we are only using Luminary099. At least until we have a working Apollo 12+ scenario. So, is there anything bad this line of code could do with Luminary099?

thewonderidiot:
@indy91, right you are. That one-character change should fix RUPT10s. It wasn't iterating all the way up to that request index.

Regarding the ifdef, I'm really not sure. I'd like to make that feature the default, especially if Luminary 131 and later are going to use it. @rburkey2005, why was that ifdef'd out?

thewonderidiot:
Also, thanks a bunch for incrementally testing all of this! :)

indy91:
And that proves once again, there is no better testing than running the most complicted program the AGC has to offer!

indy91:
When I try the self test in Aurora12 it now doesn't give any program alarm. It also doesn't use the computer activity light, but I guess that is expected.

PR #148, opened 2016-10-02
yaAGC+yaDSKY2: Updates to make DSKY handling more 'realistic'

thewonderidiot:
Here's the first batch of updates to yaAGC and yaDSKY2 to make DSKY handling a bit more realistic.
First, a gif of me testing the changes in Aurora (yes, I know it's a CM DSKY :P):
dsky_updates

And a description of what's going on:
1. Computer is started, which lights RESTART
2. V35E is entered, which performs a light test. Because the RESTART light was active, it is not extinguished after the test.
3. The RESTART light is acknowledged with RSET and turns off.
4. V35E is reentered. RESTART turns on for the light test, but goes off after the end.
5. V01N10E 77E is entered to show the state of the hardware alarm latches. Because I started with a previously used Aurora core dump, it's all 0.
6. The control pulse self test is started with V21N27E 1E. This is on my master branch, and the RUPTCHK fixes are not in yet, so RUPTCHK fails. When it fails, it tries again and again, rather quickly. 
7. Eventually the computer panics and a hardware alarm gets triggered. RESTART is lit, and the computer reboots.
8. V01N10E 77E is again used to check what caused the RESTART. It now shows 00010, which corresponds to a Rupt Lock. (I guess WHIMPER is being called from within the TIME3 ISR, since that's where the failing check is).
9. Finally, the new RESTART is acknowledged with RSET, and the light goes off.

Now for the details:

This adds the fictitious output channel 163, whose bits correspond essentially 1-1 with discretes to the real DSKY. Verb/Noun, KEY REL and OPER ERR flashing is now timed by the AGC itself (with VN being active low, and the others being active high).

The RESTART light has also been plumbed through this channel. It is set by turning the computer on, or getting a GOJAM from a hardware alarm. It can also be temporarily turned on via the Alarm Test bit in channel 13. It's extinguished via a reset bit from channel 10.

Note that none of these changes break backwards compatibility for previous DSKY interfaces to yaAGC. They simply add a new one, and modify yaDSKY2 to use that instead for some things.

My next goal is to get standby and its associated light working.

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -1486,6 +1486,13 @@ BurstOutput (agc_t *State, int DriveBitMask, int CounterRegister, int Channel)
 #define SCALER_DIVIDER 3
 static int ScalerCounter = 0;
 
+// DSKY
+#define DSKY_OVERFLOW 81920
+#define DSKY_FLASH_PERIOD 4
+static unsigned DskyTimer = 0;
+static unsigned DskyFlash = 0;
+static unsigned DskyChannel163 = 0;
+
 // Fine-alignment.
 // The gyro needs 3200 pulses per second, and therefore counts twice as
 // fast as the regular 1600 pps counters.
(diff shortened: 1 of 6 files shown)


thewonderidiot:
Added a LM.ini with the updated mappings as well. Though we have no programs that use them yet, I also added the correct mappings for the NO DAP and PRIO DISP lights. So in theory once we've got Luminary 210 transcribed they should "just work". :)

PR #150, opened 2016-10-02
Language manual updates concerning changes made to yaAGC over the past weeks

thewonderidiot:
And some other random things that I know or was able to test in the hardware sim.

 I hadn't thought of trying to access memory in bank 44 or higher, so I added that to Validation and took a look at what the hardware sim did. It wasn't very happy -- it still cycled the strands and strobed the fixed sense amplifiers, but didn't actually select any of the ropes or modules, so who knows what sort of nonsense it would have gotten back.

Anyways, my goal for this page is to one day remove all of the green text. I still need to add a section describing of all the hardware alarms, now that making them happy is required for successful program execution, and would like to move the description of all the non-fictitious I/O channels off of the developer page and onto this manual.

@rburkey2005, I didn't update changes.html, since I wasn't sure if you wanted to maintain sole control over that. I can, though, if you'd like.

### DIFF
--- a/assembly_language_manual.html
+++ b/assembly_language_manual.html
@@ -49,6 +49,7 @@
 
 
 
+
           0"</a> — Earliest preliminary design of the AGC for which I
         have any documentation, not used in any missions, and with no
         known existing code.&nbsp; There's a single document (that I
@@ -76,13 +77,6 @@
     assembler program.&nbsp; Additionally, AGC4 interpreter-language is
     covered.<br>
     <br>
-    For those who can tolerate such talk, this document serves as the
-    requirements for processing of machine language by the <span
-      style="font-weight: bold;">yaAGC</span> program, as well as for
-    the assembly language accepted by the <span style="font-weight:
-      bold;">yaYUL</span> program, and that's why this document is
-    versioned.<br>
-    <br>
     The original Apollo documentation from which the information
     presented here was derived, in roughly descending order of
     importance, is:<br>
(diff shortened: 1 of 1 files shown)


rburkey2005:
Wow, reviewing html changes is the ultimate frustration!  I didn't see anything that seemed weird, so I've merged it.

I didn't realize until a week or two ago that the i/o channels were so problematic, when I saw that: a) Sunburst was using channels completely unlike those I had started with so many years ago; and b) that I had never completed more than a fraction of the descriptions of the i/o channels on the assembly-language page; and c) that I had described i/o channels on the developer page that I should have described on the assembly-language page (or vice-versa, but not split into two places).  It seems as though they should all be on in one place, and differences should be described on a mission-by-mission basis.  Sounds like some thinking needs to be done.

@thewonderidiot Anyway, I hadn't given any thought to changes.html, but you're right.  The way that page is constructed it would be a hassle to merge changes (to changes.html) if more than one person edited it.  I'll do so.

Thanks!

thewonderidiot:
Yeah, no kidding. I would have thought that there'd be a good visual diff tool for that sort of thing by now, but it looks like there really isn't.

> Sunburst was using channels completely unlike those I had started with so many years ago

Oooo, really? That sounds interesting. Got any examples (or a page where I can find them in the listing)? One of the things I've often grumbled about is lack of documentation for every bit in every channel.

indy91:
I have looked a little bit into the input channels, too. The way NASSP is programmed will make a LM with Virtual AGC on the launchpad but without a CSM difficult. Anyway, to me it looks like the LM-1 was wired together with the Saturn the same way as a CSM. So the liftoff and guidance reference release input channel bits in the CMC are also used in Sunburst. So maybe that's where the confusion is coming from? Sunburst seems to have some sort of launch program which needs to be triggered somehow, so that's my theory.

rburkey2005:
Oops!  I think I may have been thinking about the counter registers.  
I've already looked at (and actually transcribed) 
https://github.com/rburkey2005/virtualagc/blob/master/Sunburst120/ERASABLE_ASSIGNMENTS.agc, 
where the counter registers are listed, but I don't think I did anything 
more than create a template for 
https://github.com/rburkey2005/virtualagc/blob/master/Sunburst120/INPUT_OUTPUT_CHANNELS.agc, 
so I don't think I've actually looked at the i/o channels in detail 
yet.  Embarassing.

But I think the principle of what I was saying was probably right 
anyway.  (And it applies to the counter registers!)

thewonderidiot:
Oh, now THAT is interesting. The CSM System Handbooks show the connections of the CSM to the S-IVB IU, so I wonder if the LM-1 Systems Handbook shows those connections as well.

Looking at [the index of Fort Worth's E-156N collection](https://www.ibiblio.org/apollo/NARASWoverflow/E156N.pdf), they do in fact have that document. I'm particularly fond of those systems handbooks, and am planning on taking a trip to Fort Worth in the relatively near term, so that might make it on the list of things to scan while I'm there (unless somebody else turns up with it first). :)

And yeah, counters apply equally well to the list of things that I grumble about for lack of documentation!

indy91:
Actually, I'm not so sure anymore. Rather looks to me like GRR and liftoff are normal flags and they are set by ground uplink. So no crazy input channel rewiring. Oh well, that should make it easier to implement in NASSP.

The CSM System Handbooks are nice. I once spend a long time with my head on a printout of the Entry Monitoring System schematic. I don't think a complete LM Systems Handbook is on the Internet, right?

thewonderidiot:
[This is the one I normally refer to](https://www.hq.nasa.gov/alsj/LMSysHandbk.pdf), but you're right, it's quite incomplete. So maybe it's worth scanning one of the later ones as well.

indy91:
Yeah, that one only has guidance and control subsystems. Good enough to get lost in the functionality of RUPT10/HANDRUPT.

Issue #97, opened 2016-09-21
Transcribe JET FAILURE CONTROL LOGIC (pp. 594-603)


thewonderidiot:
Transcribe the scans for the JET FAILURE CONTROL LOGIC log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/JET_FAILURE_CONTROL_LOGIC.agc). This section ranges from pages 594 to 603.

This section does not have a corresponding log section in Luminary 099, and I don't see any shared code. This section will need to be written from scratch.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Aurora12/). The original quality scans are available at [the Internet Archive page for Aurora](https://archive.org/details/aurora00dapg). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, *not* those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

7ofNine:
I 'll take this one

thewonderidiot:
Sounds good!

PR #151, opened 2016-10-03
Task96 q r axes reaction control system autopilot 562 593

7ofNine:
I'll take task #97 JET FAILURE CONTROL LOGIC as next one


thewonderidiot:
Excellent, thanks! :)

PR #157, opened 2016-10-04
Use common Makefile

jimlawton:
Change the flight builds to use a common make include file (`Makefile.inc` at top-level). 

 - This means we can modify build rules in one place. 
 - Differences in YUL options per-flight are specified using the `EXTRA_YUL_ARGS` variable, see `Solarium055` for an example.
 - Added a `format` target. This reformats all the sources in a flight directory with `yaYUL --format`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ Comanche055/Comanche055.lst
 Luminary099/Luminary099.lst
 Luminary131/Luminary131.lst
 Solarium055/Solarium055.lst
+oct2bin.proof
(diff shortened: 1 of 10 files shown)


rburkey2005:
Seems reasonable.  I often do this kind of thing, and after a while nobody understands how it works any longer (including me), but this doesn't seem to be likely to happen in this case.  I do notice that the new Makefile.inc still lists its name (in the file's header comments) as just "Makefile".

jimlawton:
Argh, didn't read your comment properly before I merged.

jimlawton:
I'll fix it.

jimlawton:
https://github.com/rburkey2005/virtualagc/pull/158

Issue #106, opened 2016-09-21
Correct code errors in the completed transcription


thewonderidiot:
Once all log sections are done, the octal listing is complete, and yaYUL supports BLK2, combine everything and proof each log section. Then celebrate victory!

thewonderidiot:
@7ofNine is working on this. Thanks!

rburkey2005:
@thewonderidiot If nobody steps forward to claim this, you can assign it to me.  

The brief description doesn't encompass the perhaps-unobvious fact that at this point the assembler will generally indicate that there are many thousands of errors, though in reality only hundreds :-), so that it's not really proofing so much as debugging.  I've always ended up doing it in the past, but I'm not oblivious to the fact that someone else might want to do it if they fully understood the nature of the task ... which is why I didn't simply arbitrarily assign it to myself.

rburkey2005:
@thewonderidiot I think there's either a miscommunication or or perhaps I misunderstand what's going on here.  I've signed up for updating yaYUL, while @7ofNine is signed up for proofing the completed transcription, but in my mind (other than program comments) those two tasks are inseparable ... yaYUL can't be debugged without debugging the transcription, and vice-versa.  Do you view this task as just a proof-reading of the comments?

7ofNine:
I guess there are different ideas about it. I could do the "proofing" independently from yaYul, i.e. what is now available is a "clean" (+/- comments)  input source that would allow the changes necessary for the assembler. That's how we already discovered some of the issues like opcodes and superbank bits.
I can do the changes to yaYul, too. The only problem here is that I don't work with *nix/Mac and have a version of yaYul that is massaged into compiling/working under Windows and this version most likely will break the integration build and everybody else. I simply can not verify any changes to yaYul in a *nix environment.

rburkey2005:
I think this should have been separated into two different tasks, namely fixing up the AGC code and proof-reading the comments, since the tools and methods status as milestones are very different for those two activities.

In what way have you changed yaYUL that makes it specific to Windows?  Perhaps you could email me a copy directly.

thewonderidiot:
@rburkey2005 Sorry, you can chalk that up to my being new at this. Namely, I didn't know that comment proof-reading was done as a separate step after transcription was complete. What tools other than an image viewer are useful for that?

rburkey2005:
I don't think comment proof-reading was ever done at all, except as a part of the transcription itself.  It seems like a good idea, though.

[...]

PR #162, opened 2016-10-04
some missing instructions and typos

7ofNine:
if you want yo can assign #106  to me. I obviously have started doing it anyways.

Block 3 is already clean


thewonderidiot:
Thanks! I would if I could (I really wish GitHub would let you assign issues to anybody...). I'll just make a note that you're working on it.

PR #163, opened 2016-10-04
Refactor build system slightly

jimlawton:
Modify makefile.inc so it can be used for the non-standard builds also (Validation, Solarium055, Sunburst120). Change those builds to use Makefile.inc.

This does not generate HTML output by default. Does anybody use that?


ohommes:
The html output is nice for presentation mode online and click through the TCs etc.
I don't mind if being off by default. But I still think it is a great way of presenting the code without actually running it if you just want to navigate the code like a source navigator.

Sunburst120/Makefile
```diff
@@ -9,5 +9,5 @@
 # 		2016-10-04 JL	Change to use Makefile.inc.
 
 BASENAME=Sunburst120
-EXTRA_YUL_ARGS="--blk2"
+EXTRA_YAYUL_ARGS="--blk2"
```
thewonderidiot: Sunburst 120 was "for AGC", so it doesn't need --blk2. It's probable that Sunburst 37 will, though.

thewonderidiot:
I use the HTML for debugging, since it has the addresses displayed by all the instructions and is more accessible than the .lst file and the original scans. It's not a big deal to me, though, if they're not built by default, since I normally also manually build with --hardware.

rburkey2005:
I think the reason it's built by default is that the VirtualAGC target needs it.

jimlawton:
Sorry about that guys. I'll default it on again.

jimlawton:
Ok thanks Mike. I'll fix that.

rburkey2005:
I didn't know anybody was actively using the HTML.  Finding that out prompted me to go in and fix something about it that has been bugging me for some time, namely that because the complete listing is split up into separate files that are only tenuously linked together by MAIN.agc.html, there's no way to globally find anything easily.  For example, if you wanted to just go to GOTOP00H, there's no way to do that unless you already know what file it's in.  Anyway, I've just changed "yaYUL --html" so that the symbol table at the end of MAIN.agc.html is now completely hyperlinked.  In other words, you can find anything globally by just clicking the link for it in the symbol table.  Not perfect, but much better than nothing.

Another cool thing would be to have some way to find out all the places any given label had been referenced _from_, but that's something for another day, if ever.  :-)

PR #164, opened 2016-10-05
yaAGC: Added support for standby mode

thewonderidiot:
The only things that operates in standby mode, for the moment, are the scalers and the DSKY. I found while doing this that the HI SCALER and LOW SCALER input channels were incrementing too slowly  -- LOW SCALER should increment at 3200pps, not 1600pps. Since those channels are how the AGC determines how much time elapsed in standby mode, time appeared to move at half speed while the computer was in standby. As a side bonus of speeding them up, TIME6 finally has the right phasing.

Here's a little gif of the light test and standby mode in action:
yaagc_standby

### DIFF
--- a/yaAGC/SocketAPI.c
+++ b/yaAGC/SocketAPI.c
@@ -259,6 +259,10 @@ ChannelInput (agc_t *State)
 			      State->InterruptRequests[10] = 1;
 			    LastInDetent = InDetent;
 			  }
+                        else if (Channel == 032 && 0 != (Value & 020000))
+			  {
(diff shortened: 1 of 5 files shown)


indy91:
I'd love to implement this in NASSP, but I'll have to dig into the code a bit first, I know there are a few differences in how this has to be handled.

thewonderidiot:
Let me know if you want me to change how anything is implemented here, if doing so would make that easier for you. The biggest thing is that State->SbyPressed needs to be reset to 0 whenever the PRO button is released, in whatever place you fill input channel 32.

PR #167, opened 2016-10-06
Initial checkin of YUL structure

jimlawton:


rburkey2005:
Do we really want to do this?  I'm not opposed to it by any means, but I never intended to transcribe it, because I thought it would just be too much to ask for.  Of course, it would be great to have it, because it would be searchable and so on.  On the other hand, we'd have literally no way to check it for correctness other than proof-reading.

Just asking.

jimlawton:
Well, I do, I can't vouch for anyone else :-). 
It will take a very long time to transcribe, may not be finished for years, and as you say we have no way of proofing...

If you prefer I can put it in a separate repo?

rburkey2005:
Of course not!  Have at it!  I just worry about dragging people along with my little obsessions, so I was trying to hold back.  If you're a willing volunteer, why not do it?  I'm not sure what the effort would be to write a MITIGUS assembler -- after all, _everyone_ writes them from time to time -- but it would have to be less than the effort of transcribing, so maybe we could come up with some way to verify correctness after all.

jimlawton:
Yeehaw.

ohommes:
Uhhh.... I have tried this question before but without success with Ron but I will ask again since it might be more relevant.

Can we have more organization in our repo:

Apollo/
   AGC/
        Colossus237
        Luminary099
   LVDC/
        stuff
   YUL/
        stuff
Gemini/
   Stuff
yaAGC
yaYUL
...

This is just an example... If not given we are all busy then I will lay low again for 6 years with this question.

rburkey2005:
Well, Jim was advocating it a few weeks ago, and my response to him was probably as unsatisfying as the one I gave you (which I have no recollection of). If everyone wants it but me, I won't oppose it. It's just that I don't see what problem it solves. It certainly makes it cleaner for someone unfamiliar with it, but is the organization of the repo something anyone is actually having trouble with? (It's a real question!)


rburkey2005:
I guess if you guys want to get together and work out a hierarchy that suits you, and Mike doesn't come up with an objection, I have no real objection either. Obviously, I'd like a chance to critique it.
I guess my main concern is the period of time when stuff won't be working. 

PR #244, opened 2016-10-09
Task106 proof completed transcription part1

7ofNine:
First set of corrections. There are more checks and updates needed when code  transcription is completed.
There are also BLK2 yaYul updates needed. see #104.


thewonderidiot:
Awesome! :)

PR #247, opened 2016-10-12
Task106 proof completed transcription part2

7ofNine:
Some more fixes for typos and missing instructions.
Most of the remaining errors (possibly all) are caused by unresolved names due to missing parts of the source


thewonderidiot:
Sweeeet. Getting close!

Issue #240, opened 2016-10-07
Transcribe WAITLIST (pp. 1062-1074)


thewonderidiot:
Transcribe the scans for the WAITLIST  log section into [its source file](https://github.com/rburkey2005/virtualagc/blob/master/Sunburst120/WAITLIST .agc). This section ranges from pages 1062 to 1074.

This section has a corresponding log section in Luminary 099, from which it has already been populated. [Aurora 12 also has a corresponding section](https://github.com/rburkey2005/virtualagc/blob/master/Aurora12/WAITLIST.agc). They're probably all pretty similar.

Reduced quality scans can be found [here](http://www.ibiblio.org/apollo/ScansForConversion/Sunburst120/). The original quality scans are available at [the Internet Archive page for BURST120](https://archive.org/details/yulsystemforagcr00nasa). Note that in both cases, the page numbers specified on this ticket are those printed on the original pages, not those in the filenames. The numbers will be close, though, so you should be able to find them pretty quickly. The start of the log section should be fairly obvious.

cloudynetwork:
I'll take this one as well, although it will take me a couple of days (it not being the weekend).

thewonderidiot:
Thanks! This is a marathon, so take as much time as you need. :)

cloudynetwork:
@thewonderidiot Cheers! I find it goes rather slow as I'm finding myself thinking at length about the differences between versions. It's really illuminating to get the opportunity to work though this.

rburkey2005:
@cloudynetwork Thanks for saying that.  The Virtual AGC project (in my mind) has several goals, which have evolved over the course of time.  But one of those goals has been to be able to present the full sweep of the AGC software development, from the beginning of Apollo to the end, so that somebody with the intelligence and the determination to do so could looked at the software and grasp, at least in part, how the AGC developers' thinking and their software had evolved over time.  Obviously, if it were done extensively that would be quite a challenge, to say the least, and I don't honestly expect anybody to ever attempt anything like that unless it's for a doctoral dissertation in the history of technology.  :-)  But regardless, it's pleasant to hear somebody else express thoughts along those lines.

cloudynetwork:
@rburkey2005 
[...]

PR #249, opened 2016-10-16
Closes Issue Transcribe WAITLIST (pp. 1062-1074) #240

cloudynetwork:


thewonderidiot:
Excellent! I can't wait to read your write-up of the differences. :)

cloudynetwork:
@thewonderidiot I think in the interests of proceeding with the transcription I will wait until the end to produce something from my notes. Also I would like to have a yaYUL assembly listing to compare with LUMINARY too. I find the listing invaluable for following the program flow in a way that is often challenging with the isolated source files.

PR #252, opened 2016-10-17
Task106 proof completed transcription part3

7ofNine:
still a few outstanding source files. not complete, yet


thewonderidiot:
Nice work! Getting close. :)

PR #280, opened 2016-11-03
According to scans should read as "=" instead of '-'

7ofNine:
A small fix I discovered. In ERASABLE_ASSIGNMENTS  it should read "=" instead of "-" as it appears currently in the source. For some reason yaYul does not complain about it.  My version of the assembler on the other hand screams about it. The correction should have no further effect as that the symbol LEMMASS1 shows actually up as defined with value 0144. It is nowhere else used in the code.


rburkey2005:
Without bothering to actually look at yaYUL, my vague recollection is that yaYUL simply ignores everything after the first operand field in an ERASE, though it prints it, and just considers it as documentation.  I think it can get away with that because this construct always appears at the _ends_ of banks, so there are no additional ERASE ops after that (or else they'd be at the wrong address).  So all it cares about is the starting address.  Probably yet another thing in yaYUL that should be fixed in an ideal world.

PR #377, opened 2016-11-18
SUBROUTINE_CALLS

7ofNine:
Shouldn't this actually be created by yaYul??


rburkey2005:
Ideally, probably.  However, yaYUL only makes a limited attempt to reproduce the summary information that YUL or GAP produced, and ultimately only really produces the octals, symbol table, and checksums.  Back in the day, I didn't know what SUBPRO did (though it seems pretty obvious now), and it obviously wasn't a comment, so I just implemented it as doing nothing.  So in the final analysis, yaYUL is going to do nothing at all with those lines, and they can be entered either as "code" or as "comments" if we want to preserve them, but yaYUL certainly won't generate them itself.

PR #426, opened 2016-12-07
Change incorrect P00 to the correct POO in already transcribed files.

7ofNine:
Affects comments, labels and operands
biggest outstanding offender is unstranscribed P51-P53.sgc


thewonderidiot:
Hahaha, nice branch name. :) Thanks!

rburkey2005:
Sorry I didn't notice this task before now, and I don't know what you (@7ofNine) did to fix this, but I'll belately point out that there are _some_ places where P00 (zero) really is used rather than POO (oh), so a global replace without inspection of the individual instances wouldn't always produce the right result. I know that my previous comments may have produced that impression, and if so, I apologize.  (Of course, if you didn't do that, and are offended by the implication that I think you might have, then I apologize for _that_. :smile:  However, if it were me, I might have done that, which is why the thought has occurred to me.)

7ofNine:
I didn’t do a global replacement. I checked several P00 and compared them with the original listing and didn’t see any where P00 would be correct instead of POO. Even in cases where it is embedded into labels it seems to be POO. Most of them are in code that I transcribed. Focusing on label definitions should resolve nearly everything else and GOTOPOOH seems to be the biggest offender. I suggest to create a task for that purpose and assign it to me . I’ll do a last verification just before the overall verification task, so that all code has been transcribed and one doesn’t have to separate transcribed files from non transcribed ones. The only tedious ones are where POO/P00 is used only in comments or labels that are unused.

rburkey2005:
Okay. :smile:  I wasn't taking into account that my context wasn't just this program version. I've been working on Sunburst the last few days, and I'm pretty sure there were instances of both POO and P00, in the comments in the scans. Probably those were later corrected in other program versions. Or I'm confused ... take your pick.

7ofNine:
So far ( not done, yet)  I have found one label (DNLADP00, still looking) that should have P00 instead of POO in it. But  I saw also a case of a COUNT pseudo operation that should have  a P00 instead of a POO. The comments and these COUNT operators  and similar “stuff” will be the work. There are currently 204 cases of POO(H) in the source.

PR #427, opened 2016-12-07
Transcribe DISPLAY INTERFACE ROUTINES (pp. 1338-1370)

kociak1984:
This was quite a challenge to be honest. Took me around 4 hours.


kociak1984:
Oh, date should be today, not tomorrow.. (is: 2016-12-08, should be 2016-12-07)
But to be honest - who cares :P

rburkey2005:
@kociak1984 There's no doubt that this stuff can be quite a challenge, particularly the more care you apply to it, which is why we really appreciate the effort!

kociak1984:
Trying to do my best.

thewonderidiot:
Indeed, thanks a bunch for the help! :)

PR #431, opened 2016-12-08
Fine tuning for POO/P00. Mainly updates to comments that contain P00

7ofNine:
Some comments actually have P00 in them


thewonderidiot:
Fantastic, thanks! :smile:

PR #433, opened 2016-12-08
Fix homemade and other typos, missing labels

7ofNine:


thewonderidiot:
Thanks! How many errors are we at now, by your count?

7ofNine:
After the P51-P53 updates there are about 50 (with my JaYul assembler). But lots of them have the same cause. Somewhere Bank 14 becomes full and it complains a lot about that. The original yaYul shows 41 fatal and 67 warnings i.e. we are in the same ball park. On the binary level it shows about 560 differences, but I am not sure if  I am really using the latest binsource file

rburkey2005:
If you mean that there might be a more-current binsource file offline, there is not.  I regard the binsource file as correct as of when the task for it was closed out.

7ofNine:
No worries. I am using a slightly modified hand created version for my comparison with my "binary" output, i.e. it is not the original binsource file. I dump the comments and some unwanted trailing spaces but I am too lazy to write a tool that would do that for me. I just checked and my version was created on Nov 27 and the latest I see is Nov 26. i.e. it's O.K. In any case it wouldn't be anything serious it only takes a couple of minutes to do it. The transcription is not completed anyways and it would be revealed in the final comparison.

PR #435, opened 2016-12-09
Transcribe PLANETARY INERTIAL ORIENTATION (pp. 1138-1146)

kociak1984:
I am actually reading those, I don't use any OCR...


thewonderidiot:
Thanks!

So far we haven't found any better way, really. We're just now starting to "use" OCR for verifying correctness of transcribed comments, but it's more hijacking an OCR engine for its character location detection than trying to get anything meaningful out of its actual recognized text. Our attempts to go straight from OCR to text have, unfortunately, not been terribly successful outside of the limited domain of the octal listings (which can be restricted to only 8 characters).

That being said, if you _do_ come up with a good way to reliably get trustworthy OCR working, we'd love to hear it! :)

PR #440, opened 2016-12-12
Task365 q r axes rcs autopilot

7ofNine:
There are 6 binary differences remaining. They all look like the need for a SUPERBANK workaround is needed and the consequential wrong bugger word.


thewonderidiot:
Great! :)

PR #441, opened 2016-12-13
Task372 correct code errors in the completed transcription

7ofNine:


thewonderidiot:
Sweeeeeet, thanks! :smile:

I'll spin up some Luminary 69 and Retread tickets tonight so we can get started on the next one!

PR #538, opened 2016-12-15
Update R31.agc

brf1:
Large amount of differences to Luminary 99. Is this one of the areas that might have changed in Rev2?


indy91:
Not from Luminary 069 to Luminary 069 Rev 2. There are two program change requests referencing R-31:

- "Correction to R-31 When LM is on the Lunar Surface"
- "Lessen Delays in R-31"

But these changes were done for Luminary 99.

thewonderidiot:
Thanks @brf1! If you keep going, keep sticking with the sections that have already had comments proofed in 99 (roughly, everything before this section). :)

PR #578, opened 2016-12-31
Add VS Code Extension

wopian:
VS Code syntax-highlighting extension for AGC, AGS, Argus and Binsource files.

https://github.com/wopian/agc-assembly
https://marketplace.visualstudio.com/items?itemName=wopian.agc-assembly


rburkey2005:
@jimlawton @ohommes I suppose one or the other of you is the logical one to look at this and at #579. I've just always accepted the syntax-highlighting as magic, and hence have no idea how to look at it myself.  :smile:

wopian:
>My only question is about the licence files. Are they necessary? They possibly conflict with the overall VirtualAgc licence, but also the licence is for the actual syntax-highlighting code which is not included here, only referenced.

README now just states what license the extension has and points to the repo LICENSE file.

Also stripped out the development side of my README, so both PRs just link to the repo + installation/setup instructions

jimlawton:
Thanks @wopian!

rburkey2005:
Thanks all.

PR #579, opened 2016-12-31
Add Atom Extension

wopian:
Atom syntax-highlighting extension for AGC files by @Alhadis.

This is also the repo GitHub's [linguist](https://github.com/github/linguist/blob/08b63e703345498e589397293ce5170917ba8067/.gitmodules#L728) uses for syntax highlighting.

https://github.com/Alhadis/language-agc
https://atom.io/packages/language-agc


jimlawton:
See comment on #578 re licences.

Alhadis:
The license is ISC. As permissive as you can possibly get.

PR #605, opened 2017-01-08
Reimplementation of EDRUPT, plus BRUPT and INDEX changes

thewonderidiot:
This is primarily a reimplementation of EDRUPT based on behaviors discovered studying the hardware sim. I'm going through a pull request instead of just pushing because it's more major of a refactoring than other small things I've been doing in yaAGC lately.

EDRUPT now doesn't have any handling as an instruction at all; it instead is taken as a generic interrupt request that can happen regardless of any inhibitions on interrupts. The highest priority pending interrupt is taken. Only when no interrupts are pending does it fall back on its old behavior of vectoring to A.

Sort of naturally combined with these changes, I made two other changes. The first was re-enabling BRUPT by default. The instruction is stored with INDEX applied, so it's now safe to allow interrupts to happen immediately after INDEX.

After these changes, all of the Aurora and bugfixed Retread tests in Borealis tests still pass fine, so I don't think I've broken anything. And EDRUPT now behaves exactly the same as the hardware sim in my temporary Validation checks of it.


rburkey2005:
I think this is something we'd want to know works in NASSP ... For an entire landing, if possible. Actually, for an entire mission, if possible.

thewonderidiot:
Yes, agreed 100%. @indy91, ping to let you know about these changes. It would be great if you could help us test them out! No rush at all, take your time with the next NASSP release -- this can sit here for a while. :)

indy91:
Will be implemented and tested of course. It goes on the pile of updates we have yet to do, starting with the standby mode. Testing it with the landing will happen as soon as this is implemented in our agc_engine.c, full missions won't be possible in the next NASSP version for a while.

rburkey2005:
I probably shouldn't have said "full mission".  I just meant that I wanted to see that it exercised for typical command-module activities in addition to a landing.  I didn't really mean end-to-end, launch-to-splashdown.  (Not that complete missions wouldn't be great, of course!)

indy91:
Oh, we will certainly fly many missions with these changes, probably not full missions just because of these changes. But I am sure we will find anything wrong with it fairly quickly.

thewonderidiot:
Great, thanks! In the mean time I'll add tests for BRUPT substitution, INDEX interruption, and various cases of EDRUPT to Borealis's instruction tests, to see if I can smoke anything out before then.

thewonderidiot:
Well, I'm working out the logic for tests surrounding EDRUPT. It's going to be very useful for the interrupted INDEX test.... but man. I think I understand now why it doesn't show up in Retread or Aurora. This instruction is a massive pain to write tests around... there are so many quirks and ways it can bite you. And you're practically guaranteed that it works if you can manage to start the tests at all, since it is just the generic interrupt control pulse sequence and if T3RUPT, T4RUPT, and KEYRUPT1 aren't working, then PINBALL isn't working.

thewonderidiot:
Hugh has provided an explanation as to its provenance!

> How did this come about, you might wonder? Well, it wasn’t because Ed asked for it, as near as I can recall. I think it was my idea to put it in there because it was a super-cheap feature that might be a powerful help to Ed, who was just starting his self-check work at that time. What you describe sounds like a good implementation of my then half-baked idea, and I’d be interested to hear if your tests find it able to serve that purpose. I guess Ed had enough of a problem just doing the normal-instruction checking within stringent schedule and ROM-usage constraints.
>  
> At bottom, it was a classical case of YBIC (whY? Because I Can!) on my part.

That explains why we haven't found another instance of it -- it's entirely possible that Ed himself never actually used it.

thewonderidiot:
Okay, I've added a new suite of tests to Borealis to accompany this pull request. Things that are checked out:

1) With interrupts inhibited, all four timer interrupts are triggered. They are then allowed to executed one at a time using EDRUPT, and the execution order is verified to ensure proper function of the interrupt priority chain.

2) While the above is being set up, the positive and -0 cases of DINC are checked out for TIME6, and TIME6 is verified to be ticking at the correct speed. It's also verified that it switches itself off (ch13 bit 15) upon completion. Not super relevant for these changes, but came about naturally.

3) With interrupts uninhibited again, an EDRUPT is used to vector to a routine in A. The EDRUPT is separated from its EXTEND with an INDEX, simulating an interrupt occurring after said INDEX. The interrupt routine checks to make sure BRUPT contains the correctly INDEXed instruction, and that ZRUPT is pointing to the instruction after the EDRUPT. Finally, BRUPT is replaced with a completely different instruction, and the code following the RESUME verifies that it was this instruction that got executed rather than the original.

With the changes included in this PR, all of these tests pass on both yaAGC and the hardware sim.

indy91:
Finally getting to this. I have implemented these changes and I noticed three things:

Firstly, in our NASSP version of the agc_engine.c file the OpCode variable was still used. This line:

`OpCode = Instruction & ~MASK12;`

Was commented out in the Virtual AGC version and now was removed with this pull request. OpCode wasn't used anywhere else. Anything bad that could result from removing this? Or, could this have previously caused isssues?

Secondly, this pull request sets SubstituteInstruction to 1. In the NASSP version there is a comment about this:

> // Remove ifdef because Luminary131 LM Autopilot code is using that feature

Reading in our forum it seems that the DAP in Luminary131 had unusually high RCS activity when this feature wasn't used. I remember that the Apollo 11 ascent stage wasn't controlled very well by the DAP. It had a hard time holding attitude at all. Today I for the first time used the ascent stage with a later Luminary version and I was surprised to see that it had no real issues controlling the ascent stage. Could the issues with Luminary99 have been caused by using this SubstituteInstruction feature?

And thirdly, this EDRUPT implementation sometimes causes the DSKY to freeze for a few seconds. When I load a scenario in NASSP I sometimes can't perform any inputs for mayb 20 seconds. Then all seems to be well again. I will check if I have implemented everything correctly, of course.

indy91:
The issue with the DSKY freezing is unrelated and was bug caused by me when I implemented the verb/noun flashing controlled by the Virtual AGC. So totally unrelated.

indy91:
Just did a perfect, automatic lunar landing with Apollo 15 in NASSP with these changes. I had the feeling there was more thruster activity, but nothing out of the ordinary. So at the very least this pull request doesn't break the lunar landing.

thewonderidiot:
The above questions were discussed in IRC, but for posterity:

1. OpCode was an unused variable, so it's safe to remove it.
2. The Luminary DAP code uses the BRUPT register in conjunction with EDRUPT to get out of its "jask". Without SubstituteIntruction = 1 enabled, an incorrect instruction would get executed every time the DAP exits, leading to presumably erratic results. This construct dates all the way back to Sundance, so all versions of Luminary would be equally affected.

thewonderidiot:
@rburkey2005, I've discussed this with @indy91 and @Thymo_, and they are on board with merging this in. It's been integrated into NASSP for a couple of weeks, now, and has been used in multiple landings (including the Apollo 15 landing video @indy91 made), as well as lunar ascent and most (?) phases of the CSM mission.

@jalex88, @indy91 says you've been flying a lot lately too -- have you seen anything out of the ordinary from the AGC over the past couple of weeks?

rburkey2005:
Sounds fine to me. It would be hard to ask for a more-thorough check. Thanks all!

PR #635, opened 2017-01-19
Update IMU_COMPENSATION_PACKAGE.agc

brf1:
Updated for Luminary 69. A few lines at the end of page 339 need checked against the binary due to garbled lines in the source printout. The illegible lines were assumed to be identical to the corresponding Luminary 99 code.


thewonderidiot:
Looks like all of the other changes were whitespace, so that sounds like a totally reasonable assumption. Thanks!

PR #646, opened 2017-01-22
SPS_BACK-UP_RCS_CONTROL

7ofNine:


thewonderidiot:
Thanks! I'm about 25% of the way through INTEGRATION INITIALIZATION, which has all been comments so far. So it should be done pretty shortly.

PR #739, opened 2017-01-23
Task535 corrections

7ofNine:
Here we are.  I have no binary differences to the binsource when assembled with two different Yul assemblers.


thewonderidiot:
Excellent, thanks! Nice work 😄

rburkey2005:
That was fast!

PR #740, opened 2017-01-24
Update KALMAN_FILTER.agc

brf1:
Ran through format option in yaYUL.


brf1:
I am also running the rest through the format option.

PR #757, opened 2017-01-31
yaAGC: Added heuristic loading of --parity -produced ROM files, and…

thewonderidiot:
… parity checking for them only. I packed the parities to try to keep the AGC State struct from growing too much.

I've tested both types of ROM files and they both work. Performing a V27N01E 5775E causes a hardware restart for ROMs produced with --parity, while using a default ROM, 00000 is successfully displayed.

Feedback is welcome. :)


thewonderidiot:
I've added --parity to yaYUL and removed the extra bank swapping that parity bits were causing in agc_engine_init. The banks are assumed to be always in the same 2, 3, 0, 1, 4, 5... order now, regardless of the presence of parity (i.e. --parity files are assumed rather than --hardware).

thewonderidiot:
Added --parity to oct2bin and converted unused words to Artemis072.binsource. @rburkey2005, if this looks good, I'll go ahead and convert the rest.

After these changes:
```
mstewart@thuban:~/agc/my_virtualagc/Artemis072$ export EXTRA_YAYUL_ARGS=--parity
mstewart@thuban:~/agc/my_virtualagc/Artemis072$ export OCT2BIN_ARGS=--parity
mstewart@thuban:~/agc/my_virtualagc/Artemis072$ make
../Tools/oct2bin --parity <Artemis072.binsource
mv oct2bin.bin Artemis072.bin
../yaYUL/yaYUL --unpound-page --parity MAIN.agc >MAIN.agc.lst
diff -s MAIN.agc.bin Artemis072.bin
Files MAIN.agc.bin and Artemis072.bin are identical
mv MAIN.agc.lst Artemis072.lst
mv MAIN.agc.symtab Artemis072.symtab
../yaYUL/yaYUL --html --parity MAIN.agc >/dev/null
mstewart@thuban:~/agc/my_virtualagc/Artemis072$
```

rburkey2005:
@thewonderidiot I guess in retrospect there's no problem with modifying the binsource files this way, because if there are other utilities that process binsource files, it could always be trivially passed back through sed or some equivalent tool before applying the other utility.  So it seems fine to me.

thewonderidiot:
@rburkey2005, I think this is ready. Full tree builds work with and without --parity applied to yaYUL and oct2bin. I also made bdiffhead display parity bits if either word being compared has a 1 for its parity bit, in case the difference detected is stemming from that. webb2burkey-rope still functions without changes since the parity bit just gets shifted out.

rburkey2005:
@thewonderidiot That's a good idea about bdiffhead.  Panic over, I guess.  Good work!  Do self-tests in (say) Aurora still work with parity enabled?

thewonderidiot:
After a momentary panic, before I realized I was accidentally running out of the tree I was using to analyze the TC trap thing.... yes, Aurora still passes just fine, as do the corrected Retread tests in Borealis, for both --parity and no --parity.

thewonderidiot:
@rburkey2005 Any final reservations before I merge this in?

rburkey2005:
Sorry, I thought it had been done.

PR #766, opened 2017-02-07
Update T6-RUPT-PROGRAMS.agc

Niklasva:
Issue #725 
Updated text + page numbers to match the Luminary 116 scans and ran it
through yaYUL with the reformatting option.


thewonderidiot:
This looks great, thanks! :smile:

PR #777, opened 2017-02-19
Task659 fresh start and restart

7ofNine:
I Think the reconstruction of the missing pages went well. I don't see any address shifts and on binary level it looks as if all detected differences are due to shifted label addresses and not opcodes.

I suggest to merge this part in but leave the task open as part of task#735 for the final code verification.


rburkey2005:
I guess I must have made some mistake diff'ing the octals, then. Not that I remember doing it.

thewonderidiot:
Sounds good. Thanks!

PR #783, opened 2017-02-28
Luminary116 Gimbal Lock Avoidance updated

f0n:
Made the mods to the gimbal lock avoidance. 
Removed the comments, but perhaps it was best to leave them on? Not sure. 
The code itself did not change from Luminary099 to Luminary116, jut the page number and the dates.


f0n:
Understood. I will update the file with the comments included.

rburkey2005:
@f0n Perhaps I should explain something that is probably not obvious.  There are two types of comments here:  The ones that begin with '#" are supposed to match what's in the page scans, 100%, and one-for-one.  The ones that begin with "##" (or "###") are "modern" comments that we've added to provide additional explanation, and don't match what's in the original printout in any way.  So the the ## comments in the file header explain what _we've_ done to create the file, as opposed to what the original developers did.  I suspect that little tid-bit is something we may have left out of the instructions.  If so, sorry about that!

f0n:
Perhaps this time it works.

Issue #622, opened 2017-01-15
Work with NASSP devs to unify agc_engine.c


thewonderidiot:
Right now NASSP and VirtualAGC have rather different versions of agc_engine.c. @indy91 and I were talking recently and agreed that it would be good to sit down and try to unify them. This issue is to track the work, to be done with him or any of the other NASSP devs. There's likely to be a lot of work on both ends here -- my prediction is that by the end we will end up adopting a lot of code that really should live in agc_engine.c. We can also look at how they've diverged and apply some lessons learned to improve the interface to future integrators of agc_engine.c.

I'm assigning this to myself for now but if anybody else feels particularly motivated, feel free. :)

PR #804, opened 2017-03-27
Moved a bunch of static things from agc_engine.c into agc_t

thewonderidiot:
@indy91, here are changes to move all of the things NASSP had made non-static, as well as the new DSKY channel 163 stuff, into agc_t. I've run the Aurora tests on it and everything still seems to be operational. Let me know how it works in NASSP. :smile:


indy91:
Unsurprisingly, adding a bunch of "State->" didn't cause any issues in NASSP. It's very nice to be able to remove all the "todo" comments from our agc_engine.c version! There are still a bunch of differences to our agc_engine, but this certainly helps with #622

Issue #629, opened 2017-01-18
TC Trap phasing is (probably?) incorrect


thewonderidiot:
I discovered something interesting while writing tests for Borealis that I'm tentatively (and hesitantly) calling an actual bug in the AGC hardware.

During my test of interrupt priorities and EDRUPT's ability to trigger interrupt routines, I need to wait for TIME3 to increment so the phasing of the timers is correct for the remainder of the test. Ed Smally had to do something very similar for his RUPTCHK:

```
                CA              S+ZERO                          
                TS              ZRUPT                           
                RELINT                                          
                AD              TIME4                           
                TS              SKEEP1                          
TENMS           CS              SKEEP1                          
                AD              TIME4                           # WAIT FOR NEXT TIME4 INCREMENT
                EXTEND                                          
                BZF             TENMS
```

Basically, he captures the value of TIME4 and then waits for it to change, using a subtraction and a BZF. This works absolutely fine on both VirtualAGC and the hardware sim. I found, though, that when I used _the exact same structure_ to wait for TIME3, I was occasionally getting TC Traps on VirtualAGC for not using any TCs or TCFs for too long. Without thinking too much about it, I inserted a NOOP (which assembles to TCF +1) and moved on.

Thinking about it later, though, I've become progressively more confused about why Ed didn't need any sort of TC/TCF in his 10ms loop. The documentation-stated behavior of the TC Trap alarm (which is backed up by the schematics) is that the time it takes to trigger a TC Trap varies from 5ms to 15ms -- so it certainly seems like his accurately-labeled TENMS loop should run into trouble here... and yet, it doesn't.

It's worth noting that EXTEND, despite really being a "TC 6", does NOT count for the TC Trap, as a TC never gets executed on the real hardware; I can back these claims up with the schematics, but basically what happens during an extend is as follows. The instruction code for the next instruction is completed being read in from memory during timepulse 7 of the last MCT of its preceding instruction. The parity pyramid, as a side effect, notices if the word loaded is exactly 00006, and if so (and it's phase 4 of timepulse 7), the extend flip-flop is immediately set. The next instruction sequence executed, then, is not TC, but STD2 (the standard stage 2 sequence, which simply takes care of loading the next instruction).

So what's going on? Why does Ed's code work? For reference, here is the circuit for the TC Trap alarm:
![tctrap_circuit](https://cloud.githubusercontent.com/assets/94056/22053094/5fba6b34-dd03-11e6-9f66-aea27ac9417d.png)
[...]

PR #805, opened 2017-04-03
Simulation of the TC Trap bug, and corrected phasings

thewonderidiot:
This pull request contains two separate, though related, fidelity improvements to yaAGC:

1. The phasings of the standby PRO button circuit and all five timers were incorrect relative to the scaler (though the periods of all and the phasing of the timers relative to each other were correct). After these changes, all things clocked from the scaler are phased correctly as per the schematics. However, without simulation of the possible TC Trap hardware bug described in #629, these corrected phasings render Aurora and Sunburst unable to pass their self-tests.
2. The three primary causes of said TC Trap bug are now simulated. They are:
    * BZF being taken to an instruction other than EXTEND
    * CS, XCH, TS, and DXCH being executed at all
    * RELINT or INHINT executing after a BZF/BZMF

It is my hope that the circuit descriptions and timing diagrams contained in ND-1021042 and/or ND-1021043 will prove the hardware sim's timings either correct or incorrect. If we can show with the manuals that the timings of FS10/F10A/F10B and the TC Trap are correct, then I think it's safe to say that the bug must have been present, since I think it is safe to assume that Aurora and Sunburst could in fact pass their tests.

Anyways, let's leave this open until we get scans of those manuals.


thewonderidiot:
I forgot to mention, but with these changes all of the Aurora/Retread self-tests still pass, as well as the Borealis extensions and hardware alarm tests. It's not clear to me whether or not this will impact the execution of mission software.

thewonderidiot:
We've finally got scans of ND-1021042, so we can begin to make some conclusions!

There's two key waveform illustrations relating to this issue. The first is the one dedicated to the TC Trap circuits:
tc_trap_timing

This diagram, and the text preceding it, confirm that the alarm starts monitoring at the F10B pulse, and finishes at the F10A pulse. If no TCs or TCFs have occurred in during that time, a restart signal will be generated synchronously with F10A.

The other relevant waveform diagram is of Sync and Timing signals. It shows the timer pulses on the same diagram as F10A and F10B:
timer_pulse_phases

The TIME4 increment request pulse T4P can be seen to increment halfway between F10A and F10B, outside of the window of the TC Trap.

So just to be doubly sure, here's the SELF-CHECK code from Sunburst 120 again:
```
RUPTCHK         INHINT                                          
                CA              S+ZERO                          
                TS              ZRUPT                           
                RELINT                                          
                AD              TIME4                           
                TS              SKEEP1                          
TENMS           CS              SKEEP1                          
                AD              TIME4                           # WAIT FOR NEXT TIME4 INCREMENT
                EXTEND                                          
                BZF             TENMS                           
                INHINT                                          
                CA              ZRUPT                           
                EXTEND                                          
                BZF             +2                              # NO INTERRUPT.
                TC              RUPTCHK                         # THERE WAS AN INTERRUPT. START AGAIN.
                CAF             S+1                             # 2 1/2 MS UNTILE NEXT T3 INTERRUPT.
```

The logic is basically as follows:

1. Positive zero is placed into ZRUPT.
2. The current TIME4 value is placed into SKEEP1.
3. SKEEP1 is subtracted from TIME4.
4. If the result is zero, go to (3). This always happens unless TIME4 has just incremented.
5. If ZRUPT is still zero, exit the loop. This is possible if _and only if_ no interrupts have occurred, as the comment indicates. The hardware automatically populates ZRUPT with the current program counter value every time an interrupt sequence occurs.
6. ZRUPT was not zero, so an interrupt occurred. Go to (1).

So let's start at (4), with a TIME4 increment, and assuming that some time during the loop an interrupt occurred.
ruptchk_trap

The TC Trap starts monitoring 2.5ms after the check in step (5) branches back to step (1). 5ms later, if no TCs have occurred, a TC Trap alarm will happen. The loop in steps (3) and (4) does not contain either; it consists solely of the instructions CS, AD, EXTEND, and BZF. The EXTEND doesn't count, because despite its coding, it's executed as a special case of the STD2 subinstruction (this behavior is confirmed on page 4-365 of ND-1021042). So during this window, the only way for a TC or TCF to have been executed is for an interrupt to have occurred. But if an interrupt is always guaranteed to occur here, then SELF-CHECK could have never passed!

It's also worth noting that the schematics for the TC trap presented in ND-1021042 match exactly with the proper schematics we have for module A13, so we're not missing any signals here:
tc_trap_circuit


So, since ND-1021042 has verified all of my findings related to the timing, I'm considering this bug all but confirmed. I simply can't think of any way SELF-CHECK could have ever worked without it.

I'm therefore voting that we move forward with merging these changes in. Thoughts? Objections?

thewonderidiot:
Cool. :smile: It's been on my TODO list for a while to update the EDRUPT section on the website, and add a section for the hardware alarms. That's been on the backburner for a while now, so I'll try to get to that sooner rather than later.

PR #806, opened 2017-05-12
Moved some special cases from WriteIO to CpuWriteIO

thewonderidiot:
This makes WriteIO safe for external callers to write to certain I/O channels. It represents a large step towards #622, removing the need for one of the last major differences between NASSP's agc_engine.c and our own.

@indy91 and I, after a few conversations, finally got to the bottom of why NASSP seemed to need extra logic added into their copy of agc_engine.c to handle channel 33, despite the schematics implying that yaAGC's behavior was correct. The answer lied in the distinction between WriteIO and CpuWriteIO. For the most part, CpuWriteIO is called by yaAGC internally whenever it writes to an I/O channel, while WriteIO can be called by external interfaces to supply in input value (SocketAPI does this in yaAGC). A few channels have special behaviors when written to by the running program (including channel 33); however, this logic was living in WriteIO instead of CpuWriteIO, meaning an external event on that channel would _also_ trigger the special behavior of the channel.

These changes move those special behaviors into CpuWriteIO, to stop yaAGC from thinking that a program has written to a channel when it hasn't. @indy91, this seems like it's working from what I can see, but it's harder for me to test this sort of change. Would you mind checking this out to see if it works with your branch? 😄


indy91:
This part of the code was deleted:
`// Don't allow the AGC warning input to be reset if the light
	  // is still on
	  if (State->WarningFilter > WARNING_FILTER_THRESHOLD)
	    State->InputChannel[033] &= 057777;`

Was that on purpose?

Also, this does not properly work yet. When I monitor channel 33 it now always appears as 76000. If I move the CMC optics switch fast enough I can change it temporarily to 76030, but soon after it will revert to 76000 again. Could that be caused by:

` State->InputChannel[Address] |= 076000;

in CpuWriteIO?`

thewonderidiot:
Nope, that wasn't on purpose, good catch!

And no, it was being caused by the subsequent call to WriteIO. The value written to Channel 33 was being allowed through to WriteIO, which was letting the program also change bits 1-10. That should be fixed now.

thewonderidiot:
Per @indy91, this has survived the landing test and appears to be working correctly. I'm going to merge this in.

PR #887, opened 2017-05-27
ASSEMBLY_AND_OPERATION_INFORMATION

7ofNine:


Sunburst37/ASSEMBLY_AND_OPERATION_INFORMATION.agc
```diff
@@ -84,20 +70,20 @@
...
 # MISSION PHASE 11 - DPS2/FITH/APS1
 # MISSION PHASE 13 - APS2
-# MISSION PHASE 16 - RCS COLD SOAK
+# MISSION PHASE 16 - RCS COLO SOAK
 # INTEGRATION INITIALIZATION
 # ORBITAL INTEGRATION PROGRAM
 # LMP COMMAND ROUTINES
-# AS206 MISSION SCHEDULING PACKAGE
-# 206 SERVICE ROUTINES
+# AS206 MISSION SCHEOULING PACKAGE
```
thewonderidiot: Hmmm... the PDF really does make this D and the one in "COLD SOAK" above look like Os. But in the high-resolution images provided by archive.org, they're very clearly D's. I think PDF compression might be fighting against us here. @rburkey2005, I'm not sure the PDF is good enough quality for transcription. We should probably switch back to the reduced quality image files.

thewonderidiot: For comparison, here's the PDF on the left, and JPEG2000 image on the right:


comparison

rburkey2005:
I agree that the PDF isn't very good, and think that archive.org's online viewer should probably be used instead, since a lot of computers can't massively view JPEG2000. However, I would be stunned if that ocr process was changing O into D or vice versa. I think something else must be at work in that example.

rburkey2005:
More likely the wavelet transformation, or whatever JPEG2000 uses for image encoding, is doing that.

7ofNine:
I'll fix it. There are many more cases in it but I started to ignore them and used the expected D. I also saw some E that looked like an L in the pdf, might be another case of unintended morphing. For the time being I'll use the high resolution one on archive.org and use up some of my bandwidth.

7ofNine:
image
Is another example. In the archive version it clearly reads as 'CHANNEL TO BE SPECIFIED'

thewonderidiot:
You mean, something in how JPEG2000s  are encoded that is getting confused in the translation process to PDFs? The J2Ks are fine, the PDF is showing incorrect letters. And I agree to an extent, but having offline images is useful when you're not always online. I always end up either using the low resolution images or converting the high resolution ones from J2K to JPG myself.

I can put together an image set if you don't want to, I guess.

thewonderidiot:
Thanks for the fixes @7ofNine! I'll merge this in when I get to a real computer. Sorry for the confusion!

rburkey2005:
The way you get lower quality images in JPEG2000 is to discard the "high frequency" components. But the transformation isn't in terms of sines and cosines, so I suppose that if you discard components in just the wrong way, you may end up with different shapes than you started with. Thus, D in place of O or L in place of E. Or, I just don't know what I'm talking about. But I'm sure it's not the ocr.

If you think the images are useful, I may as well make them. I'd have to upload them myself anyway

rburkey2005:
Okay, I've shredded that PDF (well ... it's still at archive.org, but there's nothing I can do about that other than complain), and instead there's  [a directory of JPGs]( http://www.ibiblio.org/apollo/ScansForConversion/Sunburst37/).  The JPG definitely says "COLD SOAK" (or rather, "COLD SOA<" :smile:) on page 2.

7ofNine:
looks much better with the new jpegs
image

thewonderidiot:
Thanks Ron! And thanks @7ofNine! :smile:

PR #893, opened 2017-05-29
FRESH_START_AND_RESTART

7ofNine:
All occurrences of POOH and family are spelled with 'OO' not with '00'. See symbol table listing p.1063


thewonderidiot:
Thanks!

PR #906, opened 2017-06-01
Task837 q r axes reaction control system autopilot

7ofNine:


7ofNine:
I just looked at the generated code for this section and I discovered a SBANK issue for 2CADR.
IS there any information that the used Yul assembler version wasn't fully converted to AGC (instead of BLK2), yet?
I see that the very first 2CADR pseudo operation gets assembled to
image

i.e 40006, 
while in Sunburst/120 it is assembled  to
image
i.e. 40066 but in both cases the label FILDUMMY is in bank 20 meaning below the LOWSUPER address (which is not defined in sunburst 37).

This is just a heads-up. It could still be that a not yet transcribed part in front of this section affects it but
I had to change my JaYul assembler to generate the above pattern which corresponds to what is expected in BLK2.
There are still too many binary comparison errors (> 900) to draw any conclusion.

rburkey2005:
I confess I don't know.  As I guess you're pointing out, this is from the time period when the BLK2 and AGC targets overlapped (Aurora 12 from 11/1966 was BLK2 and Sunburst 37 from just a month later was AGC), so it certainly wouldn't stretch the imagination if the implementation of the AGC target in YUL was still ongoing.  Whereas Sunburst 120 was about a year later, when the process would undoubtedly have been complete.

7ofNine:
The phenomenon might have been caused by an incomplete implementation on my side (JaYul). 
I discovered that I couldn't assemble Sunburst 120 correctly (pretty sure it worked once) and that I had some issues with it. Making Sunburst 120 assemble corrected the generation for sunburst 37, too (for this specific case at least).
Currently I can not verify it with yaYul. I think there is a bug in yaYul if there are multiply defined labels.
In the current state of the source (pretty messy) it exceeds the maximum number of passes and doesn't create a listing at all because it never goes into the the pass with WriteOutput=1. In every pass numSymbolsReassigned remains positive and it skips the above branch. Fine with me but when it exits it should at least execute a single write pass, so one can actually see what went wrong.
Something like 
((k == 0 || k >= LastUnresolved) && numSymbolsReassigned == 0) || i == maxPasses)
could do it.
When I make it work (see above) the result from yaYul is still different from the expected value. But in this state it is impossible to really compare the results because of somewhat different logic in the two assemblers.
I'll keep you posted.

thewonderidiot:
Sounds good. Thanks!

PR #938, opened 2017-06-14
A fix that corrects more than 300 binary comparison failures. A shift…

7ofNine:
… of 1 word in block 2


thewonderidiot:
Awesome, thanks!

PR #941, opened 2017-06-15
Fix typos 3

7ofNine:
Fix mainly my own typos.

FYI: there are many errors with regard to 2CADR/BBCON and the SBANK bits

### DIFF
--- a/Sunburst37/AS206_MISSION_SCHEDULING_PACKAGE.agc
+++ b/Sunburst37/AS206_MISSION_SCHEDULING_PACKAGE.agc
@@ -19,6 +19,9 @@
 ## Mod history: 2017-05-24 MAS  Created from Sunburst 120.
 ##              2017-06-07 MAS  Updated for Sunburst 37. There's a decent number
 ##                              of differences.
+##              2017-06-15 HG   Fix page number 814 -> 762
+##                              Add missing instruction TCF ENDUP
+##                              remove operand modifier CHKUPDEX +1 -> CHKUPDEX
 
 ## Page 748
 #          THE FOLLOWING ROUTINES IMPLEMENT THE MISSION SCHEDULING LOGIC AS DESCRIBED IN CHAPTER 4 OF THE
@@ -313,6 +316,7 @@ DOV70           TC              MTIMERUP                        # VERB 70
 
 DOV72           TC              MTIMERUP                        # VERB 72
 DOV71           TC              MPHASEUP                        # VERB 71
+                TCF             ENDUP
 
 #    *** C ODING  TO BE INSERTED HERE TO CLEAR OUT ALL TIMER/PHASE PAIRS
 #        W HICH H AVE BEEN SET BY LGC, SINCE GROUND REQUEST WILL SUPERSEDE
@@ -490,7 +494,7 @@ MGETUP          TC              INTPRET                         # MAKE SURE THIS
                 EXTEND
                 BZF             UPERROR
 
-                TC              CHKUPDEX        +1
+                TC              CHKUPDEX
                 CA              MPAC                            # CONTAINS DT IN SECONDS.
 
                 INDEX           RUPTREG1
@@ -565,7 +569,7 @@ MTABLE1         OCT             20010                           # MP7
 
 MTABLE          EQUALS          MTABLE1         -21D            # MP 1-6 NOT ACTUALLY INCLUDED IN TABLE.
 
-## Page 814
+## Page 762
 BADPHASE        TC              ALARM                           # ALARM WHEN MPHASE COMES DUE BUT MPHASE
 
                 OCT             00601                           # REGISTER IS ZERO (-0 MEANS INACTIVE).
--- a/Sunburst37/ASCENT_STEERING.agc
+++ b/Sunburst37/ASCENT_STEERING.agc
@@ -18,6 +18,7 @@
 ## Website:     www.ibiblio.org/apollo/index.html
 ## Mod history: 2017-05-24 MAS  Created from Sunburst 120.
 ##              2017-06-11 HG   Transcribed
+##              2017-06-15 HG   Fix value T2 2DEC 200 B-17 -> 2DEC 200
 
 ## Page 853
 #    PROGRAM NAME--ASCENT        BY--BERMAN
@@ -476,7 +477,7 @@ DP2/3H          2DEC            .666666667
 
 DTASC           2DEC            200             B-17
 
-T2              2DEC            200             B-17
+T2              2DEC            200
 
 LAMBSET         TC              INTPRET                 # CALL FOR LAMBERT SUBR.
                 CALL
--- a/Sunburst37/DAP_INTERFACE_SUBROUTINES.agc
+++ b/Sunburst37/DAP_INTERFACE_SUBROUTINES.agc
@@ -18,6 +18,7 @@
 ## Website:     www.ibiblio.org/apollo/index.html
 ## Mod history: 2017-05-24 MAS  Created from Sunburst 120.
 ##              2017-05-30 HG   Transcribed
+##              2017-06-15 HG   Fix operand XCH  -> TS
 
 ## Page 453
                 BANK            16
@@ -121,7 +122,7 @@ HOLDRATE        CAF             EBANK6
                 DCA             OMEGAP
                 DXCH            OMEGAPD
                 CAE             OMEGAR
-                XCH             OMEGARD
+                TS              OMEGARD
 
 COMNEXIT        EXTEND
                 DCA             CDUY
--- a/Sunburst37/EXTENDED_VERBS.agc
+++ b/Sunburst37/EXTENDED_VERBS.agc
@@ -18,6 +18,7 @@
 ## Website:     www.ibiblio.org/apollo/index.html
 ## Mod history: 2017-05-24 MAS  Created from Sunburst 120.
 ##		2017-06-01 RSB	Transcribed.
+##              2017-06-15 HG   Fix operand BANKCALL -> INTPRET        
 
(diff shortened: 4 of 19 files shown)


thewonderidiot:
Thanks! Yeah, I was looking at all those SBANK errors last night. I think I might need to make another attempt to understand what YUL was doing there... should be a bit easier now that Jim has transcribed and consolidated all of the Honeywell 1800 stuff. :smile:

thewonderidiot:
And wow, you weren't kidding. 241 of the differences are filtered out if you pass --no-super to bdiffhead.

7ofNine:
I have a wild and unproven guess of what it might be doing for the the SBIT, which is partially based on my implementation which is somewhat different from the yaYul implementation because of its single pass assembly.
a) the SBIT is originally kept at 00 until
b) a first reference is made to an address located in one of the super banks (i.e. >=30) and then it is set to the corresponding pattern. For Sunburst 37 it should always be 06. No other superbank is used. Looking at the output of yaYul there are many cases where 10 is used, which should not be the case at all.

The question is what does "first reference" mean? That is where I guess that it is in relation of the over all lowest pseudo address being allocated as in comparison to the actual assembly sequence. That could also go with the descriptions in YUL from where I take that the address resolution was at least a two stage process. The first pass could allocate all the memory (and hence know all the pseudo addresses) and a second pass would be able to finalize the data by just going through the missing references and resolve them. Similar what a modern linker does. Just an idea and might be completely off the rail.

thewonderidiot:
I strongly suspect you're right about the two-pass thing. When we were first talking to Hugh Blair-Smith about attempting to run YUL on an H1800 emulator, he gave us a list of gotchas and things to look out for. One of the things he mentioned was this:

> 5. The feature I was most unsatisfied with was SBANK=, a declaration which supported diagnostics that noticed whether references to superbank locations were inconsistent with the latest such declaration. If it seems to be raising false alarms, just ignore or disable them, and the object code should still be correct. I don’t know whether the GAP folks improved its function.

So it really sounds like YUL at least didn't actually *do* anything with SBANK= other than generate warnings if an SBANK= declaration you've made doesn't match what YUL determined the correct answer to be.

7ofNine:
I have just verified a completely naive version of handling the SBIT. There is no SBANK= operator in SUNBURST37(!). Forget everything that has been implemented so far. 
If the operand of a BBCON/2CADR (and BBCON*) operator refers to an address with FB < 030 (i.e. no superbit defined) set it to 00, if the address refers to an address with FB >= 030 and FB < 040 set it to 06, and if  it refers to an address with FB >= 040 (does not exist in this source) set it to 10. 
This works for me in Sunburst37 and could be a basis for later versions where there are SBANK= operators.

thewonderidiot:
@7ofNine I spent some time tonight scrolling through Luminary 210. It's possible to see what GAP thinks the "current" SBANK is by looking at the `En Sm` printed under the page number on every page. `m` will be one of `?`, `3`, or `4`. It's not the best granularity, since you can't see per instruction what the SBANK is, but it seems good enough.

As far as I can tell, the SBANK is totally naive, as you described -- it changes depending on the superbank of the last address used, switching to 3 when any address in banks 30-37 is assembled, and to 4 when anything in 40-43 is. Any 2CADRs/BBCONs/etc. appear to just use the value displayed by `Sm`, unless those bits are actually needed. In that case, if the needed value doesn't match the current `Sm`, an SBANK= directive is used to stop warnings.

It looks like generally, once the SBANK has been established, it almost never goes back to `S?`. The one case of that I've seen so far was on a page where an SBANK= directive was the last line of the preceding page -- so maybe it wasn't sure if it was supposed to be a one-shot or not yet?

Anyways, I'll try fiddling around with this in yaYUL tomorrow and see what happens.

thewonderidiot:
A couple more notes: non-oneshot SBANK='s _do_ change the `Sm`. Also, what I'm seeing doesn't match what you saw for FB < 030; it looks like GAP simply fills the superbank bits with the current `Sm` rather than setting them to 000. Maybe it's possible that YUL and GAP behaved differently here?

rburkey2005:
This discussion is so 2003.  :smile:   How many times did I think I grasped the pattern?   How many times ultimately disappointed?   But I'm glad you guys are still thinking about it.  I certainly never noticed the En Sm notation, and it would have been a great help if I had.

7ofNine:
It is somewhat disappointing. What I did for Sunburst 37 works fine. But then I tried it on Sunburst120 (after taking out all the additional SBANK='s) and it didn't work there. In some cases the lower superbank (06) pattern is set even if the referenced address is nowhere near those banks. For me that points to an undiscovered history of GAP/YUL. More option switches for yaYul.
e.g. here is what YUL generates
image
image
and here is what the method above generates for Sunburst120
003577,000232:    4173  E3,1400                                 EBANK=   LST1                          	#
003578,000233:    4173          02634 14003   ADRPCHN2          2CADR    PHSCHNG2                      	#
So far I have no idea why the lower superbit shows up.

rburkey2005:
Since the values of these bits hardly ever matter, I've wondered from time to time if YUL or GAP were even consistent in the way they generated them from one time to the next?  Perhaps if you assembled the same program a second time, YUL/GAP would have generated different superbits.  :smile:

I suppose we'll be able to find that out soon when we get our copy of AP11ROPE, though the fact that the checksums for it are identical to LMY99 suggests that the generation of the superbits probably was consistent as well.

7ofNine:
It's probably more of historic interest to find out what these assemblers where doing than of a practical use.
I am quite sure that nobody will use them for a flight to Mars :-)  and one can always create something (obviously) that will reproduce the expected results.
Nonetheless I looked into the listings on archive and I see that on the first page of the listing is a line printed one could interpret as some kind of name for the YUL assembler.
For Sunburst120 it says : YUL  2NDFROZE   and the date is 1967-12-06
For Subburst37   it says : YUL  NANSTONE  and the date is 1966-12-13
and interestingly 
for Aurora 12      it says : YUL NANSTONE  and the date is 1966-11-10
The YUL listing on archive is from 1967-12-12 which would correspond to 2NDFROZE. 
And if one looks at the ends of the single passes one finds some revision numbers:
Pass 0: Revision 19
Pass 1: Revision 28
Pass 2: Revision 27
Pass 3: Revision 10
One has a good chance that there were code affecting changes over a time span of one year.
Should be "easy" to find,  just get an older listing (8-o .

rburkey2005:
Isn't NANSTONE (and presumably 2NDFROZE) the name of the computer on which YUL was running?  But I suppose there probably is some correlation between the computer names and the dates.

7ofNine:
Possibly. You know more about that than I do. Just trying to find clues, CSI:AGC.
Looking at the En Sm tracking in Luminary it looks as if a SETLOC affects the value for Sm, i.e. it creates some tracking using the local program counter, which for me is not the case in Sunburst37 and if it is tracked somehow in Suuburst120 I haven't seen any SETLOC in there before the first occurrence of the low super bit in the 2CADR above.
The saga continues, but I won't loose sleep over it. It's a challenge, but we might not have enough samples yet to determine the inner structure of the machine. Norbert Wiener proofed that one can do it but it might take a little bit longer.

rburkey2005:
Strange coincidence.  It is has probably been decades since I thought of the name Norbert Wiener, but I briefly recollected his book Cybernetics a few days ago.  (Only to realize I couldn't remember anything he had said in the book. :smile:)

thewonderidiot:
Alright well, I scrolled through all of Luminary 210 tonight. I also took another look at Sunburst 37, Aurora 12, and Sunburst 120 -- and it turns out there's a very noticeable difference between Sunburst 37/Aurora and Sunburst 120! Around the time of the former two, `Sm` is never, ever printed in the top right. So it seems like YUL had not yet developed the idea of an "established" superbank. I think Aurora, Sunburst 37, and probably Retread all obey the rules you set out above and confirmed on Sunburst 37, @7ofNine. I also think the rules in my comment work for Sunburst 120 and later. Mods to yaYUL are in progress and I should have them going tomorrow. But it's

P.S. one small addition to my rules above: SETLOC and BANK do not affect the established superbank. Only when a word is actually assembled into that new bank is the established superbank updated. There's one point where Luminary 210 BANKs and SETLOCs into superbank 4, and then changes to a a low (< 30) bank without assembling anything into the superbank. The established superbank remains 3 throughout the page and onto the next.

P.P.S. the copy of YUL we have has all of the superbank logic, so at this point I'd put money on it being able to build Sunburst 120 and later, but actually giving us incorrect assemblies for Aurora, Sunburst 37, and Retread (although Retread might not be effected if it doesn't go into bank 30+ at all).

thewonderidiot:
Good news! I just successfully assembled Luminary 210 with all SBANK= workarounds removed. Now to go on an SBANK= killing crusade to try it out on other programs. :smile:

thewonderidiot:
Success! :smile: All programs are able to assemble correctly with all SBANK= workarounds globally removed, _except_ for Sunburst 37. Aurora 12 and Retread don't use banks 30+, so they still assemble correctly even with the later YUL behavior. Sunburst 37 is the only one we've got that uses those banks before the behavior was changed..

rburkey2005:
Good job!


rburkey2005:
@thewonderidiot Okay, I'm misunderstanding something crucial here.  I pulled everything from the repo and built it as-is ... that works.  Then I started checking all of the SBANK=  pseudo-ops to see that the ones that had been added as workarounds had been removed, and none has been accidentally missed.  But COLOSSUS 237 was full of them, so I removed them and tried to build ... it's full of mismatches.  What is it I'm missing?

rburkey2005:
@thewonderidiot Never mind, I didn't see that the changes hadn't been fully merged.

PR #942, opened 2017-06-15
Some few more typos

7ofNine:


thewonderidiot:
Thanks! Looks like we're pretty close to having only SBANK= foolishness and corresponding banksum errors.

PR #944, opened 2017-06-17
yaYUL: Refactored superbank logic and added --early-sbank.

thewonderidiot:
This removes the need for every single SBANK= workaround we have! I'm still not sure how YUL handled cussing; Luminary 210 does a couple of things that seem like they would warrant a cuss in YUL, so perhaps GAP was more lenient. I've just made yaYUL permissive for now as well.

This is just the changes to yaYUL itself. The SBANK= removal needs cleanup (header updating) and I'll push all of that to this pull request as I go.


thewonderidiot:
The check will stop failing after I've removed all the workarounds, I promise. :smile:

thewonderidiot:
Alright, I think that should do it :smile:

thewonderidiot:
I've also added --pos-checksums. Calculation of the checksum for BLK2 targets and any AGC targets prior to Sunburst 116 was slightly different, in that the sum being equal to the negative of the bank number was _not_ yet valid. This was when the requirement for two TC SELFs was dropped, and apparently their presence guaranteed that getting the positive result was always possible.

Since this is always the case for BLK2, I've also made --blk2 imply --pos-checksums, so Aurora only really needs --blk2 to assemble correctly now.

PR #946, opened 2017-07-14
yaAGC: Added simulation of HANDRUPT and its three traps

thewonderidiot:
HANDRUPT is generated by the AGC internally, based on the enable status of three traps (Trap 31A, Trap 31B, and Trap 32), and the states of various bits in channels 31 and 32. GOJAMs reset the states of the traps. All of the traps automatically reset themselves upon generating a HANDRUPT. 

Interestingly, there's no read circuitry for only the three trap enable bits in channel 13, so any attempt to read the channel will show those bits as 0, regardless of the states of the traps. I've simulated this by masking out those bits in CpuWriteIO(), shortly after setting the trap enables.

@indy91, it's all yours for testing. :smile:


indy91:
LPD commands have improved significantly when done with a joystick. Just as we thought. Thanks!

Issue #1048, opened 2017-09-03
Improve Z handling


thewonderidiot:
Z is currently treated as a 12-bit register. This is for the most part accurate, even though the register is actually 16 bits long. Those upper four bits are occasionally used for other things -- for example, during DV, if the quotient or divisor are negative, one or both of Z's sign bits can bet set in order to keep track of the signs of the dividend and remainder.

I noticed while working #1047 that Z16 is set (if it is going to be set) before the divisor value is read in during DV, meaning if you (for some unfathomable reason) were to do `DV Z`, and your quotient is negative, then your Z is going to have an unexpected sign bit set and totally change the results. That's fine, and will be handled in the updated DV code. However, I also noticed that none of the DV sequences explicitly clear these bits.

Instead, the upper four bits of Z are lost during instruction changeover. The standard stage 2 sequence, STD2, increments Z via the control pulses `RZ WY12 CI`. This means, "read the lower 12 bits of Z into one input of the adder, clear the other input, and feed a carry in", effecting Z+1 on the adder output for later updating of Z. STD2 happens after _almost_ every instruction, so while it's possible to store things in the upper 4 bits of Z, they are pretty much always immediately lost.

Not all instructions use STD2 however -- for various reasons, sometimes instruction sequences contain their own control pulses for loading the next instruction. The other sequences that can do this are:
* TC0 -- uses `RB WY12 CI`. B is holding a copy of the instruction at this point.
* TCF0 -- same as above.
* NDX1 -- uses `RZ WY12 CI`. This needs to do its own loading of the next instruction so it can modify it.
* NDXX1 -- same as above.
* RUPT1 -- uses `RRPA WZ`. Interrupts' versions of Z are handled by the interrupt priority chain.
* MP3 -- uses `RZ WY12 CI`. This is done just as an optimization, since the multiplication math only needs a few of the timepluses of this sequence to complete. Doing it here saves an MCT.
* RSM3 -- uses `RG WZ`. This is the last sequence executed during `RESUME`. It loads ZRUPT directly out of memory and writes it into Z.

Note that that last one is a bit different than all the others -- it writes ZRUPT (which can arbitrarily be written to by software) directly into Z, without filtering down to the lower 12 bits like all the other instructions. This means that the instruction stored in BRUPT, whatever it is, can actually see the upper four bits of Z set to 1s. And I'm pretty sure the instructions following RESUMES are the _only_ time this is the case.

[...]

PR #1049, opened 2017-09-04
Improved handling of Z and ZRUPT

thewonderidiot:
This implements the quirks of Z/ZRUPT as described in #1048, and includes new tests in Borealis to exercise them. yaAGC seems happy, but I'd like to run the new Borealis on hardware before merging this in, to make sure everything works as I expect.


thewonderidiot:
Oh, this also includes setting c(Q) to c(Z) on a restart, since the TC 4000 a GOJAM forces has that effect. The RSBBQ erasables should be properly updated on a restart now.

thewonderidiot:
I rewrote the Borealis test for the ZRUPT case a bit, to simply INHINT...RESUME outside of an interrupt, rather than using EDRUPT to get inside one. This saves some precious fixed-fixed memory words, and makes sure RESUME can properly work whenever you want, not only inside interrupts.

thewonderidiot:
The Borealis tests have been verified running on my FPGA, so I think these changes are good. Merging them in.

Issue #1047, opened 2017-08-30
Characterize and emulate 'total nonsense' in yaAGC


thewonderidiot:
@indy91 and I found an interesting snippet of code in Zerlina's P66 while working through the scalings to use for all of its padloaded values:
```
EXTEND                                  # IS |VDGVERT - VVECTX| < VERCRIT?
DCA             MPAC
EXTEND
DV              VBUF            +1
EXTEND
BZF             ERRFAIL
```
Leading into this, (VDGVERT - VVECTX) is in MPAC,MPAC+1, and VERCRIT is in VBUF+1. The purpose here is to decide, based on the magnitude of (VDGVERT - VVECTX), whether to use TAURODB or TAURODL for the ROD computation.

This is an interesting sort of test that I haven't seen before, and it looks like it might run into trouble in yaAGC. If (VDGVERT - VVECTX) is >= VERCRIT here, then we run into a situation where the dividend is greater than the divisor. According to Savage&Drake, this results in "total nonsense", and yaAGC presently returns random numbers in both A and L. However, if I am reading this code correctly, it looks like Don is assuming (and testing that) A receives +0 or -0 in such situations.

We haven't yet verified whether or not this is actually happening, but if such tests are done it would be a good idea to figure out exactly what 'total nonsense' entails and properly emulate it.

rburkey2005:
As usual, I didn't try to deduce what the result would be when I handled this case, and if I'm not mistaken, literally return a random value using a random number generator.  :smile:  So I wouldn't be surprised if you're right, and if this isn't another mystery for you to verify the correct behavior of.

thewonderidiot:
Well, we just ran a test with a lightly modified version of Zerlina that saved off the value (VDGVERT - VVECTX) for inspection. The typical value for that was 00000,XXXXX or 77777,XXXXX, getting sometimes up to 00004,XXXXX if multiple ROD clicks are done at once. The value currently being used for VERCRIT is 11272, so the division is quite safe, contrary to my initial belief. So luckily we don't need to handle nonsense for Zerlina to work. :)

I think I am still going to figure out what goes on in such cases, and add support for it to yaAGC and some sample cases to Borealis, just for accuracy's sake (and to nip any future questions like this in the bud).

rburkey2005:
[...]

PR #1050, opened 2017-09-05
yaAGC: Added simulation of "total nonsense" in DV and other unexpected cases

thewonderidiot:
This implements a streamlined version of the code presented in #1047, plus implementation of the various other weirdnesses associated with DV discussed there.
* All cases where the divisor is smaller than the dividend (which result in "total nonsense") should now match results gotten on the original hardware.
* DV A, DV L, and DV Q all have special handling now to match "unexpected" modifications that happen to those registers before the divisor is sampled. All such cases I've tried match the hardware sim.
* Divisors with positive or negative overflow are also now handled by the new DV register simulation, since the overflow totally changes the results. Again, all the cases I have tried with overflow match the hardware sim.

All of the existing Aurora and Retread tests still pass, so I think at this point (unless there is some particularly weird case I haven't thought of) yaAGC should always get identical results to the hardware sim for all dividends, divisors, and registers.

I haven't yet written Borealis tests for this, but my intention is to add a whole bunch of cases for all permutations of sign, register, overflow, and nonsense-ness. I'll add those to this pull request when I'm done writing them.

### DIFF
--- a/Borealis/AGC_BLOCK_TWO_EXTENDED_TESTS.agc
+++ b/Borealis/AGC_BLOCK_TWO_EXTENDED_TESTS.agc
@@ -19,6 +19,11 @@
 ##              2017-09-04 MAS  Rewrote the new ZRUPT test to simply RESUME outside of an
 ##                              interrupt instead of using EDRUPT. This saves a few words,
 ##                              and exercises RESUME's ability to work whenever.
+##              2017-09-17 MAS  Added a new table-driven extended test for DV, which checks
+##                              out "nonsense" (as described by E-2052), overflow cases, and
+##                              and division by A and L. The L cases are not yet written,
+##                              and I intend to add a separate case for DV Z when I figure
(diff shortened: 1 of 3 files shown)


thewonderidiot:
I've added the first big group of DV test cases. So far, yaAGC is passing everything I throw at it. The tests so far cover nonsense, overflow, `DV A`, and all permutations thereof. The table can also handle `DV L` tests, which I will throw in tomorrow. There's four `DV Z` tests I want to do as well, but that is harder with Z being the program counter, so I need to figure out how to do that.

thewonderidiot:
All tests I originally planned are now captured! yaAGC passed them all with flying colors. I also made it possible to express division by 0 or -0 in the table, and added a case for each. To wrap it all up, I gave the final Borealis build a run on my FPGA, and no errors cropped up -- so yaAGC should be (at least as far as I've tested so far) completely consistent with it for all DV cases.

PR #1053, opened 2017-10-11
Proofed comments for FP6 and modernized it, plus some yaLEMAP features

thewonderidiot:
While looking through the FP6 scans the other day I noticed a couple of differences from the transcribed version, so I decided to do a proofing pass. I focused _mostly_ on the comments for this one, although I did find some "code" differences (the scaling on zero constants, which obviously changed the assembly not at all). I made lots and lots of corrections, and also enforced modern standards -- all the comments are now (ideally) transcribed as originally written, and anything not original has been moved into a ##-style comment. I also added in the asterisks that sometimes precede operators, and moved the junk (e.g. "B0") that occasionally falls between the variable column and the comment column outside of the comments.

yaLEMAP received corresponding updates -- it now prints the asterisks in the column preceding the operators as in the original listing, for both .lst and .html outputs. It also passes through the junk (colored as comments) and tries a bit to at least position it in a sane manner (again, for both .lst and .html).

The one thing that's still broken is ##-style comments. They emit pretty poorly in the html. I haven't spent much time digging into that yet.

@rburkey2005 part of the reason I didn't spend too much time looking at the code during this pass was because I noticed that FP6 doesn't have a binsource, and FP8 only has a partly complete one. Do you have old binsources lying around that didn't make into the repository? If not, I think I'm going to make my own FP6 binsource to 100% totally convince myself that there are definitely no complementary errors in the code anywhere.


rburkey2005:
@thewonderidiot Thanks!  As usual, it's humbling to look back at something you thought you were being so careful and clever with, and find that it was shoddy workmanship.  :smile:

I've looked at the earliest development snapshot still available, which is basically just a tarball of my working directory.  That's from 6 months after AGS development was completed. The binsource files are in the same state (or lack of state) in which you found them today.  Apparently, I originally intended to create them and use the same proofing process as for the AGC, but changed my mind due to the (relatively) small size of the programs, and the fact that FP8 is missing a couple of pages.  The website says the following:
<blockquote>
Validation of the AGS source code is somewhat streamlined from the method used to validate AGC source code.  The following process is used:

1. The source code is entered into the computer from the assembly listings. 
2. It is assembled with the yaLEMAP cross-assembler.  As a byproduct, a ASCII file is produced with just the columns of the assembly listing containing the assembled octal values.
3. The byproduct ASCII file mentioned above is fed into a text-to-speech converter program, and the now-audible octal codes are compared visually against the original scan of the assembly listing.  The source files are fixed as needed when mismatches are found.

Obviously, it's a matter of opinion how many times the  proofing process described above needs to be repeated.  I have repeated it only until the embedded checksums are correct.
</blockquote>

The initial proofing was just as reliable as under the AGC process, I think, but to redo the proofing now, a dozen years later, it would be nice if those binsource files had been independently created.

thewonderidiot:
So it goes -- I think the octopus comment proofing method showed that no matter how hard you try, there'll still be errors that slip through.

And yeah, because these listings are so small, it's not a big deal to just whip up a new binsource. It's probably better that way anyways. :smile:

thewonderidiot:
Alright, I've finished fiddling with yaLEMAP. Both .html and .lst outputs should be nice and pretty (and formatted similarly to the AGC programs) now.

I added support for multiple checksums to binLEMAP, and have gone through the first 16 pages of the FP6 binsource. That's enough to cover the first checksum, which is correct. Everything so far (unsurprisingly) matches the assembled code.

thewonderidiot:
I have finished the FP6 binsource! Unsurprisingly, once I corrected the binsource errors, it matched the assembled code perfectly.

I made one more change to binLEMAP, in addition to those mentioned above. I noticed at the very end of FP6 and FP8 is a line printed as:
```
CHECKSUM=377372002647
```

After playing with this number a bit, I found it is calculated by summing all words in the entire memory (including those not covered by other checksums), and then swizzling the two words of the resulting 32-bit sum. So, for example, since FP6 has its checksum printed as `377372002647`, that means that the sum of all of its words should be `002647377372`. And indeed, this is the case. Anyways, binLEMAP now verifies these total checksums if given the "TOTAL" directive.

I think at this point I am done playing with the AGS tools, and FP6 should be totally done, barring maybe one final comment proofing pass. I'm going to move on to FP8 and give it the same amount of love.

PR #1057, opened 2018-01-05
Correct transcription errors in Colossus ropes

Thymo-:
I was looking up some information in Artemis' assembly and operation information when I spotted a potential transcription error by the mistyping of CMC as CGC. Upon checking the scans I saw that instead of the misspell there was actually an error between 'REQUEST WITH' and 'REQUESTWITH'.
I spoke with Mike about this and checked the other ropes. Turns out this same error exists in all Colossus ropes starting from 249 and up. 

I also added an extra level of indentation in the lines below that in Artemis to conform with the indentation that was used in the others.


rburkey2005:
Thanks!  I did lots of proofing with a special-purpose tool, but the tool is insensitive to spacing errors.

PR #1058, opened 2018-01-06
Channel 163 bit for TEMP

thewonderidiot:
In the AGC, the output to the TEMP light is a logical OR of channel 11 bit 4 (the output bit to turn on the light) and channel 30 bit 15 (the input bit from the IMU indicating its temperature is out of range). This was done so the light would still work in standby mode. This is apparently needed for the TEMP light to operate correctly in NASSP, since software appears to be lazy about maintaining the output bit.

These changes add a new bit to channel 163 that indicates what the current state of the TEMP light should be, and update all of our various DSKYs to use that instead of channel 11 bit 4.

@indy91, let me know if these changes make the TEMP light work correctly for you.

### DIFF
--- a/piPeripheral/piDSKY.py
+++ b/piPeripheral/piDSKY.py
@@ -11,6 +11,8 @@
 #				of piDSKY2.py (in which all hardware-specific stuff
 #				has been removed), because it was easier than 
 #				back-porting bug-fixes.
+#		2018-01-06 MAS	Switched the TEMP light to use channel 163 instead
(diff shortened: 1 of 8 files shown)


indy91:
These changes work very well. Previously a V37E 00E cleared the TEMP light even if the input bit was still set, now it doesn't do that anymore. And the light extinguishes as expected, so when the input bit changes.

rburkey2005:
Super, as usual!

Issue #1055, opened 2017-12-29
Improve I/O counter handling


thewonderidiot:
There's been desire recently to trigger 1201/1202 program alarms at realistic times in the NASSP Apollo 11 descent, and for them to happen for realistic reasons. This means we need to hit accurate processor load. The primary (and I think, only) thing holding us back right now is a full simulation of all of the input and output counters, and their cycle-stealing increment instructions.

This is going to be a joint project with NASSP, with @indy91 doing most of the heavy lifting on their side.

Internally, the timers already generate PINC or DINC sequences at appropriate times as according to the scaler, and incur the proper amount of stolen time. This logic will be expanded to include all output counters (GYROD, CDUXD, CDUYD, CDUZD, TRUND, SHAFTD, THRSTD, EMSD, OTLNK, ALT).

The input counters will need to start using the increment sequences as well, and we'll have to figure out how the phasing of them will work. A lot of them have logic that, for example, doesn't allow pulses through unless specific channel bits are set, or things like that, that will need to be captured.

It's my intention to fully emulate the behavior of the counter cells, so each pending increment request will be serviced according to its priority (counters with lower addresses have higher priorities).

### Open Questions
* Should we handle PIPA pre-counting logic? Counts come in from the PIPAs at a fixed frequency, clocked by the AGC, but instead of generating increment instructions directly like CDUs the pulses are filtered through a pre-counter. This block only lets PIPAs counts through if there have been more than "normal" in one direction. Under no acceleration, the PIPAs generate constant 3 pulses in one direction and then 3 in the other; the pre-counter generates no counter increments until at least 4 in the same direction are received. It also performs sanity logic making sure pulses are coming in at the right frequency and whatnot.

* Should the interfacing application interleave count requests with calls to agc_engine(), or should all of the input counters have fifos like CduFifo that work through overall combined requests at a predetermined rate?

* How and when do we notify users of changed output counters? Do we have separate "counters" that we effectively move counts into? For example, for THRSTD, every time a throttle change is incremented via DINC we could increment the separate counter the other way, so the requested throttle would slowly move from one spot to the other.

rburkey2005:
I agree with your approach, to the extent I understand it right now :smile:, and think it's what should be pursued.  But after reading it, it occurred to me to wonder if there might be a short-term, quick-and-dirty change that would allow triggering 1201/1202 alarms at will, whilst waiting for your correct fix to become available as the long-term solution?  I wonder if the following mightn't work in the very short term?

[...]

PR #1060, opened 2018-01-28
Improved scaler logic and counter simulation for yaAGC

thewonderidiot:
Here's the first pass at refactoring for correct execution priority and counter support, as laid out in #1055. The scaler logic had become an unsightly pile of spaghetti, so I've greatly cleaned it up and broken it out from agc_engine(). It should be much, much easier to follow now for anybody referencing schematics or ND-1021042.

Counters are now tracked via an array of "counter cells", which can store either plus or minus requests for any individual counter. The cells are processed in priority order, and handle cases of both requests (franken-instructions) correctly. Everything internal has been moved over to the new system and appears to be working.

I took the opportunity to clean up a lot of cruft, and some of it may or may not find its way back in (trappipa, cdulog, etc.) as I go through adding in the unique logic for each counter.

As it stands, all of the tests contained in Borealis still pass, standby operates as it is supposed to, and everything generally seems working. I still need to do more testing, though.

I made a variety of other changes working through the above:
 * Added input voltage monitoring and voltage alarm simulation for use by NASSP and other integrated spacecraft simulations.
 * Added both halves of the counter alarm (only count instructions executed, or one not executed immediately after being requested).
 * Reorganized the order of operations in agc_engine() so all activities have the correct priority -- GOJAM > counts > rupts > instructions.
 * Made INHINT, RELINT, and EXTEND prevent counts or interrupts from happening if they are the next instruction to be executed, since they are really "pseudo-instructions" that extend the length of the preceding instruction by 1 MCT.
 * Moved instruction effects ot the first MCT, rather than the last, for multi-MCT instructions.  This is much closer to the truth, generally, and makes priority simulation easier.
 * Made BZF and BZMF take an extra MCT if their branches are not taken, which we had missed.

Next up is implementing each of the counters themselves. I'm going to start with the RHC circuits, since they are pretty interesting and the interface to NASSP is very straightforward (they provide a voltage level from the RHC, and the AGC internally converts that to a number of pulses).

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -412,6 +412,10 @@
  *				themselves are not. Also added descriptive macros
  *				for interrupt numbers, so I don't have to keep
  *				looking them up.
+ *		02/01/18 MAS	Simulated the GYROCMD output counter. It doesn't
+ *				check for the target gyro at the moment, which
+ *				may be something we'll want to do, depending on
+ *				what is most useful for integrators.
  *
  *
  * The technical documentation for the Apollo Guidance & Navigation (G&N) system,
@@ -1247,18 +1251,50 @@ CounterDINC (agc_t *State, int Counter)
   if (i == AGC_P0 || i == AGC_M0)	// Zero?
     {
       switch (Counter)
-      {
-      case COUNTER_TIME6:
+        {
+        case COUNTER_TIME6:
           State->InterruptRequests[RUPT_T6RUPT] = 1;
           // Triggering a T6RUPT disables T6 by clearing the CH13 bit
           CpuWriteIO(State, 013, State->InputChannel[013] & 037777);
           break;
-      }
+        case COUNTER_GYROCMD:
+          // Gyro drive pulses have completed. Clear the active bit and
+          // channel 14 bit 10.
+          State->GyroDriveActive = 0;
+          State->InputChannel[014] &= ~01000;
+          break;
+        }
     }
   else if (040000 & i)			// Negative?
-    i = AddSP16(SignExtend(i), SignExtend(AGC_P1)) & 077777;
+    {
+      i = AddSP16(SignExtend(i), SignExtend(AGC_P1)) & 077777;
+      switch (Counter)
+        {
+        case COUNTER_GYROCMD:
+          if (!State->GyroDriveActive)
+            {
+              // Set the gyro drive active bit if it wasn't set.
+              State->GyroDriveActive = 1;
+              State->GyroDriveOut = 0;
+            }
+          break;
+        }
+    }
   else					// Positive?
-    i = AddSP16(SignExtend(i), SignExtend(AGC_M1)) & 077777;
+    {
+      i = AddSP16(SignExtend(i), SignExtend(AGC_M1)) & 077777;
+      switch (Counter)
+        {
+        case COUNTER_GYROCMD:
+          if (!State->GyroDriveActive)
+            {
+              // Set the gyro drive active bit if it wasn't set.
+              State->GyroDriveActive = 1;
+              State->GyroDriveOut = 0;
+            }
+          break;
+        }
+    }
 
   State->Erasable[0][RegCOUNTER + Counter] = i;
 }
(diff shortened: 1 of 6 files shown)


rburkey2005:
@thewonderidiot Whatever happened with this?  Did I drop the ball or something?  (It looks like --- now, a couple of years later --- it has some conflicts to resolve.)

indy91:
It's my fault mostly. I encouraged Mike to work on this back then, just to realize that I have to change basically everything in NASSP code that interacts with yaAGC for this to work. That never happened, so this never got implemented or tested in NASSP.

rburkey2005:
As long as it's not _my_ fault, I consider it fault-free.  😄  But what should a proper disposition be?  Should the issue be closed?  It would be a shame, obviously, to discard what looks to be a lot of work.  I guess there's no harm just keeping it open.

thewonderidiot:
The blocker to getting this merged into VirtualAGC (ignoring NASSP for the moment) is that I've totally broken all of the old external interface methods (through fictitious channels, etc.), and need to write a new compatibility layer to translate. Ideally in such a way that no hypothetical existing thing that we may not know about that interfaces with virtualagc breaks. Given the large amount of timing changes I'm afraid this is going to be extremely tricky, and with all of my other projects I haven't had time to dedicate to sitting down and really working through that. I do still really want to get this merged in eventually though....

yaAGC/agc_engine.c
```diff
@@ -1740,64 +1541,1066 @@ SimulateDV(agc_t *State, uint16_t divisor)
...
+
+//-----------------------------------------------------------------------------
+// These functions implement scaler timing logic. The scaler is a 33-stage
+// binary counter whose outputs are used to time many things throughout the
+// AGC. Each stage has four outputs -- the inverted and noninverted state for
+// that stage (named, eg., FS09 and FS09/), and two timing pulses of the form
+// FxxA and FxxB. Timing pulse FxxA is emitted when the stage transitions from
+// a 1 to a 0, and timing pulse FxxB is emitted when the stage transitions
+// from a 0 to a 1. The frequency of each stage can be calculated by the
+// forumula (1.024 kHz)/(2^x), where x is the number of the stage.
```
michaelfranzl, line 1612: Assuming that the comment below is true,

```
// ... Each FS03 count
// therefore happens in half that time, every 39.0625 microseconds. This turns
// out to be (10/3)*MCT.
```

then it must be

```suggestion
// formula (102.4 kHz)/(2^x), where x is the number of the stage.
```

According to https://en.wikipedia.org/wiki/Apollo_Guidance_Computer#Timing the input to the scaler was the crystal oscillator signal of 2.048 MHz divided by 20 (2 x 2 x 5). The Wikipedia article seems to describe Block I (because it mentions only 17 scaler stages), but it must have been the same in Block II.

yaAGC/agc_engine.c
```diff
@@ -1740,64 +1541,1066 @@ SimulateDV(agc_t *State, uint16_t divisor)
...
+
+    // Add the divider to the internal counter, and advance the scaler as
+    // many times as we can (which really should only be once, if that).
+    State->ScalerCounter += SCALER_DIVIDER;
+    while (State->ScalerCounter >= SCALER_OVERFLOW)
+    {
+        State->ScalerCounter -= SCALER_OVERFLOW;
+        // Increment the scaler and update input channels 3 and 4. Channel
+        // 4 shows the current value of scaler stages 6 through 19 (so it
+        // counts at 3200kHz) and channel 3 shows stage 20 through 33 (so
```
michaelfranzl, line 2323:
```suggestion
        // counts at 3200Hz) and channel 3 shows stage 20 through 33 (so
```

yaAGC/agc_engine.c
```diff
@@ -1740,64 +1541,1066 @@ SimulateDV(agc_t *State, uint16_t divisor)
...
+        }
+    }
+    else if (GtsetType == SCALER_GTRST)
+    {
+        // If a radar sync was active, stop it, generate a RADARUPT, and
+        // reset all of the related state information
+        if (State->RadarSync)
+        {
+            State->InterruptRequests[RUPT_RADARUPT] = 1;
+            State->InputChannel[013] &= ~010;
```
michaelfranzl, line 1928: This state update needs to propagate to all connected clients / peripheral simulations (via `ChannelOutput` wrapped by `CpuWriteIO`).

```suggestion
            CpuWriteIO(State, 013, ReadIO(State, 013) & ~010);
```

rburkey2005: @michaelfranzl Are you by chance making these suggestions in the expectation they'll make their way into the main branch of the yaAGC code?  I ask, because your patches are being made against a code branch that it seems clear (from the creator) is not going to be merged soon, if ever, into main.  In the meantime, your patch suggestions are being viewed as something having to do exclusively with this side-branch of the code, and are essentially being ignored as candidates for the inclusion in the main branch of themselves.

michaelfranzl: Thanks for the hint! This is OK for me. :) I'm making suggestions specific to the changes in this branch independent from it being merged, with the intent of making the code slightly more adapted for the integration in peripheral simulations.

As this does create some clutter in the comment section of this pull request, maybe it is better to simply modify my fork of this branch. My changes could then be contributed later.

For what it's worth, the conflicts in this pull request are located only in the comment preambles of a few files and can be easily resolved. My fork of this branch is already rebased on an up-to-date main branch.

rburkey2005: Oh, I wasn't making any hints about reducing clutter.  The Virtual AGC Project probably seems to many people to be 100% clutter anyway!  :smile:   I just wanted to make sure you weren't having expectations accidentally doomed to failure.  Which you're not.

PR #1061, opened 2018-09-05
Reconstructed Luminary 130

thewonderidiot:
Tonight while trying to figure something out, I went and looked up the differences between Luminary 130 and Luminary 131. To my surprise, Luminary memo 129 (http://www.ibiblio.org/apollo/Documents/LUM129-001.pdf) explicitly describes the singular difference between the two revisions. I checked the affected pages in the Luminary 131 listing, and sure enough the exact two lines described have asterisks next to them.

So unless the memo is neglecting to mention something, this is enough information to fully reconstruct Luminary 130, as we did for Luminary 99 Rev 0. I've gone ahead and done that here.


thewonderidiot:
@indy91 and I have been studying the bug a bit, to get an understanding for what exactly the bug does. According to [Luminary memo 127](http://www.ibiblio.org/apollo/Documents/LUM127_text.pdf), it has the potential to destroy some "valuable erasable memory" during an "early abort" from descent. It seems that the memory destroyed is somewhat random.

What happens is that the relative branch "GOTO +3" in the following snippet:
```
		VLOAD	GOTO
			J1PARM
			+3
 +4		VLOAD	SET		# IF J2 IS USED, SET THE
 			J2PARM		# ABORT TARGETING FLAG
			ABTTGFLG
		STODL	JPARM
```
branches to `ABTTrightGFLG` and executes that as an interpretive instruction, rather than going to `STODL JPARM` as it is supposed to. `ABTTGFLG` assembles to `04470`. Treated as an interpretive instruction, this disassembles to `STORE 0467,1` -- an indexed store to a constant erasable address, using index register X1. (0467 is in the middle of VAC area 2). And since the last load before this rogue STORE was a VLOAD, the interpreter is in vector mode, and will write out a whole six words to this random location. These six words will be copied from J1PARM, since this was the "vector" loaded.

Unfortunately, the P70 job doesn't use any index registers up until this point. The index registers aren't initialized by the Executive or Interpreter or anything; each job using them is supposed to initialize the index registers themselves. This means that the exact location written to is going to be determined by whatever value the last job to use the VAC assigned to P70 last. Because P70 is always started from the restart tables through ENEMA, I'm pretty sure that P70 always gets VAC area 1 (unless restart groups 1, 2, or 3 are active during early descent and kick off a FINDVAC job). At the very least, it's deterministic. What's not deterministic is which job last had control of this VAC area, and what state it left that VAC area's X1 register in.

So in summary, an early abort in Luminary 130 will cause 6 values, equivalent to J1PARM and the words following it, to be written to the address `(0467 - X1)`, where `X1` is the 38th word in the VAC area assigned to P70, and which will retain whatever value the previous owner of the VAC area had in it at the time of the abort.

thewonderidiot:
I've added bugger word checking based on drawing 2021152G, and (as hoped) the reconstruction appears to be correct, as all bugger words match.

rburkey2005:
@thewonderidiot Note that I've made changes both to the top-level Makefile and to the Main.annotation files for LUM69R2 and Luminary130, so be sure to pull them first if you have any other miracles up your sleeve.

There was nothing really wrong with the Main.annotation files, but I felt that they were too long due to repeating a lot of info from Luminary069 and Luminary131, as well as speculating about what annotations the true assembly listing might have on them, so I just tried to make them more concise.

thewonderidiot:
Okay, will do! I haven't done so yet, but I'm quite confident that I'll be able to recreate Luminary 97 and 98 pretty easily. I started trying to do Colossus 236, but apparently Colossus 237 was printed before GAP supported the asterisk thing, so I haven't been able to identify what changed... other than that it was something in bank 31.

PR #1063, opened 2018-11-13
Explain core memory size.

shirriff:
The text here confused Wikipedia, leading to the wrong core rope size on the AGC page.


rburkey2005:
Agreed.  Merging.  Thanks, Ken.

PR #1067, opened 2019-06-04
Fix typos.

shirriff:
Changes verified against scans: https://archive.org/details/yulsystemsourcec00hugh/page/n6


rburkey2005:
Thanks, Ken!  You fooled me, though.  For anyone else who might read this, the link to the original should actually be: https://archive.org/details/yulsystemsourcec00hugh/page/n7.

PR #1076, opened 2019-07-27
Reconstructed LUM69 revision 2

thewonderidiot:
The ~4 remaining banks of Sundial have been a slog so far, so I decided to take a break from it and try my hand at reconstructing a program using the new banksums from drawing 2021152B. This is the first, LUM69 revision 2. It was tricky, but I'm pretty confident that this is correct.

From [Luminary Memo 75](http://www.ibiblio.org/apollo/Documents/LUM75_text.pdf), we know that the only change between the two was the addition of the Boeing R-2 lunar potential model. The only module changed was B2. According to [2021152B](https://archive.org/stream/apertureCardBox467Part2NARASW_images#page/n68/mode/1up), only the bugger words for banks 11 and 13 changed:

- 11: 46137 to 00066
- 13: 72477 to 72415

In Luminary 99, the change was made by changing a lot of code for ORBITAL INTEGRATION in bank 11, and then adding more code (QUALITY1 and QUALITY2) in bank 13. Given, however, that the change to the bugger word in bank 13 is very small, I made the opening assumption that mostly likely only constants or variable/function offsets had changed that bank. 

I started by looking for changes in gravitational-related constants between 69 and 99. The first was easy to find, in J4REQ/J3. Luminary 69 has
```
J4REQ/J3    2DEC*    .4991607391 E7 B-26*
            2DEC    0
```
while Luminary 99 has 
```
J4REQ/J3    2DEC*    .4991607391 E7 B-26*
            2DEC    -176236.02 B-25
```

This constant is related to the potential model, so presumably is part of the R-2 change. However, just this change was not enough to achieve the correct banksum. The other change in the bank was found via [Luminary memo 78](http://www.ibiblio.org/apollo/Documents/LUM78_text.pdf), which describes the addition of the R-2 model into the main Luminary line, in revision 95. As part of the change, TIMEDELT was updated from 10 seconds to 20 seconds. This change, in conjunction with the J4REQ/J3 change, was enough to get the correct bugger word for bank 13.

For bank 11, I simply pulled over all potential-model-related differences between Luminary 69 and 99 in ORBITAL_INTEGRATION.agc. I created a new tag to position QUALITY1 and QUALITY2 at the end of bank 11 instead of bank 13. The two new padloaded erasables E3J22R2M and E32C31RM were added to the first two available locations in unswitched erasable memory (this is one location lower than in Luminary 99). With the code changes backported, and the new positions for QUALITY1/2 and the erasables, the banksum of 11 is also correct.

For validation I created a new script, Tools/check_buggers.py, that takes in an assembled rope binary and a new file that lists bugger words from one of the 202115x drawings. It checks that every listed bugger word matches, and raises an error if not. LUM69R2 is able to pass this test, with perfect agreement between the reconstructed binary and the 2021152 bugger words.

If all of this looks agreeable, I'll go back and add the same validation for my reconstructed Luminary 130, as well as Colossus 237.


thewonderidiot:
Oh -- and @indy91 has tested this in NASSP and it appears to function correctly.

rburkey2005:
Terrific, Mike!  Very impressive.

rburkey2005:
I don't know if you're familiar the Harry Potter movies, but it reminds me of Hermione showing the boys a huge book and commenting that it was something she had checked out of the library a couple of months earlier "for light reading".  :smile:

thewonderidiot:
Haha, thanks! :smile: I know exactly what you're referring to, heheh.

I added and pushed bugger word checks for Colossus 237 and Luminary 99 Rev 0. I also did so for Luminary 130, and (as I had hoped back when I did it) the reconstruction appears to have been correct, as the bugger words all match. I've updated that pull request with the check.

PR #1078, opened 2019-07-29
Luminary 97 and 98

thewonderidiot:
Here's two more reconstructed Luminary versions: 97 (the original release for 11) and 98 (intermediate development). The bugger words for 97 match those listed in drawing 2021152D. We don't know any bugger words for 98, but considering it was made as a stepping stone to 97, it is presumably also correct.

Both of these revisions were quite straightforward, thanks to the detailed descriptions in Luminary [memo 83](http://www.ibiblio.org/apollo/Documents/LUM83_text.pdf) and [memo 85](http://www.ibiblio.org/apollo/Documents/LUM85_text.pdf), and the extremely helpful change-indicating asterisks emitted by GAP.

I based the Main.annotation comments off of what you did for LUM69R2, but (of course) feel free to change them. :smile:

Working backwards from Luminary 99:
**Luminary 98 to 99**:
-  Anomaly 70 fix. Memo 85 helpfully states that there were no code changes here, simply a relocation of erasables. Looking at redesignation-related erasables, there are four that clearly stand out as moved:
image
Their original locations were replaced with "dummy" tags to preserve the relative locations of all other tags that referenced them:
image
This change was undone simply by moving the erasables back to their original location.
- PCR 775. This PCR introduced doppler compensation for the LR. Two erasables, RADSKAL and SKALSKAL, were introduced (their two lines have asterisks). A chunk of code at the beginning of POSUPDAT in SERVICER. 
image
To undo these changes, all lines containing asterisks were replaced with code from Luminary 69.

**Luminary 97 to 98**:
- Anomaly 65 fix. A new chunk of code, SETITRCTR, was added to the end of bank 11 (in ORBITAL INTEGRATION) and a call to LAMBERT in P34-P35, P74-P75 was replaced with a call to SETITRCTR.
image
image
To undo the change, the SETITRCTR code was deleted, and the call in P34-P35, P74-P75 changed back to LAMBERT.
- Down-Telemetry word order bit change. A new piece of code called "WOTEST" to check timing for writing the word order bit was added to the end of the bank, and the lead-in to the program changed to jump to it.
image
image
This change was undone by deleting the WOTEST code and changing the line that jumps to it back to "CAF BIT7". The associated comments were restored from Luminary 69.


thewonderidiot:
Frustratingly, we have the banksums for Luminary 96 as well, but I'm not seeing a memo that describes the changes between 96 and 97. It seems as though only bank 5 changed. But unfortunately, the changes between 97 and 99R1 are spread out enough that any asterisks that would have marked changes between 96 and 97 would have gotten wiped out with further changes.

rburkey2005:
Looks good to me ... as long as you understand that means that the explanation sounds convincing and reasonable to me, and not that I've gone through and checked each step you've described for myself.  Though obviously having the checksums match up in the end is pretty convincing just in and of itself.

This is great work, Mike, thanks!

rburkey2005:
@thewonderidiot Remind me of why we think 97 was the initial release for Apollo 11?

thewonderidiot:
R-700 volume III lists it as such, and it was the first set of ropes to actually be manufactured after Luminary 69/69R2:
image

rburkey2005:
My assumption regarding 2021152 was that there was a tape for each manufactured rope and vice-versa.  As you noted before, there's a tape for Luminary 96.  Perhaps my assumption is in error, and the tapes and ropes aren't in 1-to-1 correspondence.  But if so, I wonder what selection criteria there were for making tapes.

thewonderidiot:
You know what? You're right, and R-700 is wrong. Luminary 96 was the first release, and was manufactured, as shown in drawing 2021112B. Luminary 97 was released to correct whatever the problem was in bank 5. It was a single-module remanufacture of B1, changing B1's part number from 2003972-951 to 2003972-1021. Luminary 99 further changed B1 to 2003972-1041, and finaly 99R1 changed it again to 2003972-1071.

Well, shucks. Now I really want to know what the change between 96 and 97 was...

rburkey2005:
Speaking of which (not really), I thought I'd take a look at Artemis 71 (checksums!) vs Artemis 72 (assembly listing!) and experiment with recreating 71.  Given that there are no COLOSSUS or ARTEMIS memos related to it so far, that's probably really tough.  But I saw enough to become confused.  When the '*' markings appear in the assembly listings to indicate changes, what are they changes _from_?  The next lower-numbered revision?  According to 2021154, Artemis 71 and 72 differ only in banks 1, 33, and 34.  Yet in the 72 assembly listing I see lots of '*' change markers in bank 21.  And some in other unexpected banks too.  I think there are too many to relate to program-comment changes, and too many to imagine that they all cancel out.  What am I not understanding about these markers?

thewonderidiot:
The asterisks are per subroutine revision, not per program revision. So any asterisks you see in the subroutine SKIPPER, for example, date from the last time SKIPPER was revised, not the last time Luminary was revised. That's why our Luminary 99R1 listing still has asterisks dating back to Luminary 97 in it -- each of the individual subroutines was only revised once between 97 and 99R1.

thewonderidiot:
The opening section of the program, which is listed as subroutine "(MAIN)", does indicate changes between each revision of the program proper. The final page of the "main" program tends to be a page that instantiates each of the subroutines with SUBRO calls. For development versions of the software, they tended to just do things like "SUBRO    SKIPPER", to get the latest version. However, for released versions, they added another argument to SUBRO to specify the exact subroutine revision -- e.g., "SUBRO   SKIPPER  029".

The SUBRO table for Artemis 72 has a single asterisk in it, for subroutine MENELAUS:
image

This means that the only change between 71 and 72 was in that subroutine, and all asterisks in the MENELAUS subroutine are different between 71 and 72. Presumably, in Artemis 71, MENELAUS was revision 060, or possibly earlier. Any other asterisks in (MAIN) also differ between the two. All other asterisks outside of (MAIN) or (MENELAUS) happened earlier in Artemis development, pre-71.

rburkey2005:
Yes, that was part of my confusion (though I didn't mention it), since it didn't match the asterisks throughout the rest of the listing.  Thanks!  Just out of curiosity, is this behavior something you've inferred, or was told, or something from the documentation?

Unfortunately, it still leaves me with a problem, since the checksum changes in bank 33 do seem to come from Menelaus, but those in banks 1 and 34 don't.  Perhaps I just didn't read far enough in the listing.  That's a problem for another day.  :smile:

thewonderidiot:
Something I inferred while studying the Luminary 99R1 asterisks a while ago. I'm pretty sure it's right, though; I haven't found any countexamples yet.

How many asterisks are there in MENELAUS? Are you thinking there aren't any at all in banks 1 or 34? I think I see which bank 33 changes you're referring to:
image

If these are the only changes, then the punch card numbers might hold the secret. All affected lines seem to hold their "original" four-digit punch card number, except for card 11235, "TCR DPAGREE". If this card was introduced as part of the change, then the offsets of all code in the bank after that card would have been shifted up by one. Any references to that code in other banks would also be affected. So perhaps the addresses towards the end of bank 33 shifted a bit, and banks 1 and 34 both referenced them? Do the banks 1 and 34 banksums become correct if you delete the line "TCR DPAGREE"?

rburkey2005:
I can't answer those questions thoroughly, since I've just been looking at it idly to get an idea of the difficulty.  I.e., I'm not prepared to look at it seriously right now.  I've seen no asterisks at all for bank 1, but I've only looked through about the first 1000 pages.  I'll have to come back to it when I'm willing to give it more time vs assembly drilldowns and LVDC transcription.

PR #1080, opened 2019-07-31
Reconstructed Comanche 051

thewonderidiot:
Here's the first reconstructed CMC program -- Comanche 51, the original release for Apollo 11. This one is trickier since we're lacking Colossus memos from the time. It turns out, though, that the differences between it and Comanche 55 are exactly the same as the differences between Luminary 69 and 69R2, and Comanche 45 and 45R2 -- the R-2 lunar potential model.

I decided to start here, based on comments in Comanche 55 about coding "MODULE CHANGE". There's surprisingly few asterisks in Comanche 55, but there's a few in ERASABLE ASSIGNMENTS for the R-2 padloads, and one for the J4REQ/J3 constant change. Based on all of this, I tried backing out all of the R-2 stuff, and sure enough, with only these changes removed, the banksums all lined up perfectly right away. The ORBITAL INTEGRATION code was replaced with the pre-R-2 version from Luminary 69.

I'm pretty sure, by the way, that Comanche 54 and 55 are the same (possibly only differing in the NASA author drawing dash number), since there are no asterisks whatsoever in the (MAIN) program or the subroutine table. Also, based on the dates, I think that Comanche revisions 52 through 54 were where the R-2 model was first developed and tested, after which it was merged into Comanche 45, Luminary 69, and Luminary 95.

This has gotten me thinking a lot about Comanche 45, since it's only a tantalizing six revisions away. There's a whopping 30 (!) bugger word difference between 45 and 51, though, which is making it a tough nut to crack, especially with the dearth of asterisks. My guess is that the offset of something changed in erasable or fixed-fixed, and caused all of the code that references that thing to slightly change... but I haven't found a consistent explanation yet. @indy91's been looking into the PCRs implemented between 45 and 51 to see if we can try to hone in on what some of the changes were. But if we can get to 45, we can get to 45R2, and make another flight version.


rburkey2005:
Terrific!  And more importantly, plausible.

You're thinking much bigger than I am.  Instead of thinking about reconstructing Comanche 45, I've been sitting around grousing about not being able to get Margaret's copy of Comanche 44.  If you were able to reconstruct 45, and we were eventually able to cross-check it someday from 44, that would be amazing.

rburkey2005:
@thewonderidiot I've had a naive idea, which sounds too stupid to even mention.  As if that ever stopped me!  I wonder if there would be any benefit at all of trying to crowd-source an attempted 51 -> 45 reconstruction by appealing to the mailing list?  Yes, it's not a rote kind of thing like transcription is ... but transcription is _boring_!  I can't help but notice on the other hand that reconstruction seems to be kind of an intellectual game.  People (other than me) seem to like games, and if introduced to the rules, there might be a few people who'd like to play it.

thewonderidiot:
Hmmmm, yeah, that's a really good point. I'll write something up describing what we're trying to do and the tools we have (how asterisks work, etc.) and send it out to the mailing list tonight or tomorrow. You're right, it's definitely more of an intellectual game, and I've been having quite a bit of fun doing it for the past few days. :smile:

thewonderidiot:
Also yes, I've been thinking the same about Margaret's Comanche 44. Perhaps in a few months we could try bugging her again, this time from the context of being able to recreate even more versions from it....

rburkey2005:
I think we need to develop an alternate route to her.  While she responds politely to me, I fear _my_ inquiries have been at a level of approximately 0 on her scale of interest for quite some time.

thewonderidiot:
@rburkey2005 here's another thing I just found that would also be a huge help: Purdue has the Norton programmed equations document for Colossus 2 (i.e. Comanche 45): http://collections.lib.purdue.edu/fa/pdf/msa5_armstrong_n.pdf

That would almost certainly be enough for us to root out most of the differences...

thewonderidiot:
I didn't get a chance to work on this last night, but I have been thinking about it, and want to run my current plan by you before I implement it over the next couple of days.

What I'm thinking is that I'll create a new branch in the repository for each version that we have the bugger words for that has not yet been reconstructed. On each branch, I'll create a folder for the program in question, containing files from the closest available version we have. I'll create *.buggers files for each program and set up the Makefiles to use them.

Once that's in place, I'll open a new issue for each program-to-be-reconstructed that links to the branch for that program. The description of the issue will contain all of the information I can gather that may be helpful in performing the reconstruction: the scanned listings, the Norton document if we have it, applicable memos or GSOPs, etc. These issues will serve as the discussion hub for working on each program.

Lastly, once all of the infrastructure is set up, I'll send out an email to the mailing list linking to each issue, and describing the process and the tools (e.g. how asterisks work in the listings). I'll probably also put this information on a wiki page and link to it from each issue.

How does this sound?

rburkey2005:
It sounds fine to me!  You might want to put just one or two of them out there at first, though, because there's no way of knowing what kind of response there's going to be to this.  If nobody (or only a couple of people) respond, you might have wasted a lot of time on it for nothing.

Though as far as what _I_ meant to suggest was concerned ... I was really only thinking about reconstructing 45 from 51.  My reasoning was that there were so many banks changed and so little guidance from external documentation that there would be a lot of trial and error involved, so that getting a gang of people to mess with it might conceivably help.  Or might not.

But thinking past that to _all_ of the missing versions with checksums is certainly fine too.

rburkey2005:
I was also thinking about making a posting myself, in advance, that just talks about the availability of all the new revisions you had constructed.  That might serve the purpose of getting a few people thinking about it before you mad your posting.  What do you think?

thewonderidiot:
That sounds good to me -- I'd say go for it!

I was originally thinking just 45 too, but I think at this point I've gotten most of them that can be reconstructed just by sitting down and messing with it for a few minutes. The rest of them are going to take some work -- so even if there's no biters, the issues will at least be a good place for me to take notes as I work on things.

PR #1088, opened 2019-08-15
Addition of 3 completed parts and WIP models.

ManoDaSilva:


ManoDaSilva:
@rburkey2005 
1. I wanted to be as complete as possible with the attached .jpg files, which I find really useful for interpreting the CAD model (threads, surface, etc). Solidworks files would have been useful for anyone willing to correct an error I could have made (.STEP files are solid, not parametric; any modification might require extensive work).  
I'll remove them :) 

2. No problems about the README.md, I'll leave it as it is and include my remarks on each separate .md file. 

3+4. I used the convention on the parent folder, but didn't notice it was only if the file was modified for a specific purpose. I'll correct this right away.


Sorry about the confusion! I'll close this request and create a new one compliant with your remarks.

rburkey2005:
@ManoDaSilva  Considering you're the first one with an actual submission (as opposed to Real Soon Now :smile:), there are bound to be teething difficulties.  I feel like apologizing for being so narrow as to what I'll accept!

For whatever it's worth, the rationale behind the policies is simply that STEP is a standard available in most modeling systems, including open-source ones, whereas Solidworks, however useful (it's what the mechanical engineers use in my company) is still an expensive proprietary system only available to a limited subset of users.  Frankly, every mechanical (or electrical) engineer I speak to has a different preference for the proprietary solutions they want to use.  To keep the repository size from exploding, all such proprietary solutions are excluded.

What I'd recommend is simply to mention in the readme file that the STEP file was exported from Solidworks.  Anyone who definitely wanted those would then be likely contact you directly through your GitHub handle and ask for them.  That would be more useful in terms of getting bugs fixed anyway, I think, since you could exchange your thoughts on the subject more readily.

ManoDaSilva:
No need to apologize! I totally understand you have to keep a strict structure in such a large project.

I myself use Solidworks at work, hence my tendency to use it for this ;) 

I'll append this info on the next pull I'll do, probably tomorrow or the day after, once I give the finishing touches to the next models. Some drawings are really ambiguous and need to be compared, one revision against the other. If all else fails, I can always compare the dimensions to other parts that mate with it.

PR #1091, opened 2019-08-16
Added 3 new models

ManoDaSilva:
Front cover, front plate and front plate connector. Added licensing and more infos regarding the model's translation from the drawings.


rburkey2005:
They look great, thanks Manoel!

PR #1092, opened 2019-08-19
More models

ManoDaSilva:
Here's the alarm indicator, cover, rubber gasket and some spelling/grammar corrections.


thewonderidiot:
@rburkey2005 If it helps, here's pictures of a real one I got to look at recently (1006316, but everything except for the alarm lights should be identical): https://photos.app.goo.gl/bxWAifsWsHnj5ofH7

rburkey2005:
Cool pictures.  It doesn't help much on the front side, since there's a cover on it.  On the back, though, you can clearly see the header (which I misspoke of as a connector), and so I was wrong and no wires have to enter the back indentation.  However, it's _also_ clear that the holes for the header's pins are clearly missing from the model ... and are entirely missing from the original drawing as well

I think that drawing 1006387 isn't the fabbed part at all.  I think it's just a reference from which it's easy to get the overall dimensions of the part.  But I don't see anything in 2003994's bill of materials which could actually be the drawing for the fabricated part.  It must be part of one of the subassemblies of 2003994, which makes more sense than it being a direct child subassembly of 2003994.

thewonderidiot:
What you're getting at here is the difference between Specification Control Drawings (SCDs) like this one, and part/assembly drawings. SCDs are specifications for the procurement of third-party components -- so like what you'd get in a transistor or resistor datasheet. This, and the digital indicator, just happen to be larger, assembly-level parts procured externally. All of the SCDs have a note on the first page that either give the names of approved vendors of the part (if the SCD is early enough) or tell you to consult drawing 1002034, which is a big listing of that information, which NARA unfortunately doesn't have.

The original drawings for these things would be manufacturer-specific, and not released directly to NASA, so there's unfortunately no chance of finding them at NARA. These SCDs are the best we'll get for such things.

thewonderidiot:
The one I posted pictures of earlier has been fully painted over, but the one in the Computer History Museum's DSKY, and this one from the Smithsonian, were both manufactured by Oppenheimer, and have Oppenheimer's part numbers for it: https://airandspace.si.edu/sites/default/files/images/collection-objects/record-images/A19960356000cp01.jpg

rburkey2005:
I can believe the part was externally manufactured, but It's a bit hard for me to believe that they simply left the design of this part up to the outside vendor.  Nevertheless, even if that's true, my point is that this particular 3D model, 1006387D, is unfortunately not something of much use in building a DSKY because while it may be accurate with respect to the original drawing, the drawing is very much unlike the original physical part.

thewonderidiot:
They could leave the internal design up to the manufacturer because they specified everything they cared about in the SCD. Per the MIT drawing standards:
image

I promise you won't find any more detailed drawings than these from MIT. You'd have to contact Oppenheimer, if they even still exist. And I disagree that it's not useful while building a DSKY; the SCD specifies everything you need to know about this piece, from dimensions to pinouts to voltages, just not the exact internal construction.

rburkey2005:
Okay, even before the comment, I had already had second thoughts about my claim that they wouldn't have just left it up to the manufacturer.  I concede that point.

None of which alters my _main_ point, which is this:  Of what use is a 3D model of a part that does not have the functionality needed in a DSKY?  It's somewhat equivalent to going into an auto parts store to buy a battery, and having them sell you a block of plastic that's the right size and shape and weight to go into your car, but is otherwise non-functional.

In other words, should this 3D model be included in the repository, and if so, why?

rburkey2005:
(Sorry, I hit the "close" button by accident.)

rburkey2005:
What I would suggest is that the model for 1006387D can be included, but it should be renamed as something like "1006387D-NonFunctionalDimensionalModel.stp" to make it clear that it's only good for giving the external dimensions.

thewonderidiot:
To follow your analogy, I would argue that it's useful to know what a car battery looks like if you go in to buy one -- what the exact shape of it is, how far apart the terminals are, where the screw points are, etc. If you're going to build a car, you do need to know all of this information, even if your source for this information doesn't deal directly with the chemistry of the cells inside.

In other words, all of the information contained in this model is necessary for anybody making a DSKY. They have to further modify it, sure, but they need to conform to the dimensions shown, and it's useful to be able to visually see it as a 3D model instead of having to hunt down the SCD.

I think it would be clearer to just put things like this in a subfolder called "SCDs" or something, perhaps with a README that explains what SCDs are and why these parts are separated.

ManoDaSilva:
Sorry about the confusion, I didn't really know where or how to express the fact that 1006387D is a model with only the external dimensions specified. Do I create a new folder, rename the file, both? 
Agreed with @thewonderidiot : I would also say this model is useful for anyone willing to re-create the indicator alarm (myself included), albeit with some modifications (creating a pocket for the PCB and connector for instance). 



### Part numbering issue
This also reminds me another problem I noticed with the E/L indicator assembly 2003988-011: if you look at [the subassembly listing](https://www.ibiblio.org/apollo/2014999-221.html#2003988-011), it specifies 2004699-001 as cover/frame. However, when you look at the drawings, this part is the alarm indicator cover, not the E/L indicator: the window opening is too small for the display. 

Digging a bit further in the subassembly **drawings** (not the auto-generated listing), here's what I understand:
* E/L indicator assembly (2003988-011) specifies frame 2004699-**001**
* Alarm indicator assembly (2003897-041) specifies frame 2004699-**000**

The same part number has two configurations: one for the E/L cover with a tall window, another for the alarm cover with a smaller window. There seems to be a small error in the auto-generated listing, where both assemblies specify 200469-**001**. 

Which inevitably brings me to this issue: I didn't find any reference to 2004699-001. I can "extrapolate" its dimensions from the 000 configuration and extend the window up to .182in from the frame's edge (which would make it perfectly symmetrical), but that wouldn't be 100% accurate until we have the drawings in front of us.

thewonderidiot:
Yeah, sadly our copy of 2004699 seems to be too old to have the -001 variant drawn. And we don't have any prospects for finding a newer version if it at the moment, sadly.

I do know somebody who has one of these, though, and can ask him to measure both sides of it to see if it is exactly symmetrical, if you'd like.

thewonderidiot:
And no worries! These are just growing pains, to be expected as we work through how to handle these drawings. This sort of thing always comes up when we wade into something new. You're the first person who's been digitizing them and submitting them for general use, so really, thank you!

rburkey2005:
@ManoDaSilva Regarding the possible error in assembly drilldown 2003897C, I don't think you're correct.  The original drawing 2003897C calls out 2004699-000 for 2003897-011 and -021, but calls out 2004699-001 for 2003897-031 and -041, and that's just what the automated drilldowns show.

findTable

However, it's very plausible that I've made errors elsewhere in the assembly drilldowns, so keep your eyes open!

Regarding the file-naming problem.  If we thought there were going to be lots and lots of these SCD-based static models that didn't actually do anything, it might make sense to invent a system for handling them (such as a special folder to hold them).  Right now we have exactly one of them, it's unclear if there will be any more.  [The file-naming convention that has already been published](http://www.ibiblio.org/apollo/ElectroMechanical.html#3D_Models) handles it, as I suggested in my earlier post.  I don't insist on the specific name I suggest earlier ("1006387D-NonFunctionalDimensionalModel.stp"), as long as it's something that makes it clear nobody should 3D-print the thing and expect it to work.

ManoDaSilva:
Okay, I just renamed the file as you suggested and added an addendum to the attached .md file. I'm now just gonna add "-000" to the 2004699A model. 

@thewonderidiot confirming that fact would really help! Thank you very much! 
By the way, your photos could also help me understand the general outline of unclear parts (for instance the 2003947 main plate assy)... Do you have any other?

thewonderidiot:
Okay, I'll ask. And I can send you the full link to the album those pictures came from, although for various reasons I don't think I can share the whole thing publicly yet. Be forewarned that it's a thermal test unit and not all components inside are as they would be in a flight DSKY. Can I send them the email address you've used for these commits?

rburkey2005:
Thanks, Manoel!

ManoDaSilva:
> 
> 
> Thanks, Manoel!

With pleasure!



> 
> 
> Okay, I'll ask. And I can send you the link to the full album those pictures came from, although for various reasons I don't think I can share the whole thing publicly yet. Be forewarned that it's a thermal test unit and not all components inside are as they would be in a flight DSKY. Can I send them the email address you've used for these commits?

Thank you! I'll keep that in mind :) 
Yes, It should be the right one.

ManoDaSilva:
I'm having some troubles with the 2004699A frame; it doesn't seem to fit inside the the opening of the 2004929- front cover: both width and height of the frame (2,765x4,569) are bigger than the cover's opening  (2.68x4.48), including in the drawings. Drawings and pictures of Block II DSKY's clearly show it shouldn't be colliding...
image

Would other revisions alter dimensions of parts, or are there different frames/covers with other part numbers?

rburkey2005:
For whatever it's worth, looking at 2004929D and 2004699A, I'd conclude that the dimensions are as you say they are. 

But it appears to me that DSKY 2003994-121 (which I thought these parts were for) calls out 2004739 rather than 2004929.  I imagine we should be using 2004739.  Admittedly, 2004739 isn't dimensioned very thoroughly.  However, 2004739 does imply a width of 6.415-3.560=2.855 and a height of 4.925-0.275=4.65, which would be large enough for 2004699.

thewonderidiot:
This came about as the design of the DSKY evolved over time. The original design didn't have the indicator covers at all. The windows in the front cover, P/N 2004929, were sized appropriately, to hug the two indicator modules tightly:
image

Then, there was a brief time where only the alarm indicator, and not the E/L indicator, had a cover -- DSKY P/N 2003994-021. This had a front cover with P/N 2004739-000, which we have a drawing for. That drawing is, as Ron says, isn't dimensioned very thoroughly. However, note #1 says "MATL MAKE FROM 2004929-021". So essentially what they did was just take the existing 2004929 covers, and widen the alarm window a bit so that the newly added cover could fit through. I have seen a picture of one of these DSKYs, but can't find it at the moment.

This configuration didn't last long, however, and a cover was added to the E/L panel as well in DSKY P/N 2003994-031. The front cover was further revised to 2004937-001, and the E/L window was widened in the same way as the alarm window (this isn't shown on our copy of 2003947, which only covers the -000 variant, but is pretty easy to confirm via pictures).

So in summary, 2004937 front covers started life as 2004929 front covers, and simply had their windows widened to make room for the indicator covers. For DSKY 2003994-121, you'd start with 2004929, and then adjust both windows as shown in 2004937.

thewonderidiot:
Also, my friend sent measurements for his E/L indicator cover:
> Best I can measure right now, the "as built" sides seem to be 0.182", top is 0.161", bottom is 0.188" (tried a few times too).

ManoDaSilva:
> 
> 
> This came about as the design of the DSKY evolved over time. The original design didn't have the indicator covers at all. The windows in the front cover, P/N 2004929, were sized appropriately, to hug the two indicator modules tightly:
>
> 
> Then, there was a brief time where only the alarm indicator, and not the E/L indicator, had a cover -- DSKY P/N 2003994-021. This had a front cover with P/N 2004739-000, which we have a drawing for. That drawing is, as Ron says, isn't dimensioned very thoroughly. However, note #1 says "MATL MAKE FROM 2004929-021". So essentially what they did was just take the existing 2004929 covers, and widen the alarm window a bit so that the newly added cover could fit through. I have seen a picture of one of these DSKYs, but can't find it at the moment.
> 
> This configuration didn't last long, however, and a cover was added to the E/L panel as well in DSKY P/N 2003994-031. The front cover was further revised to 2004937-001, and the E/L window was widened in the same way as the alarm window (this isn't shown on our copy of 2003947, which only covers the -000 variant, but is pretty easy to confirm via pictures).
> 
> So in summary, 2004937 front covers started life as 2004929 front covers, and simply had their windows widened to make room for the indicator covers. For DSKY 2003994-121, you'd start with 2004929, and then adjust both windows as shown in 2004937.

Got it, this makes perfect sense! I'll create 2004937 and adjust the dimensions. 



> 
> 
> Also, my friend sent measurements for his E/L indicator cover:
> 
> > Best I can measure right now, the "as built" sides seem to be 0.182", top is 0.161", bottom is 0.188" (tried a few times too).

I'll also copy 2004699A-000 and adjust the window to those dimensions for 2004699A-001 :) 


We're getting there! I still have some doubts with the position of some holes.... This will be confirmed when making the entire assembly.

ManoDaSilva:
> 
> 
> This came about as the design of the DSKY evolved over time. The original design didn't have the indicator covers at all. The windows in the front cover, P/N 2004929, were sized appropriately, to hug the two indicator modules tightly:
>
> 
> Then, there was a brief time where only the alarm indicator, and not the E/L indicator, had a cover -- DSKY P/N 2003994-021. This had a front cover with P/N 2004739-000, which we have a drawing for. That drawing is, as Ron says, isn't dimensioned very thoroughly. However, note #1 says "MATL MAKE FROM 2004929-021". So essentially what they did was just take the existing 2004929 covers, and widen the alarm window a bit so that the newly added cover could fit through. I have seen a picture of one of these DSKYs, but can't find it at the moment.
> 
> This configuration didn't last long, however, and a cover was added to the E/L panel as well in DSKY P/N 2003994-031. The front cover was further revised to 2004937-001, and the E/L window was widened in the same way as the alarm window (this isn't shown on our copy of 2003947, which only covers the -000 variant, but is pretty easy to confirm via pictures).
> 
> So in summary, 2004937 front covers started life as 2004929 front covers, and simply had their windows widened to make room for the indicator covers. For DSKY 2003994-121, you'd start with 2004929, and then adjust both windows as shown in 2004937.

Got it, this makes perfect sense! I'll create 2004739-000 and adjust the dimensions. For 2004739-021, I'll use the Alarm window size for the E/L one. 


> 
> 
> Also, my friend sent measurements for his E/L indicator cover:
> 
> > Best I can measure right now, the "as built" sides seem to be 0.182", top is 0.161", bottom is 0.188" (tried a few times too).

I'll also copy 2004699A-000 and adjust the window to those dimensions for 2004699A-001 :) 


We're getting there! I still have some doubts with the position of some holes.... This will be confirmed when making the entire assembly.

rburkey2005:
@ManoDaSilva By the way, I don't know if you're aware of it (since it's not obvious), but you can edit your earlier comments (without having to create an entirely new one that's only slightly different).  Click on the little "..." at the upper right of the comment, and select Edit from the drop-down box that appears.

ManoDaSilva:
> 
> 
> @ManoDaSilva By the way, I don't know if you're aware of it (since it's not obvious), but you can edit your earlier comments (without having to create an entirely new one that's only slightly different). Click on the little "..." at the upper right of the comment, and select Edit from the drop-down box that appears.

Whoops, actually I double posted... I didn't notice I had already answered on another tab of my browser....

Issue #1093, opened 2019-08-20
Reconstruct Luminary 178


thewonderidiot:
### Branch
[luminary178](https://github.com/virtualagc/virtualagc/tree/luminary178)

### Changed Bugger Words from Closest Program (Zerlina 56)
```
Bugger word mismatch in bank 02; actual 45071 != expected 51710 (diff = 04617)
Bugger word mismatch in bank 03; actual 57614 != expected 70325 (diff = 10511)
Bugger word mismatch in bank 00; actual 71552 != expected 72407 (diff = 00635)
Bugger word mismatch in bank 01; actual 60422 != expected 40225 (diff = 20175)
Bugger word mismatch in bank 04; actual 17700 != expected 17535 (diff = 00143)
Bugger word mismatch in bank 05; actual 71636 != expected 62775 (diff = 06641)
Bugger word mismatch in bank 06; actual 71104 != expected 63272 (diff = 05612)
Bugger word mismatch in bank 07; actual 67560 != expected 10063 (diff = 57475)
Bugger word mismatch in bank 10; actual 55562 != expected 44335 (diff = 11225)
Bugger word mismatch in bank 11; actual 10055 != expected 05114 (diff = 02741)
Bugger word mismatch in bank 12; actual 51645 != expected 51622 (diff = 00023)
Bugger word mismatch in bank 13; actual 25102 != expected 23036 (diff = 02044)
Bugger word mismatch in bank 14; actual 67656 != expected 70355 (diff = 00477)
Bugger word mismatch in bank 15; actual 47477 != expected 45155 (diff = 02322)
Bugger word mismatch in bank 16; actual 16122 != expected 16353 (diff = 00231)
Bugger word mismatch in bank 17; actual 47533 != expected 50247 (diff = 00514)
Bugger word mismatch in bank 20; actual 62166 != expected 62561 (diff = 00373)
Bugger word mismatch in bank 21; actual 45174 != expected 70434 (diff = 23240)
Bugger word mismatch in bank 22; actual 77334 != expected 26775 (diff = 50337)
Bugger word mismatch in bank 23; actual 50263 != expected 70552 (diff = 20267)
Bugger word mismatch in bank 24; actual 71411 != expected 21371 (diff = 50020)
Bugger word mismatch in bank 25; actual 51552 != expected 52021 (diff = 00247)
Bugger word mismatch in bank 26; actual 04201 != expected 01576 (diff = 02403)
[...]

PR #1095, opened 2019-09-18
Reconstructed Luminary 178

thewonderidiot:
After a month of work as described in #1093, I've finally (somewhat unexpectedly) finished reconstructing Luminary 178! It turns out the Densmore Luminary memos are very good, and describe essentially all of the changes well enough to sort out when they happened. I've tested it for basic functionality in standalone yaAGC, and @indy91 created a padload for it and performed a test landing using the NASSP Apollo 14 Pre-PDI scenario. It passed the landing with flying colors!
image

Only six lines of code had to be guessed/written anew; the other changes were all either present in one of Luminary 131, Zerlina 56, or Luminary 210, _or_ described well enough in the memos that only one code option was possible.

The first two of the six "new" lines are in P63START4 (`CAF ONE; TS LRPOS`). The flowcharts showed the location of this code, and that LRPOS was initialized with 1. The naive implementation here seems to be correct.

The remaining four are a different implementation of the remode check in RRDESEND. The naive change to the flag checking code in Luminary 210 didn't work. However, changing to another pattern frequently used to test flags gave the right sum (with the right number of words in the bank), so I'm pretty sure it's correct.

Since this reconstruction has gone so much further than all of the others so far (which have only been a revision or two away from something we had), there's a lot more room for cosmetic errors in this. Specifically, the exact location of erasable definitions (i.e., the order that they are presented in ERASABLE ASSIGNMENTS) and comments throughout are practically guaranteed to have errors. For comments in particular, sometimes the Luminary memos simply say things like "Comments were updated for E3", giving an entire bank in which comments may have changed. I've done my best to pick comments from the closest listing for each chunk of code, but there are definitely places I have it wrong.

All of this is cosmetic though; after having gone through everything I'm quite confident that the code itself and the binaries produced are correct. But to build more confidence, I'm planning on taking this and pushing backwards to Luminary 173, and maybe/hopefully 163. Each of those is described by a single memo instead of many, so it should be a much easier job.


rburkey2005:
@thewonderidiot A masterpiece, Mike, truly.  Awesome work. I'd say it's worthy of being on a résumé, except that that would minimize it.  I'd say it's worthy of an award, except (excepting Margaret Hamilton) nobody hands out awards for the AGC.  Excuse my gushing.  Mainly, it's been a sore spot with me that we've had (more or less) all of the flown Luminary versions except 178, and now you've filled that gap.

@indy91 And thanks, Nik. Having it flown in the sim is the final icing on the cake.

indy91:
Yes, Mike did an outstanding job! What I had to do was quite simple really. Between Luminary 1D and 1E the padloaded parameters and their addresses changed almost not at all, so I took an Apollo 14 pre PDI scenario I had lying around, changed the address of just a single padloaded parameter and ran the scenario. The AGC complained just a little bit that I had suddenly given it another rope (41203 alarm), but after that it worked flawlessly.

rburkey2005:
@indy91 I'm in the process of writing it all up for the website, and there was a lovely description there before of how you had adapted Luminary 210's ephemeris to fly simulated Apollo 14 landings.  I almost regret having to remove it to make way for the updated situation.

indy91:
@rburkey2005 Yeah, I felt the same way when I removed the modified rope from our repository. I saved a backup of it though, it will definitely not be lost or forgotten that it worked great for flying Apollo 14!

thewonderidiot:
Haha, thanks a lot! :smile: To be honest, I'm still a bit shocked that it actually came together. My first approach of removing all of the Zerlina-specific stuff from Zerlina and then only using the Luminary 179 through Luminary 183 memos to back out changes ended up failing pretty badly... I think I still had 34 out of 36 banks wrong after that. That's what pushed me in the direction of just doing a 3-way diff of every section and trying to pin every single difference I saw on a particular revision. And for that I think I had to go through over a thousand pages before I had more than 4 banks right at the same time!

Only one flown Luminary (LM131 Rev 1) remains to be reconstructed... but I'm going to do Luminary 173 and 163 first, I think, while I'm still in the 1D mindset.

Issue #1097, opened 2019-09-21
Reconstruct Luminary 163


thewonderidiot:
### Branch
[luminary163](https://github.com/virtualagc/virtualagc/tree/luminary163)

### Changed Bugger Words from Closest Program (Luminary 173)
```
Bugger word mismatch in bank 02; actual 51712 != expected 51657 (diff = 00033)
Bugger word mismatch in bank 01; actual 57434 != expected 57432 (diff = 00002)
Bugger word mismatch in bank 04; actual 17477 != expected 17543 (diff = 00044)
Bugger word mismatch in bank 05; actual 63025 != expected 63463 (diff = 00436)
Bugger word mismatch in bank 06; actual 63065 != expected 63025 (diff = 00040)
Bugger word mismatch in bank 07; actual 06662 != expected 07000 (diff = 00116)
Bugger word mismatch in bank 10; actual 44335 != expected 44343 (diff = 00006)
Bugger word mismatch in bank 11; actual 05120 != expected 05171 (diff = 00051)
Bugger word mismatch in bank 13; actual 22756 != expected 22746 (diff = 00010)
Bugger word mismatch in bank 14; actual 70076 != expected 70266 (diff = 00170)
Bugger word mismatch in bank 15; actual 44165 != expected 76433 (diff = 32246)
Bugger word mismatch in bank 22; actual 26455 != expected 26425 (diff = 00030)
Bugger word mismatch in bank 23; actual 56772 != expected 56724 (diff = 00046)
Bugger word mismatch in bank 24; actual 21334 != expected 15077 (diff = 04235)
Bugger word mismatch in bank 25; actual 51551 != expected 70474 (diff = 16723)
Bugger word mismatch in bank 26; actual 00476 != expected 00446 (diff = 00030)
Bugger word mismatch in bank 27; actual 74706 != expected 65044 (diff = 07642)
Bugger word mismatch in bank 30; actual 53576 != expected 57143 (diff = 03345)
Bugger word mismatch in bank 31; actual 44462 != expected 77260 (diff = 32576)
Bugger word mismatch in bank 32; actual 40304 != expected 74747 (diff = 34443)
Bugger word mismatch in bank 33; actual 53244 != expected 65027 (diff = 11563)
Bugger word mismatch in bank 34; actual 52335 != expected 53204 (diff = 00647)
Bugger word mismatch in bank 35; actual 45732 != expected 57716 (diff = 11764)
[...]

PR #1099, opened 2019-09-22
Reconstructed Luminary 163

thewonderidiot:
Luminary 163 has also been reconstructed! The process is described in #1097.

This was based off of Luminary 173, and was even easier to put together than Luminary 173, since there's not one, but three memos that describe all of the changes between the two. The problem point for this, like for Luminary 173, was ACB L-11, but I've discussed the resolution to that elsewhere.

I did have to write one "line" from scratch for this, but it's very trivial. Between Luminary 164 and 167, the erasable TOOFEW moved from its old location (sharing with QSAVED) to its final resting place, the unswitched unshared location 1354. This was unused in earlier revisions. However, Luminary 152 through 163 _did_ use locations 1355 and 1356 as SERVDURN and DUMLOOPS. To make sure they assembled to the correct place, I added the line
```
FILLER      ERASE
```
to use up location 1354, and gave a descriptive comment about why I put it there.

There's a particularly fun piece of data we get from this reconstruction: Luminary memo 148 has a table showing the number of words remaining in each bank for Luminary 163. [As shown in this comment](https://github.com/virtualagc/virtualagc/issues/1097#issuecomment-533852340), this reconstruction matches that table exactly. Which isn't surprising, really, but it adds even more confidence that the reconstructions for 163, 173, and 178 are all correct.

Anyways, as mentioned in the Luminary 173 pull request, I haven't added either program to the top-level Makefile to prevent merge conflicts. They'll need to be added to it after merging.


rburkey2005:
I'll do the Makefile stuff.  I have to change the VirtualAGC program anyway, to include 163 and 173 support, so it's only natural.

PR #1100, opened 2019-10-02
Retread 50: Completed disassembly of new extended verbs

thewonderidiot:
I've decided to take a bit of a break from reconstruction and spend some time focusing on all of the rope modules we've dumped recently. To kick things off, I went back to Retread 50 and finished off the disassembly of the new stuff in bank 11. No more pesky UNKxxxx labels!

It turns out a lot of it was able to be correlated to various chunks of code in Aurora, just to a lesser extent than the stuff I already had done. The only thing that was still almost entirely different from other things we have was Verb 43, for gyro pulse torquing. To help with the disassembly, I created a flowchart of how it works (matching the MIT flowcharts style as closely as I could):
Retread 50 Pulse Torque Flowcharts (5)

To use this verb, you apparently have to manually enter three double-precision numbers into location 1210 (which I've guessed was named "OGC"), for the outer, middle, and inner gyro commands, in that order. You then invoke verb 43, which will send the correct amount of pulses to each gyro.

I also changed the name of the section from BANK11.agc to EXTENDED_VERBS_FOR_MODING.agc, which is what Solarium has. I went with this since Solarium is the closest rope we have in time to Retread, and all of the new verbs are related to IMU moding.

Let me know if any of the changes are unclear, or if you have better suggestions for names of things.

### DIFF
--- a/Retread50/ERASABLE_ASSIGNMENTS.agc
+++ b/Retread50/ERASABLE_ASSIGNMENTS.agc
@@ -7,6 +7,7 @@
 ## Website:     www.ibiblio.org/apollo/Restoration.html
 ## Mod history: 2019-06-12 MAS  Recreated from Computer History Museum's
 ##				physical core-rope modules.
+##              2019-10-01 MAS  Assigned names to the UNK12xx erasables.
 
 ## Page 5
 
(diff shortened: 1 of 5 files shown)


rburkey2005:
Thanks, Mike!  It's hard to see how there would be much ground for objecting to any of that.

PR #1101, opened 2019-10-02
Update README.md

lazymutt:
Added many wikipedia links for cross reference, edited formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,67 +10,67 @@ Virtual Apollo Guidance Computer
 [travis-image]: https://travis-ci.org/virtualagc/virtualagc.svg?branch=master
 [travis-site]: https://travis-ci.org/virtualagc/virtualagc/branches
 
-The Apollo spacecraft used for lunar missions in the late 1960's and early 1970's was really two different spacecraft, the Command Module (CM) and the Lunar Module (LM).  The CM was used to get the three astronauts to the moon, and back again.  The LM was used to land two of the astronauts on the moon while the third astronaut remained in the CM, in orbit around the moon. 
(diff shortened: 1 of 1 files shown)


rburkey2005:
@lazymutt Most of your changes make sense to me, Todd, and seem good, but I do question a few of them ... possibly because my understanding of markdown syntax isn't perfect.  

Perhaps you'd go back and review what you've done with back-ticks.  As I understand it, the markdown `single-tick` construct is used for indicating code sequences.  That makes sense to me to use additionally for things like command-line sequences (which is one of the things you've done), as well as for names of directories, but you seem to have gone further than that, and basically replaced all quoted words or phrases with back-ticked ones.  (Forgive me if I'm exaggerating the extent of the changes.)  _For example,_ I see `Virtual AGC`.  A while there _is_ a program called "VirtualAGC" (no space) whose name would be appropriately ticked if it appeared, the two-word phrase "Virtual AGC" (whether quoted or unquoted in the original text) is the name of the project itself.  It might be appropriately quoted or left without special formatting, but it's hard for me to interpret it as code.  My impression is that there were other back-ticked examples that didn't really seem to fall into the `code` category either, though I did not make a list.  If you'd look over the back-ticked changes and make sure they're really appropriate, I'd appreciate it.

On a different matter, dual spaces after periods and colons is a common though not universal convention ... in particular it was the convention I was taught.  In other words, it is not a mistake that needs correction.  By googling I see that (at least according to the Chicago Manual of Style) it has now gone "out", and single spaces are "in", so I understand the motivation for changing it.  But if you make it your business to change my use of spaces, and then later to change it all back when the old convention becomes "in" again, I think you've got your work cut out for you.  :smile:  Fortunately, there's no particular need for consistency, so there's no need to undo the changes just to please me.  But personally, I'm going to keep doing the way I've been doing it for the last 45 years. 

Thanks!

rburkey2005:
Actually, I see now that no matter whether you use one space or two after periods or colons, the markdown processor seems to always cause them display as a single space.  So I guess it really doesn't matter one &iota; (iota) what your or my own preference on the matter is!

lazymutt:
I appreciate your concerns! 

Coming from a printing background, the double spaces thing is a pet peeve of mine. I apologize for subjecting you to it. :)

It looks like I was a bit hamfisted with some of the formatting. I'll go back over my suggestions more closely and resubmit my PR.

PR #1102, opened 2019-10-02
Update README.md

lazymutt:
Added many wikipedia links for cross reference, lightly edited formatting.


rburkey2005:
Frankly, I hadn't realized until I googled it that the 1-space thing had now become essentially a universal consensus except among us old folks.  So I should have just kept my mouth shut about it, even though I'm sure my fingers are going to continue to do what they've always done in the past.

rburkey2005:
I've merged it in, thanks!

PR #1106, opened 2019-12-09
update titles up to entry 339

7ofNine:


rburkey2005:
Thanks!

rburkey2005:
I do have some comments, regarding the LEO drawings.  Regardless of what it says in the title block (LEO-200, LEO-280, etc.), I think the drawing number we need to actually use is a combination of the DRAWING NO and EO NO fields.  Thus:

- Link 46:  LEO200-1, sheet A02
- Link 53:  LEO200-1, sheet A01
- Link 64:  LEO200-1, sheet A04
- Link 280:  LEO280-1466, sheet A01
- Link 283:  LEO280-1467, sheet A04
- Link 294:  LEO280-1467, sheet A01
- Link 298:  LEO280-1673, sheet A05
- Link 326:  LEO280-1840, sheet A01
- Link 339:  LEO280-1840, sheet A01
...

Otherwise, you end up with lots and lots of drawings called (for example) LEO-280, but they're mostly different from each other.  This way, they're at least _mostly_ unique (except for the case of 326 and  339, where the sheet number doesn't change for some reason).

I got this idea that this is how the numbering system should be from Link 845, LEO390-209.  If you look at that drawing, the DRAWING NO field actually says "LEO390-" (instead of "LEO-390"), with the clear implication, I think, that what's in the DRAWING NO field of the title block is not the full drawing number.

rburkey2005:
There's actually a huge heap of thousands of LEO drawings (boxes 686-688 contain almost nothing but LEO drawings), and it would be helpful to have those for discussing this issue, but unfortunately I won't be scanning them until (probably) some time in 2021.

7ofNine:
So far it is only a small number of them. When the time will come one might possibly better understand how these mountains of drawings were actually organized and then it should be a relatively small effort to update them if necessary. Until now I am not sure how they are actually organized in the sense of hierarchy as well as history.

For the time being I’ll add the EO number.

PR #1107, opened 2019-12-10
change LEO-280 drawing numbers to LEO280- with EO number

7ofNine:


rburkey2005:
Thanks!  My comments about LEO were general, and I wasn't insisting on any immediate action.  After all, I could have done it myself if I was so concerned.  :smile:  But I do think it's better to get ahead of it.   It's not just that I expect that when we get to boxes 686 and 687 we'll find a couple of thousand LEO-280's; it's that on the basis of the small sample we have right now, it appears that there may be 30-40 LEO's per box, so that by the time we get to 686 we may have found 5000 of them.  If we didn't have a usable system of distinguishing between them, it would be quite an effort to go back through them and change them all.  Of course, that's based on the notion that you or I or anybody would have the energy to have processed the titles on all those boxes, which may be a bit of a stretch.  :smile:

7ofNine:
I just did the change because I was already on the page and it was a pretty small change. If there are further changes necessary in the future it should be easier to have the “wrong” ones already in a consistent form. This might allow to update/change/modify them using a script or at least find tem easily with the help of a tool. For some of the LEO’s  the only change seems to have been a check mark and sometimes  the title is a bit of a head scratcher when they relate to multiple drawings. Not quite sure how they tracked all their paperwork at Grumman. What does LEO actually stand for? Something like “Large Engineering Order”?

 

On the side I noticed that the links on the https://virtualagc.github.io/virtualagc/AgcDrawingIndex.html#GAEC page for the LM drawings are broken.

 

As they say in German “mühsam ernährt sich das Eichhörnchen <https://de.wiktionary.org/wiki/m%C3%BChsam_ern%C3%A4hrt_sich_das_Eichh%C3%B6rnchen> ” , only a few more thousand drawings to go.

rburkey2005:
On 12/10/2019 09:13 AM, 7ofNine wrote:
> I just did the change because I was already on the page and it was a 
> pretty small change. If there are further changes necessary in the 
> future it should be easier to have the “wrong” ones already in a 
> consistent form. This might allow to update/change/modify them using a 
> script or at least find tem easily with the help of a tool. For some 
> of the LEO’s the only change seems to have been a check mark and 
> sometimes the title is a bit of a head scratcher when they relate to 
> multiple drawings. Not quite sure how they tracked all their paperwork 
> at Grumman. What does LEO actually stand for? Something like “Large 
> Engineering Order”?
>
I expect the "L" is probably for "LEM".  "EO" is undoubtedly 
"Engineering Order".  Presumably "DW" is for "Drawing".  thus an LEO is 
an engineering order for a change to an LDW or its associated physical item.
>
>
> On the side I noticed that the links on the 
> https://virtualagc.github.io/virtualagc/AgcDrawingIndex.html#GAEC page 
> for the LM drawings are broken.
>
I hadn't given that any thought to those links before, but it's what I 
would have expected.  It's a general phenomenon in that almost all links 
on virtualagc.github.io are broken.  The reason for this is that the 
virtualagc.github.io site isn't an exact mirror of the ibiblio.org site, 
in that it only contains the HTML and images displayed on the website.  
It can't contain the 10+ gigabytes of drawings and documents from the 
main website.

In other words, in my view, the broken links aren't a bug as such. It's 
true that the broken links can be "fixed", and there are a few cases 
where I've done so when they became really inconvenient to work around, 
but in general the fixes are distasteful to me and I'd prefer to just 
let it as-is.  (Fixes include things like changing all URLs to absolute 
ones, which would make the HTML much bigger in some cases and would 
break the website for anyone who downloaded it to use on a local drive, 
or javascript to rewrite all the links at runtime, which is something I 
really want to minimize.)  Admittedly, if there were a "free" fix that 
wasn't distateful, it would be a nice thing to have.
>
>
> As they say in German “mühsam ernährt sich das Eichhörnchen 
> <https://de.wiktionary.org/wiki/m%C3%BChsam_ern%C3%A4hrt_sich_das_Eichh%C3%B6rnchen> 
> ” , only a few more thousand drawings to go.
>
>
>

PR #1108, opened 2019-12-12
add modified script for LEM technical drawings.

7ofNine:
This is a slight modification of the normalizeUnpunchHtml.py script

### DIFF
--- a/Scripts/normalizeLEMUnpunchHtml.py
+++ b/Scripts/normalizeLEMUnpunchHtml.py
@@ -1,0 +1,71 @@
+#!/usr/bin/python3
+# The idea of this program is that it takes one of my AgcDrawingIndexXXX.html
+# files, and for each DRAWINGNUMBER/DOCTYPE pair, duplicates the title of the
+# first instance found into all of the other instances (for different revs,
+# sheet numbers, frame numbers, and copies).  The point of this is that it 
+# allows me to just adjust the very first instance's title and to painlessly
+# propagate that change to all of the other instances with the same drawing
+# number and document type.  Otherwise, it's rather painful and time-consuming
+# to do that.  At the same time, it performs a slight amount of cleanup, such
+# as eliminating spurious empty lines which SeaMonkey composer adds in for some
+# reason.  I may also eventually use it to standardize boilerplate stuff that
+# also appears in the files, but it doesn't do that right now.
+#
+# The program does not attempt to parse arbitrary HTML, and relies on the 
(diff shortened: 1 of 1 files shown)


rburkey2005:
@7ofNine I think I might also change it so that the default for an EO title (the first time it's encountered for a particular drawing number) will be the same as for a previously-encountered DW title (with the same number).  That shouldn't affect the way you're using it, but would help out lazy guys like me.  Comments?

rburkey2005:
And come to think of it, I should also make it so that when it encounters a LEO part number, the default title should be the most-recent EO title encountered or (failing that) the most-recent DW title encountered.  Those assumptions would almost always be correct, other than variations in punctuation or abbreviations.  Again, it wouldn't affect the way you're using it at all, I think.

rburkey2005:
With your mod, there isn't actually any reason to distinguish between EO and DW scans, is there?  There might be if they were interspersed randomly, but since the HTML has always been pre-sorted so that the EO scans follow the DW scans, the distinction between EO titles and DW titles is already taken care of by your changes to the caching algorithm.  So I may just get rid of that distinction too.

7ofNine:
Go for it. It only affects me positively. I am also lazy and the fewer words I have to type the better it is. I only set it up based on the small set of drawings I have seen. What one has to do is to verify that the titles are actually correctly generated but my understanding is that one should do that anyways for the drawing number, i.e. one should have a short peek at the single drawing anyways. I have a local copy of all the drawings and go through them according to drawing number using a quick photo editor. That is nicer to my bandwidth. This way I see the actual drawing number and the title. Your suggestion might even be faster for me, too. ignoring EOs and LEOs would give them  a title and quite often they are quite similar, i.e retyping (or copy and paste) could be replaced by some small editing. I can try it and get a feeling if it works faster for me. I have another variation anyways that uses script arguments instead of "<" and ">". I am lazy, it saves typing of two characters.

7ofNine:
That sounds correct. If they get “classified” by the sorting. If that is guaranteed it shall work.

rburkey2005:
I've pushed the changes to the script I mentioned.  Also, caching of LEO drawings is eliminated completely, because any given LEO drawing number (such as LEO280-252) will typically appear many times, but with different titles every time, since the title will be from whatever the last LDW drawing with a checkmark was.

rburkey2005:
If you decide to download and use this version of the script, be careful with it at first.  I _think_ it's working, but considering the potential for damage, care is warranted.

PR #1110, opened 2019-12-19
Update titles up to entry 681

7ofNine:
also mark illegible drawings in red


rburkey2005:
Thanks!

rburkey2005:
By the way, a few comments.  (Not saying to _change_ anything, just commenting on the way I personally am doing things!)

Regarding LEO drawings:  While none of them lists a title internally, other than phrases like "SEE DWG", I've been titling them identically to the LDW drawings they're paired with.

Regarding LDW drawings with missing title sheets:  Quite often, there's an associated LDW E.O. associated with it.  While the DW and EO titles usually aren't _exactly_ the same, they're always very, very close, usually differing only in how various things are abbreviated.  So instead of leaving the drawing-title blank, I just use the E.O. title for the drawing itself ... perhaps with something in the Notes column to explain what I did.

Regarding LDW drawings with no title sheet and no E.O.:  Sometimes there's an associated LEO drawing that not only lists the associated drawing numbers but also their titles.  When that happens, I've been using the title listed on the LEO.

Regarding LDW drawings with no title sheet and no E.O. and no relevant LEO:  Instead of leaving the title blank, I've been marking it "TBD".  I'm not sure that's better than leaving them blank, but it at least it has the advantage of making it obvious that it has been checked rather than accidentally omitted.

Just my 2 cents.  Feel free to continue as seems best to you.

7ofNine:
I was wondering myself about some of the situations you are describing below. I tried to follow as closely as possible to the actual title as it is given on the drawing but looking at them after  a while using something like “SEE DWG” is not really very informative. I’ll try to remember your suggestions for the the future.

rburkey2005:
I was doing the "see dwg" thing at first myself, and stopped doing it primarily because it was easier for me to let the python script fill in the titles automaticlaly.  :smile:  However, today I introduced a downloadable spreadsheet containing all the LEM drawing from all the boxes, which brings the problem into focus:  if the spreadsheet is sorted and you have 10 LEO280-296's in a row all having the title "see dwg" ... well, as you say, it's not as informative.

PR #1113, opened 2019-12-30
finalize titles box 502

7ofNine:


rburkey2005:
Thanks!

rburkey2005:
@7ofNine  Plus, this actually completes all of the titles for the boxes scanned _so far_.  Unfortunately, the scanner has been broken for a couple of weeks, and I haven't heard anything yet about when it will be repaired ... so even if you're inclined to continue with additional boxes, there's nothing left to do for a while insofar as drawing titles are concerned.

PR #1118, opened 2020-06-15
introducing CMake

scivision:
would perhaps help make building on many platforms easier, as that is the *raison d'être* of CMake.
I just built one target here to show how it would look-the top-level CMakeLists.txt invoking each directory and target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,6 @@
+cmake_minimum_required(VERSION 3.12)
+project(virtualAGC LANGUAGES C)
+
+set(NVER \\\"2019-09-22\\\")
+
+add_subdirectory(yaYUL)
--- a/yaYUL/CMakeLists.txt
+++ b/yaYUL/CMakeLists.txt
@@ -1,0 +1,19 @@
+cmake_minimum_required(VERSION 3.12)
+project(yaYUL LANGUAGES C)
+
+set(CFILES GetOctOrDec.c ParseBBCON.c ParseEQUALS.c ParseINDEX.c ParseXCADR.c yaYUL.c
+IncPc.c ParseBLOCK.c ParseERASE.c ParseInterpretiveOperand.c Pass.c yul2agc.c
+Parse2CADR.c ParseCADR.c ParseEqMinus.c ParseOCT.c PseudoToSegmented.c
+Parse2DEC.c ParseCHECKequals.c ParseEqualsECADR.c ParseSBANKEquals.c SymbolPass.c
+Parse2FCADR.c ParseEBANKEquals.c ParseGENADR.c ParseSETLOC.c SymbolTable.c
+ParseBANK.c ParseECADR.c ParseGeneral.c ParseST.c Utilities.c)
+
+add_compile_options(-Wall)
+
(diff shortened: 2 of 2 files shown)


rburkey2005:
Thanks, Michael!  It's a nice demonstration of concept, and I've merged it.  I certainly would like to have a cleaner, more bullet-proof way to build Virtual AGC on multiple platforms, and perhaps cmake would be a good way to go.  It's just (speaking for myself) that there aren't enough hours in the day to do even the things which are on my plate already, so there's no time left over for it.  Somebody else would have to take that ball and run with it.  :frowning_face:

scivision:
Great, I just wanted to see if CMake was an acceptable direction. I think implementing CMake incrementally might be the way to get it done, as trying to replace Make right away would be a big effort. However, just getting a basic simulation going via shouldn't be too bad, and build up from there.

scivision:
In that vein, what would be your suggestion for the next couple targets to enable via CMake? I thought yaYUL was the most fundamental of all, and it happed to be easy to do.

Thymo-:
yaAGC and yaDSKY2 are (I think) the most commonly used components. Those would be good candidates.

rburkey2005:
Yes, I'd agree with Thymo about the targets.

I would make this observation, even though perhaps it may reveal some (or a lot of) ignorance on my part about cmake.  The problem with the make system for Virtual AGC now isn't that it doesn't build on all the desired platforms ... it's that when those platforms themselves change later (such as XCode on Mac OS changing from X.Y to X.Y+1, or gcc changing from one version to the next, or whatever), the associated makefile system in Virtual AGC may stop working.  In my experience, getting stuff to start working again in situations like this is usually much easier if more-common tools are used than if less-common ones are used.  In other words, it's not the difficulty of setting up the system that's the biggest issue, it's the maintenance of the system, which may have to be performed years or even decades later by somebody who wasn't originally involved.  

To make it more specific, suppose a complete cmake system has been created to replace the make system in Virtual AGC.  And then it breaks.  Who is going to fix it?  Me?  I know nothing about cmake, and past experience shows that I'm unlikely to learn enough about it to be effective. You?  If you're committed to long-term maintenance, then great!  If it's just a fun side project that you expect to forget about after putting it in place, not so great.  :smile:  To guard against this, I would expect that both the make and the cmake systems might have to be kept in place, rather than the one replacing the other; in which case, is it even worth your effort doing the work in the first place?

Anyway, not trying to discourage you from doing it, but just pointing out some issues about radical change that sometimes may not immediately occur to people.

scivision:
I don't mean to replace Make necessary, but to provide a lower-maintenance way to build the most popular AGC targets. Generally CMake makes long-term maintenance across operating systmes easier because parameters can be less hard-coded vs. Make. 
For example instead of describing 5 different targets explicitly in Make, one target can be described covering all those platforms in CMake. Also the ability to cross-compile becomes generally more feasible with CMake.

I have a good amount of experience with CMake, so this is a bit of a feasibility check to see if it's useful to the AGC community, and if so then more effort can be spent on corner cases etc.

rburkey2005:
Great!

PR #1119, opened 2020-06-16
CMake: add yaAGC

scivision:
tested on Windows and Linux

yaAGC executable is created.

build/compile_commands.json is created, helping linters like Visual Studio, Visual Studio Code.

yaAGC.h includes winsock2.h before windows.h to avoid dozens/hundreds of warning messages

---

as for most CMake projects, build from the top virtualagc/ directory by:

```sh
cmake -B build
cmake --build build -j
```
Like Make, `-j` builds in parallel

The executables and libraries are created under build/ with the same hierarchy as the source directories e.g. build/yaYUL/yaYUL.exe   build/yaAGC/yaAGC.exe


rburkey2005:
I foolishly merged these in the wrong order (1120 first, and then 1119), thus causing a conflict that I had to resolve manually in CMakeLists.txt.  You might want to look at it to make sure it's that you were trying to put there.  I'm not much of a gitmaster, unfortunately.

scivision:
yes I think it's correct, thanks

Issue #1124, opened 2020-06-17
Reconstruct source files for Sundance that match binary dumps


thewonderidiot:
[The initial disassembly of our Sundance modules is now mostly complete.](https://github.com/thewonderidiot/sundance)

This issue is to track translation of this disassembly to proper source code files, separated into proper log sections. The final goal is produce code that assembles to exactly match our particular set of dumped modules:
B1: 292
B2: 302
B3: 302
B4: 302
B5: 292
B6: 306

...with the only binary differences being those caused by address offsets induced by revision changes, or by the addition of code that our module set is missing altogether. This will involve, roughly:

1. Making all necessary code changes.
2. Resolving all incompatibilities caused by shifting addresses between banks in modules of different revisions. 
3. Assigning names to all symbols that don't exist in Luminary.
4. Reconstructing pieces of code that are completely missing from our mix of modules.

I'm planning on documenting all major differences (new/deleted log sections, program rewrites, etc.) in comments here, but leaving out the hundreds of small differences that aren't as important. Wherever possible, names for things will be taken from Luminary memos or other original documentation; names will only be made up from scratch if we definitely have no other source.

thewonderidiot:
### Address Offsets By Revision

Because memory shifted around a lot between revs 292, 302, and 306, I plan to tabulate here all of the shifts that I am able to detect. This will be useful for symbol resolution purposes.

thewonderidiot:
### PADLOADS

This log section was introduced in Luminary 33, and so does not exist in Sundance.

thewonderidiot:
### CONTROLLED CONSTANTS

This log section was introduced in Luminary 52, and so it doesn't exist in Sundance either. This has a big impact on the assembly, as all of the constants that were concentrated into this section were distributed throughout the listing, mostly nearby code that uses them.

thewonderidiot:
### Missing Code

Unfortunately, with our particular mix of module revisions, there are a few pieces of Sundance code that are entirely missing. The ones I know about, so far, are:

- **UNK7766** This 10-instruction routine was added between Sundance 292 and Sundance 302. Luminary has no equivalent. It is called by several extended verbs, and appears to potentially have something to do with mark displays. My guess is that it is changing bits in EXTVBACT and/or FLAGWRD4.
[...]

PR #1127, opened 2020-07-18
Sundance reconstruction

thewonderidiot:
Here are reconstructed source code files for Sundance, put together using Luminary 69 as a base listing, and applying all differences identified in the [https://github.com/thewonderidiot/sundance](module disassemblies). Special credit to @indy91, who not only did all of the testing of it in NASSP, but also learned interpretive code, assisted with the disassembly, helped reimplement a missing function, and named a bunch of things. :)

This build was put together to match, as closely as possible, the binary dumps. Everything (with just a couple of exceptions) assembles to the same address address as in the dumps, which allows binary diffing of the assembly output with the original dumps.

Due to our mix of modules, and the relatively large number of changes between Sundance 292 and 306, our module set is mutually _very_ incompatible. These source code files assemble to have 827 differences from the dumps; each of these differences is listed and explained in [SundanceXXXDifferences.txt](https://github.com/thewonderidiot/virtualagc/blob/sundance_reconstruction/SundanceXXX/SundanceXXXDifferences.txt). Almost all of these differences are the result of address shifts in banks caused by the addition or deletion of code, as enumerated in [this table](https://github.com/virtualagc/virtualagc/issues/1124#issuecomment-645817548). As part of the build, the Makefile compares the output of `bdiffhead` to this file, to ensure that no unexplained differences between the build and the dumps exist.

Due to the version mix, in the source files, whenever a revision change occurs, I left comments in the form of `## Sundance 302`, which should hopefully help make it clear to anybody looking at the code which revision of Sundance they're looking at in that particular spot.

Sundance is an extremely interesting program. It is very much a "missing link" between Sunburst and Luminary. Its original release was closer to Sundisk than it was to Colossus, which makes it our earliest manned flight program. And it shows, in a lot of places!

Based on the GSOP, and what we've seen in other flight programs, we were expecting it to only contain things needed for the Apollo 9 mission. However, to our great surprise, the dumps unveiled extremely early-development descent, ascent, and abort programs. Even more surprising, it contains P10 and P11, two lunar launch time prediction programs that we previously thought never even made it to a released program. They work, but their display code is extremely buggy and causes a restart every time you run them.

A super early P57, the lunar surface alignment program, is also present. Unlike P10, P11, P12, and P63, it isn't even startable via V37. This is presumably because, from what I can tell, it is the buggiest of all -- it has multiple places where interpretive code runs straight into basic or vice-versa, which is a recipe for disaster. It's quite unusual to see bugs so bad in flown programs!

One of the more interesting differences is in the lunar landing guidance equations. The tables used there were completely restructured between Sundance and Luminary. In Luminary, there are many tables per entry point or parameter, with each entry in a table corresponding to a landing phase. In Sundance, [there was instead a table per phase](https://github.com/thewonderidiot/virtualagc/blob/sundance_reconstruction/SundanceXXX/LUNAR_LANDING_GUIDANCE_EQUATIONS.agc#L38), with entry points and parameters mixed together.

But perhaps the most interesting thing is that the landing and ascent code _works_. For the most part, at least -- you have to manually toggle some flagbits and if you don't PRO to enable redesignations in P64 you get a restart in P65. And it's much more sensitive to large delta-H than Luminary. But despite that, @indy91 was able to perform a nice soft landing.

Interestingly, not everything works in lunar orbit in Sundance. Verb 82 (orbital parameter display) and verb 83 (rendezvous parameter display) both display complete garbage when around the moon, for reasons we're not entirely sure. Both received quite a few code changes, though, even for Colossus 237. But it's interesting that key programs like these were nonfunctional around the moon this late.

Anyways, back to these source files. For the most part, our particular mix of modules was quite good. Everything DAP-related is revision 302 (which is nice since we have the Sundance 302 DAP study guide, and there were many DAP changes between 292 and 302). All of the landing/ascent/abort code is revision 292, but this is actually fine because we know for sure that none of this changed between 292 and 306. And we have revision 306 for the verbs and nouns, so the astronaut-facing interface exactly matches that in the flown revision.

There are, sadly, two missing functions:

- **UNK7766** -- this function resides in fixed-fixed memory (for which we only have rev 292), and was added in Sundance 302. It is only called from extended verbs. We believe it either messes with EXTVBACT, or does something with marks, but we're not sure what, yet. For this build I have it equated to `TCQ` so that it performs nothing.
- **TIGINT** -- this function performs orbit integration to TIG-30 for BURNBABY and to TIG for P47. It was located in bank 36 in rev 292 (we have rev 306), and in bank 32 in rev 306 (we have 292) -- which means we don't have a copy of it at all. Its functionality was replaced with MIDTOAVE in Colossus and Luminary. @indy91 and I have supplied our own implementation for this build, which, while shorter than the original, appears to work fine in NASSP. We believe it's fully working because when using it, burns properly update the state vector. An implementation of this function that does nothing results in a completely corrupted state vector.

Now that this is complete, my plan is to make a build of Sundance that gets as close to revision 306 as possible, by pulling changes back from Luminary that we know (from bank offsets or GSOPs) went into Luminary. This isn't just for fun -- there were some bugfixes in P32 that NASSP needs to be able to properly follow the checklists. @rburkey2005, do you have a suggested name for this? I've been tentatively calling it SundanceYYY locally but am very open to suggestions.

As always, please feel free to change comments in the headers or Main.annotation or anywhere else. :)


rburkey2005:
@thewonderidiot I was thinking about the naming as I read the top part of your comment, though the answer that had formed in my head is probably too flip.  What occurred to me was SundanceMongrel.

A more-sensible, pedestrian answer might be something like Sundance292-306.

rburkey2005:
@thewonderidiot I think you should go ahead an merge it (with a name change, if you want) yourself, because there's no way I can hope to bring myself up to speed overnight enough to think of overriding anything you've done.  In particular, I'm sure whatever name you want is fine.  If I think anything needs tweaking, I can certainly do it later.  And I can look into writing it up for the website as well.

And by the way, great work!

rburkey2005:
@thewonderidiot I want to begin working on the web-page writeup, so I'm thinking about merging this myself (perhaps first thing tomorrow) and keeping the name "SundanceXXX".  It occurs to me that once I do that, I'd prefer not to have any name-changes later, since I don't want to have to search-and-destroy all of the hyperlinks to SundanceXXX that I'll be making.  SundanceXXX is probably as good a name as any other.  So if there's any feedback on these notions, today would be a good time for it.  :smile:

thewonderidiot:
Haha, yeah, I haven't merged it yet because I've been pondering on the name. Not so much just the name of this one, but names for both this build that is a mix of revisions, and the upcoming speculative 306 build that won't quite be 306. Ideally that would make the difference clear. Maybe keeping this SundanceXXX and the next one... Sundance306ish?

rburkey2005:
That sounds very sensible to me.Sent on my Virgin Mobile Samsung Galaxy S® 5

PR #1130, opened 2020-09-07
Mechanical

ManoDaSilva:
The main housing is finally finished. It took me a while... The drawings are very often ambiguous, lines were overlapping.... But it should match the other parts without issues.


rburkey2005:
Thanks, much appreciated!

PR #1131, opened 2020-09-08
More parts! 

ManoDaSilva:
Worked on the elusive Universal DSKY Handling Fixture, mainly from the only assembly diagram and photographic evidence, which allowed me to get a somewhat accurate model of the panel and brackets. 
The handle is a parametric drawing which was available on the G&N drawing database. 

Also finished two gaskets: main housing and rear housing.


rburkey2005:
@ManoDaSilva Thanks!

PR #1132, opened 2020-11-02
Prevent compile warning in Visual Studio

Thymo-:
We're doing some work on getting NASSP and virtualagc synced up again. See #622 

Visual Studio has deprecated functions linke scanf, strcat, strcpy, etc. because they can be dangerous if you don't know what you're doing. Of course we know what we're doing (right?) and want to keep portability with other OS' so we can safely disable this as we already have in NASSP.


rburkey2005:
Thanks!

Issue #1134, opened 2020-12-04
Reconstruct Comanche 044 from Comanche 051


thewonderidiot:
### Branch
[comanche044](https://github.com/virtualagc/virtualagc/tree/comanche044)

### Changed Bugger Words from Closest Program (Comanche 051)
```
Bugger word mismatch in bank 02; actual 51322 != expected 73065 (diff = 21543)
Bugger word mismatch in bank 01; actual 50017 != expected 50116 (diff = 00077)
Bugger word mismatch in bank 04; actual 53747 != expected 62734 (diff = 06765)
Bugger word mismatch in bank 05; actual 74460 != expected 74466 (diff = 00006)
Bugger word mismatch in bank 06; actual 63024 != expected 63035 (diff = 00011)
Bugger word mismatch in bank 07; actual 73341 != expected 73403 (diff = 00042)
Bugger word mismatch in bank 11; actual 44621 != expected 44661 (diff = 00040)
Bugger word mismatch in bank 12; actual 51003 != expected 51215 (diff = 00212)
Bugger word mismatch in bank 13; actual 74727 != expected 74770 (diff = 00041)
Bugger word mismatch in bank 14; actual 56756 != expected 47777 (diff = 06757)
Bugger word mismatch in bank 15; actual 67323 != expected 70265 (diff = 00742)
Bugger word mismatch in bank 16; actual 55656 != expected 70262 (diff = 12404)
Bugger word mismatch in bank 17; actual 65302 != expected 50463 (diff = 14617)
Bugger word mismatch in bank 20; actual 77241 != expected 77251 (diff = 00010)
Bugger word mismatch in bank 22; actual 65672 != expected 47471 (diff = 16201)
Bugger word mismatch in bank 24; actual 55773 != expected 56001 (diff = 00006)
Bugger word mismatch in bank 25; actual 72651 != expected 72660 (diff = 00007)
Bugger word mismatch in bank 26; actual 64710 != expected 51445 (diff = 13243)
Bugger word mismatch in bank 30; actual 40077 != expected 40124 (diff = 00025)
Bugger word mismatch in bank 31; actual 66412 != expected 66431 (diff = 00017)
Bugger word mismatch in bank 32; actual 66036 != expected 65511 (diff = 00325)
Bugger word mismatch in bank 33; actual 54404 != expected 54370 (diff = 00014)
Bugger word mismatch in bank 34; actual 52350 != expected 52401 (diff = 00031)
[...]

PR #1135, opened 2020-12-06
Reconstructed Comanche 44

thewonderidiot:
Here's an attempt at reconstructing Comanche 44, as documented in #1134. I've started with 44 instead of 45 because the change between the two was done in a special way to make sure only module 3 was affected, and I haven't figured out exactly how they did it yet.

All of the banksums match up, and I feel pretty good about most of the changes, but before you merge this you should read the final comment on #1134. For all of the reconstructions so far, I've been able to argue the changes I've made with concrete evidence, but for this one change I have a more flimsy "well, maybe the changed the pairing of these opcodes while they were changing out the variable... it does at least make the banksum right...". @indy91 and I haven't been able to come up with anything better, though.

If this seems acceptable to you, I'll open up another ticket on figuring out rev 45 from this. Once we get to 45, getting to the flown 45 rev 2 should be pretty much free, since I believe that change exactly matches that done for 51 to 55.


rburkey2005:
It may be "go fever" talking, but I'm willing to accept it, particularly since it doesn't break anything.  BTW, if you haven't already, look at _my_ final comment in #1134.

rburkey2005:
@thewonderidiot Note that I slightly modified the annotations file, so be sure to pull it if you want to make any changes yourself.

rburkey2005:
Plus, I deleted the comanche044 branch, since GitHub suggested doing so, though it could apparently be restored.

PR #1139, opened 2020-12-24
Reconstructed MANCHE45 rev 2

thewonderidiot:
Finally! This differs from Comanche 45 in the following files:

- ERASABLE ASSIGNMENTS
- TAGS FOR RELATIVE SETLOC AND BLANK BANK CARDS
- ORBITAL INTEGRATION
- RTB OP CODES

All of the changes are identical to the R-2 lunar gravity model changes made between Comanche 51 and 55, and so everything has been lifted directly from the Comanche 55 listing, without any changes.


rburkey2005:
@thewonderidiot The website is updated for Comanche 45/2, and the annotations have been added.  I was starting to fade quite a bit at the end, so it probably wouldn't hurt to look them over.

thewonderidiot:
They look good to me! Thanks for taking care of all of that, and merry Christmas! :smile:

PR #1142, opened 2021-04-04
Fix various symbol name and other minor typos

smithery1:
I've written an AGC assembler to learn a little more about the computer. It's based solely on my reading of the original MIT/IL documentation that you've collected, but of course takes your yaYUL-formatted source files as input. 

I've run it on at Luminary099 and Comanche055 so far. It generates complete and correct binary output for those and, for fun, it also outputs end-of-listing information that mimics the GAP output on the original scans. I've spot-checked those listings, which mostly involves comparing entries between my output and the scans, or the first and last entry on each page for large tables, or the first and last entry on every Nth page for very large tables. Then I've fixed any discrepancies that appear, which is the source of this PR.

There are 10 location symbol adjustments, 5 COUNT name changes, and 2 miscellaneous changes.

The symbol differences are mostly minor transcription errors in symbol names, causing them to appear out of order in the EBCDIC-sorted informational tables. I've verified the proper symbol names by the symbol's location in the original tables, and by staring closely at the scans for those cases where the sort problem is a result of a character change to one with a similar glyph (ZEROD to ZEROO or P00DAPAD to POODAPAD for example.) I've also added a couple of unreferenced symbols that were omitted from the transcriptions.

The COUNT operand changes are mostly typos or transcription errors similar to the ones above. There is one interesting one in Comanche055/IMU_CALIBRATION_AND_ALIGNMENT.agc that appears to be a typo in the original code that was "corrected" in transcription - "$$/P03" with a zero appears in several places elsewhere in the original, but here it should be "$$/PO3" with the letter O.

I also added a empty page comment for page 882 in Comanche055, and fixed the printout date in Luminary099/Main.annotation.

These don't have any effect on the binary output, of course, and I've verified that with a yaYUL run as well to ensure I didn't accidently break anything. And it's not an exhaustive check of all symbol names. But I think it's nice to have for consistency and the changes here help my assembler reproduce some of the original GAP output.

### DIFF
--- a/Comanche055/ANGLFIND.agc
+++ b/Comanche055/ANGLFIND.agc
@@ -15,6 +15,8 @@
 ##				CADR ZEROEROR.
 ##		2017-01-07 RSB	Fixed comment errors detected in cross-diff vs
 ##				Colossus 237 and Colossus 249.
+##              2021-04-04 ABS  Fixed minor symbol name transcription errors as
+##                              verified by sort order in end-of-listing tables.
 ##
 ## This source code has been transcribed or otherwise adapted from digitized
 ## images of a hardcopy from the MIT Museum.  The digitization was performed
@@ -425,7 +427,7 @@ LOOPSIN		SLOAD*	RTB
 		STADR
 		STORE	16D,2		# C8=-SIN(THETA)SIN(PHI)SIN(PSI)
 		RVQ			#    +COS(THETA)COS(PHI)
-ENDOCM		EQUALS
+ENDDCM		EQUALS
 
 		BANK	15
 		SETLOC	KALCMON1
@@ -622,7 +624,7 @@ VECOFANG	VDEF	RVQ
 
 NOGOM2		INHINT			# THIS LOCATION ACCESSED BY A BZMF NOGO -2
 		TC	BANKCALL
-		CADR	ZEROERROR
+		CADR	ZEROEROR
 		
 NOGO		INHINT
 		TC	STOPRATE
--- a/Comanche055/ERASABLE_ASSIGNMENTS.agc
+++ b/Comanche055/ERASABLE_ASSIGNMENTS.agc
@@ -32,7 +32,9 @@
 ##		2017-01-22 RSB	Fixed comment errors detected in cross-diff vs
 ##				Artemis 72.
 ##		2017-03-16 RSB	Fixed a ##-style comment.
-
+##              2021-04-04 ABS  Fixed minor symbol name transcription errors as
+##                              verified by sort order in end-of-listing tables.
+##
 ## This source code has been transcribed or otherwise adapted from digitized
 ## images of a hardcopy from the MIT Museum.  The digitization was performed
 ## by Paul Fjeld, and arranged for by Deborah Douglas of the Museum.  Many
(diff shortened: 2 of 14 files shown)


rburkey2005:
@smithery1 Thanks!  I haven't looked at this in detail yet, but I'm undecided exactly what to do with it.  I mean, the problems will definitely be fixed, but I'm unsure what to do with your patch. The problem is that misnamed symbols tend to propagate throughout multiple versions of the code base (of which we now have a large number, including some in unmerged local branches).  So a correction to just a couple of code versions really doesn't do too much to fix the overall problem, as it may leave 5-10 (or more) versions uncorrected.  It's an odd situation, in which having a list of the misnamed symbols is almost more useful than having a patch for a couple of versions, since it makes it easier to perform a search-and-destroy on all code versions.  I'll have to think about it for a while to figure out an appropriate action.

smithery1:
Ah, I see and half-suspected as much. By code base, I assume you mean the various AGC source code releases? And that there are some releases that are not in github yet? I'm happy to continue to contribute to this if I can.

Perhaps a tool of some sort would be useful to point out potential discrepancies across versions. It would still require someone to check its output and make the changes, but it might be able to narrow down the search. To reiterate, the issues I saw so far fell into three basic camps - location symbols, COUNT card operands, and a "## Page" comment for a blank page. Since I dump out this information and more anyway, I can think on that and see if I can come up with something as I move across assembling the other versions.

What I'll try to do too is come up with a reasonably complete list, rather than the spot-checking I've done here. If there's value to updating the symbols, we might as well do all we can.

Since this is a longer-term goal, I can close this PR and move discussion into an issue, where I can update occasionally as things come along until there's consensus and occasion to merge. We can also have a master list of symbol changes at that time, for use in local branches. Or if you want to consider it or have other ideas, just let me know.

smithery1:
So musing aside, here's what I propose.
1. I close this PR and open an Issue instead
2. Over some period of time I gather as complete a list of symbol and other such changes as I can across all AGC code bases in main
3. We review any potential complications as they come up and the full list once gathering is complete in the Issue page
4. If all is ok, I prepare and submit a PR for the changes to main
5. The complete list is available in the Issue page for anyone to apply to local repos

rburkey2005:
Yes, in talking about the different versions of the AGC software, I mean things like Colossus 237, Colossus 249, Comanche 44, Comanche 45, Comanche 45/2 (MANCHE45R2), Comanche 51, Comanche 55, ....  They have all already been made consistent with each other, by extensive diff's and lots of manual effort.  No additional diff'ing (or any other tool or effort) is needed to accomplish that.  Except that if you fix (say) Comanche 55, you now have to go and fix 237, 249, 44, 45, ....  And I suppose there could still be inconsistencies (and thus errors) that slipped past in spite of everything, which could be detected by someone looking at it from a fresh perspective.

As far as I know, there's only one AGC version in progress that hasn't been merged into the repository so far, and that's a reconstruction of Comanche 67.  It's in the repository, but you have to use the "comanche67" branch rather than the main branch to see it.  I've been thinking lately about merging it into main, in spite of the fact that it's not really ready.  That's another factor.

In so far as the problems you've discovered already, I could of course go through your patch and Identify all of the changes in order to clean up the other software versions.  All I really meant is that I wasn't sure whether I should merge the patch and then try to fix the other software versions, or whether I should reject the patch and then clean all of the software versions.  I haven't finished thinking about it yet, but the latter option (rejection) doesn't seem to have many benefits over the former option.

But yes, if you want to try and get together a more-comprehensive list of symbol typos across a lot more AGC versions before merging fixes, that would probably be even better.

smithery1:
Thanks, it makes sense. I think by a tool I meant something to help me identify if a symbol was present across multiple versions, so I could more easily identify which other versions were affected by a fix. But that may not be necessary, and I'll figure it out at any rate.

I'll also make a note to check the commanche67 branch, to avoid having this affect its merge schedule.

I feel like if I'm going to suggest these changes, the polite thing to do is to take care of it across the board as best I can. I'll do that and check back in in a bit.

PR #1144, opened 2021-05-24
Add WebAssembly compilation target for yaAGC

michaelfranzl:
This pull request contains a minimal set of changes which adds a new compilation target for `yaAGC`: WebAssembly (Wasm). The resulting executable `yaAGC.wasm` is about 130 kB in size and is very portable. It will execute in all Wasm runtimes, which include, but are not limited to, internet browsers. This opens up the interesting possibility of distribution over the internet and client-side execution in most browsers, but also of embedding into a wide array of other contexts.

I also created a JavaScript wrapper (called `webAGC`) for the resulting executable `yaAGC.wasm` (see https://github.com/michaelfranzl/webAGC). The source code for a demonstration is also part of it (see https://github.com/michaelfranzl/webAGC/tree/dev/demo). A live version of the demo is hosted on my GitHub page https://michaelfranzl.github.io/webAGC/demo/ . (Please note that, to keep the JavaScript development workflow simpler, this demo currently requires a modern browser supporting "import maps", e.g. based on Chromium 89 or newer. But this is just a limitation of the demo itself, and not of any of the used components).

The `ChannelOutput` and `ChannelInput` functions were implemented to work with fast ringbuffers to eliminate the possibility of slow foreign function calls. Six functions were added as entry points into the WebAssembly context: to read from and write to mentioned ringbuffers, to set the fixed memory and read the erasable memory, and to reset and step the CPU. All code related to network sockets had to be disabled using pre-processor directives because there is no notion of networks in WASI/WebAssembly.

If the `printf`'s in the yaAGC source code were disabled, the executable would no longer depend on any WASI system calls (like `fd_*`), which would make it even more portable (and about 2 kB smaller).

My work was inspired by the project https://github.com/siravan/moonjs , but provides an independent implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,13 @@ On Fedora 22 or later you may encounter that the wxWidgets doesn't have the wx-c
 * You will need SUNWgnome-common-devel, SUNWGtk, SUNWxorg-headers, FSWxorg-headers, SUNWncurses, SUNWtcl, SUNWtk and SUNWlibsdl
 * You will also need GNU readline 6.0, wxWidgets 2.8.9 (with `configure --disable-shared`), Allegro 4.2.2 (with "configure --enable-shared=no --enable-static=yes") and to put `/usr/local/bin` and/or `/usr/local/bin/wx-config` linked into your `PATH`.
 
+## WebAssembly
+
+* Requires `clang` from the LLVM project, plus a C standard library which compiles down to WASI system calls. Clang can directly emit WebAssembly since version 8.  [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) provides a WebAssembly toolchain (clang plus C/C++ standard libraries).  A build from the source of [wasi-sdk revision a927856 ("use llvm 12.0.0 release")](https://github.com/WebAssembly/wasi-sdk/tree/a927856376271224d30c5d7732c00a0b359eaa45) has been tested with the Virtual AGC components listed in the section "WebAssembly" below, but the project also [supplies pre-built packages](https://github.com/WebAssembly/wasi-sdk/releases) for various platforms which should work just as well. In the build scripts of Virtual AGC, it is assumed that `wasi-sdk` is installed at `/opt/wasi-sdk`, but you can customize this path (see section "WebAssembly" below).
+* Requires `wasm-opt` (from the `binaryen` package) for WebAssembly code optimization.
+* Requires `wasm-strip` (from the `wabt` package) to minimize the size of the WebAssembly code.
+* Optionally, `wasm2wat` (from the `wabt` package) to translate binary code to human-readable text representation.
+
 
 More information at http://www.ibiblio.org/apollo/download.html#Build
 
@@ -254,6 +261,37 @@ To match the default setup of the installer program execute the following:
 You can make a desktop icon called *Virtual AGC* that links to `/VirtualAGC/bin/VirtualAGC`. The image normally used for the desktop icon is found at `/VirtualAGC/bin/ApolloPatch2.png`.
 
 Unfortunately the ACA simulation (joystick) programs do not work in this environment.
+
+
+## WebAssembly
+
+In the Virtual AGC build scripts, it is assumed that `wasi-sdk` is installed
+at `/opt/wasi-sdk`. You can customize the path by setting
+`WASI_SDK_PATH=/path/to/wasi-sdk` on the command line when executing `make`.
+
+For all builds to WebAssembly, put `WASI=yes` before `make`.
+
+Currently, only the following Virtual AGC components can be compiled for the
+WebAssembly target:
+
+### yaAGC
+
+If you have any leftover build artifacts in the `yaAGC` directory, run
+`make clean` in it.
+
+To build, simply `cd` into the root directory and do:
+
+`WASI=yes make yaAGC`
+
+This will produce `yaAGC.wasm` (about 100 kB).
+
+To additionally get a text representation, go into the `yaAGC` directory and
+run:
+
+`make yaAGC.wast`
+
+This will produce `yaAGC.wast` (about 900 kB).
+
 
 # Endnotes
 
--- a/yaAGC/Makefile
+++ b/yaAGC/Makefile
@@ -76,6 +76,7 @@
 #				else's builds.
 #		2016-11-12 RSB	"Permanently" disabled READLINE.
 #		2016-11-18 RSB	Explicitly include socket library in Solaris.
+#		2021-05-13 MKF	Defined WASI and targets yaAGC.wasm, yaAGC.wast.
 
 LIBS=${LIBS2}
 
@@ -84,7 +85,11 @@ PREFIX=/usr/local
 endif
 
 .PHONY: default
+ifdef WASI
+default: yaAGC.wasm
+else
 default: yaAGC
+endif
 
 #---------------------------------------------------------------------------
 # The use of libreadline adds a command-history to yaAGC, but may have some
@@ -124,6 +129,30 @@ endif # MACOSX
 endif # NOREADLINE
 
 CFLAGS2 += -DGDBMI
+
+ifdef WASI
+WASI_SDK_PATH ?= /opt/wasi-sdk
+
+CFLAGS2 = \
+	--target=wasm32-wasi \
+	--sysroot ${WASI_SDK_PATH}/share/wasi-sysroot \
+	-DWASI=yes \
+	-O3 \
+	-flto \
+	-fwhole-program-vtables \
+	-fvirtual-function-elimination
+
+LDFLAGS2 = \
+	--no-entry \
+	--export-dynamic \
+	--unresolved-symbols=import-functions \
+	-L ${WASI_SDK_PATH}/share/wasi-sysroot/lib/wasm32-wasi \
+	-lc \
+	--export=malloc \
+	--export=free \
+	--import-memory \
+	--lto-O3
+endif
 
 # Alberto Galdo's iPhone mod.
 ifdef IPHONE
@@ -183,6 +212,14 @@ yaAGC:	$(OBJECTS) checkdec.o libyaAGC.a ${NATIVE_WINAGC}
 	${cc} ${CFLAGS} ${CFLAGS2_NATIVE} -o yaAGC ${OBJECTS} checkdec.o \
 	libyaAGC.a -L. ${STATIC} ${LIBS} -lpthread -lyaAGC -lm ${CURSES}
 
+yaAGC.wasm: wasm.o agc_engine.o agc_engine_init.o agc_utilities.o ringbuffer_api.o ringbuffer.o
+	${WASI_SDK_PATH}/bin/wasm-ld ${LDFLAGS2} -o $@ $^
+	wasm-opt $@ -Oz -o $@
+	wasm-strip $@
+
+yaAGC.wast: yaAGC.wasm
+	wasm2wat $< > $@
+
 libyaAGC.a: ${LIBOBJECTS}
 ifdef IPHONE
 	${IPHONE_DEV}/usr/bin/ar -rc $@ $^
@@ -195,7 +232,7 @@ endif
 	touch ../yaDEDA/src/main.c
 
 clean:
-	rm -f yaAGC libyaAGC.a *.o *~ *.bak *.elf *.o68 *.o8 *.rel *.exe *-macosx
+	rm -f yaAGC libyaAGC.a *.o *~ *.bak *.elf *.o68 *.o8 *.rel *.exe *-macosx *.wasm *.wast
 
 install:	yaAGC
 	cp yaAGC ${PREFIX}/bin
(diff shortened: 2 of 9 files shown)


rburkey2005:
Terrific, Michael, thanks!  I've merged in your code.  I'll look into writing up something for the website in the next few days.  I haven't given any thought to the `printf` suggestion as of yet.

rburkey2005:
@michaelfranzl By the way, I've played a little with your live demo, and I like it very much.  However, I've never found anything that I liked so much that I couldn't think of a few irritating suggestions to make about improvements :smile:, so here are a couple of observations.  Obviously, feel free to ignore them!

- I like the view of how erasable memory is changing, though I think that you'd have a more-authentic Apollo experience if you used 5-digit octal rather than 4 hex digits to display it.  It's true that that would only show 15 bits, but since the 16th memory bit is almost always parity, I don't think you lose anything valuable in not displaying it.
- I think it would be nice to break out a view of the CPU registers separately:  A, L, Q, etc.  It's true that the registers mostly correspond to the first few words of bank 0, of course, but naming them (and seeing the program counter) would still be useful.
- And some minimal usage instructions and explanation of whatever "Load program" is chosen might be nice.  Or at least a hyperlink back to somewhere that might have that kind of material.
- I notice that the "DSKY key input" requires lower-case characters and doesn't respond to upper-case ones.  Might be good to allow both.
- I think the Halt button is more of a Pause.  Is that right?  I think of Halt as completely ending execution of a program, something like Pause+Reset.

I have more ideas too, but it's easy to keep making more and more suggestions when you want _somebody else_ to implement them, so I'll restrain myself to these few.

Again, though, great job!

michaelfranzl:
@rburkey2005 thanks for the positive feedback, and for the improvement suggestions! Points 1, 4 and 5 are done. I also added auto-typing after pasting key strokes into the input field. The live demo is updated (you may need to do a full browser page reload holding Shift to refresh the browser cache). The break-out of the registers, and some program documentation already were on my to-do list, and I plan to do that soon.

rburkey2005:
So ... I'm trying to write this up on the website, and specifically am working on build instructions.  Which aren't as obvious to me as they make have been to you.  :smile:   At any rate, I have errors I don't understand.

I have installed wasi-sdk 12 (in ~/Desktop/wasi-sdk-12.0/), and have built the latest versions of wabt and binaryen from source, installing their executables in /usr/local/bin (which is in my `PATH`).  I don't suppose it matters, but this is in Linux Mint 19.1, 64-bit.  

From the yaAGC/ folder, I run the following command
<pre>
WASI_SDK_PATH=~/Desktop/wasi-sdk-12.0 WASI=yes make yaAGC.wasm yaAGC.wast
</pre>

Here are the messages I get:
<pre>
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o wasm.o wasm.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o agc_engine.o agc_engine.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o agc_engine_init.o agc_engine_init.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o agc_utilities.o agc_utilities.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o ringbuffer_api.o ringbuffer_api.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/clang --target=wasm32-wasi --sysroot /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot -DWASI=yes -O3 -flto -fwhole-program-vtables -fvirtual-function-elimination -c -o ringbuffer.o ringbuffer.c
/home/rburkey/Desktop/wasi-sdk-12.0/bin/wasm-ld --no-entry --export-dynamic --unresolved-symbols=import-functions -L /home/rburkey/Desktop/wasi-sdk-12.0/share/wasi-sysroot/lib/wasm32-wasi -lc --export=malloc --export=free --import-memory --lto-O3 -o yaAGC.wasm wasm.o agc_engine.o agc_engine_init.o agc_utilities.o ringbuffer_api.o ringbuffer.o
wasm-ld: error: unknown argument: --unresolved-symbols=import-functions
Makefile:216: recipe for target 'yaAGC.wasm' failed
make: *** [yaAGC.wasm] Error 1
</pre>

All of the clang and llvm functions are version 11 in wasi-sdk12, as they should be, and the command "`wasm-ld --version`" confirms this.  The command "`wasm-ld --help`" does not list any `--unresolved-symbols` command-line option.

Any ideas?  Is it possible you have llvm 12 installed rather than the llvm 11 found in wasi-sdk 12?

michaelfranzl:
The missing `--unresolved-symbols` switch indicates to me that you are using an older version of `wasi-sdk` than I did. I should have specified in the `README` of this pull request exactly which version I used; apologies for the inconvenience. I did not download one of the release packages, which it seems you did; I built `wasi-sdk` from sources, from the currently latest commit which is `a927856376271224d30c5d7732c00a0b359eaa45`.

This is how to compile `wasi-sdk` from sources:

```sh
# Build a specific commit of wasi-sdk
git clone --recurse-submodules https://github.com/WebAssembly/wasi-sdk.git # about 1.5 GB download
cd wasi-sdk
git checkout a927856376271224d30c5d7732c00a0b359eaa45 # use llvm 12.0.0 release
make
```

You need not `make install` after that. You could simply set an absolute path to the build directory `WASI_SDK_PATH=/path/to/wasi-sdk/build/install/opt/wasi-sdk`.

rburkey2005:
Well, [you specifically said](https://github.com/virtualagc/virtualagc#webassembly) that you used wasi-sdk 12.0.0.  That's the latest release listed on the wasi-sdk website, and as my comment stated, it's the one I used.  It's true I used the prebuilt release, but it's hard to see why a prebuilt 12.0.0 would differ from 12.0.0 built from source.  Nevertheless, I'll try building from source anyway.

And I did not do a `make install`.  As my comment stated, I set WASI_SDK_PATH on the 'make' command line.

rburkey2005:
Okay ... so to continue beating a dead horse, you did _not_ use wasi-sdk 12, the latest release, but instead used not-yet-released wasi-sdk code.

Furthermore, my previous comment was wrong:  You never said you had used wasi-sdk 12; rather, you said you had used llvm 12.

I guess the summary of that is that you were accurate, but that I was confused by the versioning of the wasi-sdk releases.  Yikes!

Well, I guess that's good in some ways.  At least in my writeup I can warn about what _not_ to do.  :smile:

rburkey2005:
Okay, it seems to work fine.  I'll update the website's download page with instructions.

Thanks!

rburkey2005:
@michaelfranzl I may have spoken too quickly.  Not about building yaAGC.wasm, which works as far as I can tell, but about running the webAGC demo from my local computer rather than from your live-demo site.

Along the way, I encountered a few difficulties in following the instructions, but worked around them, so I won't bore you with my problems. But I finally got to the point of firing up Chrome and navigating to http://localhost:8000/demo/.  I was using your webAGC code, unaltered, including your version of yaAGC.wasm.

The demo page appeared, along with the image of the DSKY, but there was no octal display, and the simulated AGC program (Validation) didn't appear to run.  I looked at the javascript errors; here's a screenshot, since I'm too lazy to type in what it said:

VirtualBox_Linux Mint 19

The error location is simply the end of your `main()` function, and thus isn't very meaningful.  At least, not to me, with my minimal javascript skills. 

Any thoughts?

michaelfranzl:
The relevant part in the error message is: "Incorrect response MIME type. Expected application/wasm". As a security measure, browsers will only run WebAssembly files which are served with the HTTP response header `Content-Type: application/wasm`. I would bet that in your case, the HTTP server doesn't recognize that `yaAGC.wasm` is a WebAssembly file, and so it simply assumes a general-purpose type for binary files, `Content-Type: application/octet-stream`. You could use the browser's Dev Tools to verify this. This cannot be fixed from within the source code, it is an operations problem.

In the README of the webAGC demo, I made the shortcut of just saying "serve the parent directory via HTTP", but it seems that I need to be more precise there to avoid errors like this.

Because you must already have `npm` available, it is probably easiest if you use the `http-server` package. It should recognize the file type correctly:

```sh
npm install --global http-server
cd path/to/webAGC
http-server # it prints the served URL(s) to use
```

As an alternative you could use `ruby` version 3.0.1, with the `webrick` gem, running `ruby -r un -e httpd . -p 8000`. This is what I used during development.

If you run into further problems with the demo, I would suggest that we use the issue tracker over at https://github.com/michaelfranzl/webAGC/issues to keep this pull request and the demo development separate.

rburkey2005:
@michaelfranzl Now that that's all working, I've updated [the website's download page](http://www.ibiblio.org/apollo/download.html#WebAssembly) accordingly.  As you may have noticed, I'm not always 100% accurate (in spite of my most-sincere efforts).  So if you're at all interested, you might glance at it for fact-checking purposes.  I suppose the most-recent couple of entries in [the change log](http://www.ibiblio.org/apollo/changes.html) might benefit from a little scrutiny too.

michaelfranzl:
> So if you're at all interested, you might glance at it for fact-checking purposes.

Gladly.

> [...] but to run much faster than if they were instead ported to JavaScript

and

> However, the WebAssembly version should be somewhat faster than the JavaScript version of yaAGC.

WebAssembly instructions run at near native speed (I read somewhere); but it still runs in the same virtual machine as JavaScript. And JavaScript nowadays is also just-in-time compiled and optimized to native or near-native machine instructions; so there should not be much difference at runtime. Same for asm.js. The big difference is the efficient binary representation (i.e. faster download) and the much higher parsing speed of WebAssembly.

> However, I'm told that you really want at least ruby 3, which in Mint 19 requires you to build the ruby language from source.

This was incomplete information from my side. As mentioned here https://github.com/michaelfranzl/webAGC/issues/1#issuecomment-850590564 , just `gem install webrick` will install the newest version (currently 1.7.0), in any Ruby version, and it will serve the .wasm file correctly.

> Now, when you look at these instructions, and in particular note that there is a client-server setup, you may wonder why I've said that webAGC is a browser-only solution.  I don't really know the answer to that at the moment.  I think that if you were developing your own web app based on webAGC, it could be done without requiring a server at all.

I'm not sure if I understood this correctly, but I'd like to try to give an answer:

Browsers can read files even from hard disk and display static websites (html, js, css) just like, say, a word processor displays a word document. Strictly speaking, no HTTP servers are required in that scenario. However, when using more modern browser features, like JavaScript modules, WebAssembly, WebRTC etc., the browser insists that content be served over the network. This doesn't mean that servers need to have any 'business logic' implemented; they can just be static content delivery pipes. In this sense, the `webAGC` demo is indeed a pure browser/client solution; it is completely unaware of a server.

rburkey2005:
Thanks for checking!
> > [...] but to run much faster than if they were instead ported to JavaScript
> 
> and
> 
> > However, the WebAssembly version should be somewhat faster than the JavaScript version of yaAGC.
> 
> WebAssembly instructions run at near native speed (I read somewhere); but it still runs in the same virtual machine as JavaScript. And JavaScript nowadays is also just-in-time compiled and optimized to native or near-native machine instructions; so there should not be much difference at runtime. Same for asm.js. The big difference is the efficient binary representation (i.e. faster download) and the much higher parsing speed of WebAssembly.

Oops!  I meant to change "much faster" to "somewhat faster" everywhere, but it looks like I missed one place.
> 
> > However, I'm told that you really want at least ruby 3, which in Mint 19 requires you to build the ruby language from source.
> 
> This was incomplete information from my side. As mentioned here [michaelfranzl/webAGC#1 (comment)](https://github.com/michaelfranzl/webAGC/issues/1#issuecomment-850590564) , just `gem install webrick` will install the newest version (currently 1.7.0), in any Ruby version, and it will serve the .wasm file correctly.

I remember writing about ruby 3, but the download page doesn't seem to say that any more.  I guess I must have corrected it before you looked at it.
> 
> > Now, when you look at these instructions, and in particular note that there is a client-server setup, you may wonder why I've said that webAGC is a browser-only solution.  I don't really know the answer to that at the moment.  I think that if you were developing your own web app based on webAGC, it could be done without requiring a server at all.
> 
> I'm not sure if I understood this correctly, but I'd like to try to give an answer:

Your interpretation below is correct.  I don't like having to run ruby or light-server, or any HTTP server at all for something that runs entirely on the local machine.  But yes, static serving of files does answer my main question.
> 
> Browsers can read files even from hard disk and display static websites (html, js, css) just like, say, a word processor displays a word document. Strictly speaking, no HTTP servers are required in that scenario. However, when using more modern browser features, like JavaScript modules, WebAssembly, WebRTC etc., the browser insists that content be served over the network. This doesn't mean that servers need to have any 'business logic' implemented; they can just be static content delivery pipes. In this sense, the `webAGC` demo is indeed a pure browser/client solution; it is completely unaware of a server.

All right ... rather than rage against the machine, I'll take your word for it.

PR #1145, opened 2021-05-27
CMake: corrections, add more targets

scivision:
* skip yaDSKY2 if Wx not present
* handle time.h detection more correctly
* add yaLEMAP, yaOBC, yaAGS, Tools
* general project additions made since last CMake update
* add example target for Luminary099 using yaYUL

This was tested on GCC and Clang with MacOS, Windows, and Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,8 +17,10 @@ if(WIN32)
 endif()
 
 add_subdirectory(Tools)
+add_subdirectory(Luminary099)
 
 add_subdirectory(yaAGC)
+add_subdirectory(yaAGS)
 add_subdirectory(yaASM)
 add_subdirectory(yaDSKY2)
 add_subdirectory(yaLEMAP)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,14 @@
-cmake_minimum_required(VERSION 3.12)
-project(virtualAGC LANGUAGES C CXX)
+cmake_minimum_required(VERSION 3.14...3.21)
+project(virtualAGC
+LANGUAGES C CXX
+HOMEPAGE_URL "https://github.com/virtualagc/virtualagc")
 
-# this helps linters e.g. Visual Studio Intellicode work properly
 set(CMAKE_EXPORT_COMPILE_COMMANDS on)
 
-set(NVER \\\"2019-09-22\\\")
+set(NVER \\\"2020-12-24\\\")
+
+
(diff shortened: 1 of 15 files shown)


rburkey2005:
Looks good to me, though since I'm completely ignorant about cmake, probably anything would look good.  :smile:  And the changes to LVDC seem logical enough.  I'm merging this and closing.  Thanks!

rburkey2005:
@scivision So ... the Virtual AGC website's download page covers build instructions for a lot of platforms, but for some reason I never covered anything about cmake builds of Virtual AGC.  In fact, I find that I don't remember much about it, other than saying something like "go for it!" when asked if it was desirable.  Too old and too few brain cells, apparently.  :frowning_face: 

I think this situation should be corrected, to the extent that build instructions for cmake should be documented, and there should be a maintained list of what it builds and what it doesn't.

I thought we might try the following:  [I've created a wiki page for it here on the GitHub repo](https://github.com/virtualagc/virtualagc/wiki/Building-Virtual-AGC-Using-CMake).  Temporarily, at least, I'd link from the main website back to the wiki page for CMake instructions.

That only makes sense if you're able to edit the wiki page.  I _think_ you should be able to edit it, though it may be that I or another of the site maintainers would need to approve the changes.  Of course, since you're the subject expert, all changes would be approved.  But I admit I'm not sure if it works that way.  If you want to try it out, make some edits and see what happens.  Of course, even if does work, it might still prove to be quite a hassle.

And if it doesn't work ... well, we could figure out some other method to get the documentation out there efficiently.

rburkey2005:
It looks as though the wiki settings allow anyone to edit them, so I suspect it should at least work.

scivision:
Thanks--I filled in the CMake wiki. For many users, CMake may be an easier choice, and may also make cross compiling easier in the future. For now I didn't do any CMake cross compiling.

I was haphazardly adding targets that looked easy to add. Likely there is some interesting program I've omitted.

rburkey2005:
Super!  It didn't even ask me to approve it.  :smile:  Thanks!

I'll update the main website to link to the wiki page, probably tomorrow.  In general, while I like the main website to be the principal source of info, it's definitely a hassle that I'm basically the only one who can update it.  So at least for the near future, it's a big advantage that you can maintain the cmake wiki page yourself.

If there's anything more that's needed, such as specific versioning of cmake, etc., don't hesitate to add it.  You don't need to restrict yourself to just the topics I originally added to the page.

PR #1147, opened 2021-05-30
Issue 1143: Fix various symbol name and other minor typos

smithery1:
To recap, I compared at least the first entry in every column on every page of every end-of-listing table with the scans, and most changes came from there. Others came from various assembler cusses.
* Every change for a code base with scans matches the scans
* Every change for a code base without scans matches the scans from which the code originated, or as noted in a small number of cases where the change was to derived code

The comparison was not exhaustive, in particular I sampled most of the tables as described and did not validate every entry. But every change was applied across all applicable code bases.

See the associated Issue for more information and obviously feel free to ask any questions.


rburkey2005:
Looks good!  Great job!  Thanks!  All of which sounds like a pretty inadequate expression of appreciation, but it's the best I have in my word arsenal.

I'm merging it and writing it up.

PR #1148, opened 2021-07-26
Added --no-resume cli option

ThymoNL:
By default yaAGC looks for a file called `core` to reload the state of the erasable from a previous run. This is done to mimic the behavior of the non-volatile erasable memory used in the AGC and as a convenient option to continue running the program at a later time. However I don't always want to resume with a previous memory state and start completely fresh instead. My workflow used to consist of:
1. Run yaAGC
2. Make a change and recompile the rope
3. Remove the core file
4. Run yaAGC again.

There are occasions where I make small changes to a rope in quick succession and having to remove the core file gets old very fast. Or one might for example want to do a fresh run and compare that to a saved state later.

This commit adds a new convenience option to yaAGC, `--no-resume`, which allows you to skip the reloading of erasable and start with a completely fresh state instead, compressing steps 3 and 4 in my workflow. I've had this commit in a local copy of virtualagc for a couple of years now and it's proven to be useful to me, perhaps others might also find this option useful.

### DIFF
--- a/yaAGC/agc_cli.c
+++ b/yaAGC/agc_cli.c
@@ -102,6 +102,9 @@ static void CliShowUsage(void)
 "                         otherwise clean memory, in lieu of pad-loads.  Not\n"
 "                         required for subsequent runs.  Note that this option only\n"
 "                         has an effect if there is no existing core-dump file.\n"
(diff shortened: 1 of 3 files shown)


rburkey2005:
Makes sense to me.  Merging.  Thanks!

PR #1149, opened 2021-08-06
yaAGCb1: Fixed a few errors identified by SELF-CHECK

thewonderidiot:
These are fixes for yaAGCb1 to make it able to run Solarium 55's SELF-CHECK. Details are in #1150.

### DIFF
--- a/yaAGCb1/executeOneInstruction.c
+++ b/yaAGCb1/executeOneInstruction.c
@@ -42,6 +42,18 @@
  *                              instructions implemented now, though I've
  *                              had no way to test DV so far.
  *              2016-09-27 RSB  Hooked up DSRUPT.
+ *              2021-08-07 RSB  Fixed a bunch of errors identified by
+ *                              running SELF-CHECK in Solarium:
+ *                              * AD and INDEX now edit their arguments
+ *                              * SL no longer shifts into bit 14
+ *                              * DV now sets LP to either 140001 or 140000
+ *                                depending on inputs.
+ *                              * DV with negative numerators now works.
+ *                              * DV of equal-magnitude numbers now sets
+ *                                A and Q correctly.
+ *                              * CCS and SU now respect overflow.
+ *                              * TS A now preserves the value of A on
+ *                                overflow.
  */
 
 #include <stdlib.h>
--- a/yaAGCb1/executeOneInstruction.c
+++ b/yaAGCb1/executeOneInstruction.c
@@ -126,7 +126,7 @@ edit(uint16_t flatAddress)
     }
   else if (flatAddress == 023)
     {
-      regSL = ((regSL << 1) & 0077776) | ((regSL & 0100000) >> 15);
+      regSL = ((regSL << 1) & 0037776) | ((regSL & 0100000) >> 15);
     }
   return;
 }
@@ -498,14 +498,22 @@ executeOneInstruction(FILE *logFile)
               sign = -sign;
               denominator = ~denominator;
             }
-          quotient = numerator / denominator;
-          remainder = numerator % denominator;
-          if (quotient > 037777)
-            quotient = 037777;
-          if (sign < 0)
-            quotient = ~quotient;
-          regA = quotient;
-          regQ = ~remainder;
+          if ((numerator >> 14) == denominator)
+            {
+              regQ = ~denominator;
+              regA = (sign > 0) ? 037777 : 0140000;
+            }
+          else
+            {
+              quotient = numerator / denominator;
+              remainder = numerator % denominator;
+              if (quotient > 037777)
+                quotient = 037777;
+              if (sign < 0)
+                quotient = ~quotient;
+              regA = quotient;
+              regQ = ~remainder;
+            }
           if (sign > 0)
             regLP = 1;
           else
--- a/yaAGCb1/executeOneInstruction.c
+++ b/yaAGCb1/executeOneInstruction.c
@@ -349,6 +349,7 @@ executeOneInstruction(FILE *logFile)
           goto resumeFromInterrupt;
         }
       else agc.INDEX = fetchedFromOperand;
+      edit(operand);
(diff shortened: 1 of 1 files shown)


indy91:
Sadly this doesn't change anything for me. Aside from the DSKY locking up during the self test.

I have taken a different approach and tried to track down why Solarium is calculating an angle wrong.  Long story short, I think I have tracked it down to the function ARCTRIG, which uses the inputs SINTH and COSTH.

COSTH = 0.850901186466217
SINTH = -0.525536343455315

This should result in 328.3° being written in THETA, but actually results in 121.7°. How we get from this interpretive code to finding a bug in the emulator is something I don't know, but at least I am quite sure that this one calculation is where things go first wrong for me during the prelaunch procedures. One notable thing is, there are two other cases where ARCTRIG is used successfully, in both cases SINTH is a positive number. Might have something to do with it...

rburkey2005:
Running the self-check, eh?  What a novel idea!  Too bad I never thought of it.  As usual, thanks for cleaning up my messes.

It makes me wonder (vaguely) about whether John Pultorak's block 1 emulator (yaAGC-Block1) could run the self-check.  Well ... so vaguely that I may or may not actually try it out sometime.  I discarded yaAGC-Block1 and wrote yaAGCb1 instead, mainly because I'm too lazy to read other people's code, and hence didn't want to figure out how to integrate John's code into my own infrastructure, which I admit is not a very flattering picture of my own motivation.  :frowning_face:   But John mimicked control sequences in yaAGC-Block1, which I've never had the patience to do, preferring to instead in yaAGCb1 to follow higher-level behavioral descriptions of the instruction set.  I suspect the latter is faster, execution-wise.  But the former is more likely to be accurate, so John's code could conceivably actually work.  On the other hand, John never had any block 1 code to try out, and certainly never had Solarium's self-check routines, so it would be pretty amazing if it did work correctly out of the box.

Forgive the digression!  Integrating into NASSP is the important thing.

rburkey2005:
> I have taken a different approach and tried to track down why Solarium is calculating an angle wrong. Long story short, I think I have tracked it down to the function ARCTRIG, which uses the inputs SINTH and COSTH.
> 
> COSTH = 0.850901186466217 (octal 6635, 11244)
> SINTH = -0.525536343455315 (octal 73627, 54714)
> 
> This should result in 328.3° being written in THETA, but actually results in 121.7°.

Sorry to be dumb, but I can't help but notice that 121.7 times pi is 382.3.  Is that of any relevance?

indy91:
The relationship between the two numbers is actually 90° - 328.3° = 121.7° (wrapped around by 360°). I have taken a look at ARCTRIG, but I can't quite find the one spot that could be responsible for this error. The function uses signs, double subtraction etc. so it could be any of these operations, or more than one causing the problem.

rburkey2005:
One thing I used yaAGC-Block1 for was to run it side by side with yaAGCb1 and to compare the registers on an instruction by instruction basis. At this point, I've forgotten how to do that, so it's probably not worthwhile to try it just yet.  But if the solution doesn't present itself after a while, it might become worthwhile, since it might pinpoint exactly where the calculation goes astray.Sent on my Virgin Mobile Samsung Galaxy S® 5

From: Niklas Beug ***@***.***> 
Date: 8/6/21  7:24 AM  (GMT-06:00) 
To: virtualagc/virtualagc ***@***.***> 
Cc: Ronald Burkey ***@***.***>, Comment ***@***.***> 
Subject: Re: [virtualagc/virtualagc] yaAGCb1: Fixed a few errors identified by
  SELF-CHECK (#1149) 



The relationship between the two numbers is actually 90° - 328.3° = 121.7° (wrapped around by 360°). I have taken a look at ARCTRIG, but I can't quite find the one spot that could be responsible for this error. The function uses signs, double subtraction etc. so it could be any of these operations, or more than one causing the problem.



—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.

[

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/virtualagc/virtualagc/pull/1149#issuecomment-894223811",

"url": "https://github.com/virtualagc/virtualagc/pull/1149#issuecomment-894223811",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

indy91:
Latest change fixes the wrong angle calculated by P01. No change of the behavior in P05 where the AGC drives the IMU away from the correct alignment with gyro pulses. But this could be caused by feeding the PIPA data into the AGC in the wrong way, I just copied over some code from the Block II emulator for that. But progress at least!

thewonderidiot:
With these latest changes, yaAGCb1 now completely passes Solarium's SELF-CHECK. I've had it running for about half an hour now with no program alarms. :smile:

rburkey2005:
:+1:

rburkey2005:
@thewonderidiot I merged the changes but didn't delete the branch, since I didn't know if you were still making changes and it might mess you up somehow.

thewonderidiot:
Sounds good! There's no more fixes yet, but Niklas is going to try launching for the first time tomorrow, so who knows what the future holds. :smile:

indy91:
After some last minute problems (had the uplink switch still wired to channel 33, oops, caused the AGC to freeze) I managed to get to liftoff and then to orbit. Seems to work as expected, it went from P02 to P04 to P11 to P14. Not sure when and if it should switch to the next program, but I think so far it looked very good. No emulator issues anymore that I can notice. Before I can proceed further I need to do some more updates to the LVDC timeline and implement the mission control programer properly so that the rest of the mission will work.

Looks very good so far! I'll keep you updated.

PR #1151, opened 2021-08-09
Updated instructions about setting up the WebAssembly demo

michaelfranzl:
Because JSPM recently became deprecated, I had to change the instructions to run the webAGC demo. As a positive side effect, the instructions became simpler, the demo now also supports Firefox, and doesn't require bleeding edge browser versions.  I don't expect that the instructions will change again, but I can't guarantee it, so maybe it's better to decouple the VirtualAGC website from the instructions by just saying:

> You now have to serve the webAGC/demo directory via HTTP. Follow the instructions in the README file of the demo subdirectory.

This pull request changes the text `In the webAGC/demo/ directory, run [...] so please do not do it` to the sentence above.

As the entire section is on one line, I can't provide a more transparent diff.


rburkey2005:
@michaelfranzl Thanks!  Merging.

PR #1152, opened 2021-08-24
Reconstructed Luminary 096

thewonderidiot:
Here's a reconstructed Luminary 96... finally! It was made possible by the discovery of anomaly report LNY-59 in the MIT Museum's collection (picture below).

While updating the header comments, I realized that I had formerly incorrectly put Luminary 97 as the first Apollo 11 release, and Luminary 99 as the second, so those have both been updated to call them the second and third, respectively.

With this reconstruction complete, we finally have _every_ software build ever manufactured for Apollo 11 -- Comanche 51 and 55, and Luminary 96, 97, 99, and 99R1.

As an aside, LNY-59 concerns incorrect ephemeris constants, and was filed (and likely corrected) by none other than Jim Kernan. I talked to Don about it, and he and I both agree that this is almost certainly what Jim is remembering when he is thinking of the theoretical Luminary 99R2.

lny-59


rburkey2005:
That's fascinating.  The incident obviously made a big impression on Jim, since he kept repeating stories about it over the years, so you'd naively think that he'd have the details right.  However, I just finished reading a book on memory, and it says that this is exactly the kind of error you'd suspect his memory to have.  The book said that human memory is a read-writeback system, where every time you recall a memory, you then write it back ... but you've slightly edited it in the meantime.  So what you remember the next time is the edited memory.  Like a game of Telephone, where the details of the story change slightly every time it's passed on, until finally it doesn't resemble the original.  So a strong memory, like this one of Jim's, which he obviously remembered over and over, gradually morphs into something more interesting and dramatic.  Happened in Luminary 96?  Not so interesting.  Happened at the last minute in the final release?  Stuff of legend!

PR #1168, opened 2022-03-05
Donovan's First Pull

dgionis:
Pulling so that I may add early RTL files.


rburkey2005:
I apologize for the delay in commenting.  In reviewing the changes you've submitted, I wonder whether you really intended to push these changes upstream rather than to your own fork?

If so, you'll need to provide more background about what this is all about.  Most importantly, I literally have no idea what the changes are meant for.  Something to do with "mif" files, I gather, but only because I googled it, and what little I found doesn't immediately justify any software changes (in the virtualagc repository) without additional explanation.

Beyond that big-issue question, I have many issues at the detail level:

- Why has CMakeLists.txt been changed in the way it has?
- What is the entire new MAIN/ folder for?  
- And is "MAIN" really an appropriate name for it?  
- Other than Custom.py, MAIN/ seems to contain entirely temporary files, so why are those temporary files being committed to the repository?
- Why is MAIN/MAIN1.acg.html named "acg" instead of "agc"?
- Why are RAM.mif and ROM.mif in the top-level folder, rather than in MAIN or elsewhere?  I assume that whatever it is that they're for, they're probably specific to whichever specific version of the AGC software they're built against.  So presumably they should be stored in the folders specific to those AGC software versions ... Luminary099/, Comanche055/, and so on.  Even if it's appropriate for them to be in the top-level folder, the generic names they're given ("RAM" and "ROM") don't link them to any specific AGC software version, so they would be essentially useless anyway; nobody would ever know which software version they could be used with.
- In yaYUL, this business of adding 7 to `bank` I find very disquieting.  Is this something that's _necessary_ for implementing your mif files, or is it some unrelated idea that is supposed to make the existing code cleaner?  Or something else?

Those are the things that occur to me off the top of my head.

thewonderidiot:
I'm going to go ahead and close this to reduce notification noise, since we haven't heard anything back from them and I think it's pretty clearly not something we're going to want to merge in. @dgionis you're welcome to reopen this at some point in the future, if you can answer all of Ron's questions and clean up all of the unnecessary additions and breaking changes.

PR #1171, opened 2022-05-21
Add missing symbol BIT7 to Luminary099 FLAGWORD_ASSIGNMENTS

navistartronic:
This modification to the source code is to match that of the scanned source listings as found at http://www.ibiblio.org/apollo/ScansForConversion/Luminary099/0076.jpg

Source code change in the pull request is in Luminary099 FLAGWORD_ASSIGNMENTS.agc, Page 76:

Added missing definition of symbol BIT7 (commented out to match that of the others).


rburkey2005:
Thanks!  I'm merging it.

rburkey2005:
Note to self (and anyone else interested):  I checked a few other versions (Apollo 10, 12, and 13) of the LM source code, and it doesn't look like there's anything to fix in them.  Curious!

PR #1178, opened 2022-08-24
Add metric system units to readme

Piturnah:
Include the weight and dimensions of the AGC an DSKY in kilograms and cm (weight to 3.s.f. as given by imperial measurements) for people more accustomed to metric system

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Other branches of the repository often contain very different files. Here are so
 * 36,864 words of read-only memory, equivalent to 69,120 bytes.
 * Maximum of about 85,000 CPU instructions executed per second.
 * Dimensions: 24"×12.5"×6".
-* Weight: 70.1 pounds.
+* Weight: 70.1 pounds (31.8kg).
 * Power supply: 2.5A of current at 28V DC
 * Real DSKY.
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Other branches of the repository often contain very different files. Here are so
 * 2048 words of RAM. A "word" was 15 bits of data—therefore just under 2 bytes (16 bits) of data—and so the total RAM was just 3840 bytes.
 * 36,864 words of read-only memory, equivalent to 69,120 bytes.
 * Maximum of about 85,000 CPU instructions executed per second.
-* Dimensions: 24"×12.5"×6".
+* Dimensions: 24"×12.5"×6" (60cm×30cm×15cm).
 * Weight: 70.1 pounds (31.8kg).
 * Power supply: 2.5A of current at 28V DC
 * Real DSKY.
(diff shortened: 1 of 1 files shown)


rburkey2005:
Sounds like a good idea.  Unfortunately, some storms and power outages here have left me momentarily without a completely-working computer, so I may not be able to look at this for a couple of days.

rburkey2005:
@Piturnah You've opened up a can of worms here.  :smile:  

I felt compelled to double-check your numbers, and if they had represented a straightforward conversion of inches to cm, I would probably have accepted them without further thought.  But the metric figures you give for the AGC are lower-precision than those you give for the DSKY, and the ones for the DSKY vary curiously (with 8" being presented both as 21 cm and as 20.3 cm).  I presume you must be quoting at least some of these numbers from other sources, rather than having made the conversions yourself.

All of this started making me wonder where my original figures in inches and pounds had come from anyway, and whether they were even correct at all?  Yes, they agree with what wikipedia says ... but so what?  And besides, for all I know, the wikipedia article may be quoting the Virtual AGC website, so if I start relying on wikipedia it may be a vicious circle of cross-referencing.

But once I started looking into the actual dimensions I immediately saw further complications, such as the fact that there were multiple models of the AGC and DSKY used in different missions, and they may not necessarily have all been absolutely identical in physical dimensions.  (None of which I knew when I originally wrote those dimensions into the web page, back in 2003.)

So I think there are a couple of possible approaches to fixing this, neither of which is quite solved by your patch.  We could:

1. State explicitly that the dimensions and weight are approximate.  Round them in imperial units to the nearest half inch and pound.  Round them in metric units to the nearest cm and half kg; or else
2. State explicitly that they _may_ vary slightly by mission.  Give the imperial measurements to the full precision to which they are known.  Link specifically to the documents from which these numbers are known.  Give the metric units to a precision compatible with the imperial units from which they are converted.  (For example, if the AGC length is given to 3 significant figures in pounds, then it should be 3 significant figures in cm as well.)

Regarding the accuracy of the original imperial figures, I find the following:

- For the Apollo 11 Lunar Module, physical dimensions of the AGC come from engineering drawing 2003993-031 (rev T), of which we have several scans of greater or lesser legibility, but [the measurements are reasonably legible at this hyperlink](https://archive.org/details/AgcApertureCardsBatch2Images/page/n281/mode/1up?view=theater):  24.250"&times;12.433"&times;5.974".
- For the Apollo 11 Lunar Module, physical dimensions of the DSKY come from [engineering drawing 2003956 (rev B)](https://archive.org/details/apertureCardBox459NARASW_images/apertureCardBox459NARASW/page/n566/mode/1up?view=theater): 8.124"&times;8.000&times;6.91".
- As for weights, I don't think we know the actual weights for any specific mission.  The latest reported weights I'm aware of are from [the "Weight & Balance Report" of April 1, 1968](https://www.ibiblio.org/apollo/Documents/E1142-59.pdf#page=8):  AGC 70.1 pounds, DSKY 17.8 pounds.

So ... I don't intend to accept your patch as-is, but instead think that it should be changed to reflect one the two approaches mentioned above.  Since I bestirred myself to make the effort to research it, I favor the 2nd approach at the moment.

Would you prefer to cancel your patch and resubmit it along those lines, or would you like me to just make the changes myself?  Or perhaps you have an alternate suggestion?

rburkey2005:
And since inches and cm are roughly the same order of magnitude, as are pounds and inches, by "significant figures" I mean "number of digits to the right of the decimal point".  A bit of precision is lost in the metric units that way, but that's find since they're conversions rather than measurements or design elements.

rburkey2005:
It occurs to me that that's a pretty pedantic response to your fairly-innocuous request.  I'm accepting the patch, but intend to apply my own correction to it afterward.  Thanks!

PR #1181, opened 2022-09-24
Simplify WebAssembly build instructions

michaelfranzl:
Last year, when these instructions were written, it was necessary to either compile wasi-sdk from a specific git commit, or wait until wasi-sdk version 13 was released.

Meanwhile, wasi-sdk is at version 16, so the version check and the compilation from source are no longer necessary. It is always possible to use the pre-built wasi-sdk release files.

This allows to simplify the WebAssembly build instructions.

The upgrade of wasi-sdk to version 16 makes it necessary to change the VirtualAGC WebAssembly compiler arguments, for which I will create a separate pull request to the master branch.


michaelfranzl:
I had an outdated base branch, I'm going to create a new pull request.

PR #1187, opened 2023-02-03
Schematics: typo in the part number for 2004968-001 COVER, FRONT

rrainey:
This looks to be a scan error from the source aperture card.  The "8" looks like a "3" on the scan.  Inserted the proper id for COVER, FRONT in this assembly.  Cheers.


rburkey2005:
Thanks!  Unfortunately, the change also has to be incorporated into the HTML files of the drawing trees, which isn't exactly automatic since I have to manually run the software to do it ... I think.  And it has been so long now since I've done it last that I don't remember how and will have to research it.  So it may be a while before the HTML files actually reflect the change.

rburkey2005:
Having accepted your patch and merged this commit, I've been looking at it more closely, and I don't think you're (fully) correct.  According to your correction, 2004968 "Housing, Front" is a component of 2004968 "Housing, Front".

Don't you think it's more likely that 2004963 is correct, but that it's the description rather than the part number that's a misprint on the schematic?  Namely, that it should have been 2004963-001 "Plug, push button switch, AGC DSKY"?

rburkey2005:
Hmm ... no, I guess not.  I guess you were right.

thewonderidiot:
Just chiming in to say, yeah, it's very common for inseparable assembly drawings to reference themselves in the component list, alongside the various other parts like helical inserts. The [fixed memory module housing 2004100](https://archive.org/details/apertureCardBox459NARASW_images/apertureCardBox459NARASW/page/n1155/mode/1up?view=theater) even takes it a step further and separately specifies two different pieces, 2004100-001 and 2004100-002, which are brazed together to form the full 2004100.

PR #1188, opened 2023-02-05
Schematics: typo in later variant of FRONT HOUSING CSV

rrainey:
I came across this somewhat randomly; this is not part of a methodical check of CSVs. 

While I'm here it might be worth noting that the drawing for the correct assembly, 2003984, is missing from the aperture cards. For what it is worth, 2003875 appears to be an earlier variant of the pushbutton design - a CSV file is already present for 2003875 - perhaps that's worth noting in this CSV. I was unsure how best to do that.


rburkey2005:
@rrainey This change doesn't appear to be correct to me.  The FIND-table in drawing 2003949G says:

- 2003894 is used in 2003949-031
- 2003875 is used in 2003949-021
- 2003984 is used in 2003949-011

And that's what the CSV file currently says as well.   I think.

Or are you asserting that the original drawing is in error?

As far as part substitutions are concerned, the software that uses these CSV files to autogenerate the HTML assembly drilldown pages is sometimes smart enough to do that on its own.  For example, at this moment I'm looking at parent assembly 2003994-091, which calls out 2003949-041; but since 2003949-041 doesn't exist, which is noted by different coloring on the HTML drilldown page, the suggestion to use 2003949-031 has been added automatically by the software.

However, in the case of 2003894-211 or 2003984-211 (say) vs 2003875-211, the software isn't smart enough to make that suggested substitution.

And there's presently no provision in the software to make such a suggestion manually, I think.  (Though it's an interesting idea of something to implement _if_ someone were making a comprehensive review vs merely a one-off fix.)   The closest thing to it is to replace every entry in the DRAWING column like

> `2003894-211`

with

> `2003894-211 or 2003875-211`

The problem with this is that these two alternatives would be portrayed in the drilldowns as absolutely equivalent official drop-in replacements, and not as a last-ditch attempts to find something that can be cobbled in as a workaround.  So I'd prefer not to do it.

rrainey:
OK, well, I see my error now. I hadn't caught the three different variants of switches and locked-in on the transposed digits for 2003894 vs. 2003984. Disregard. 

Thanks, you are correct.

PR #1189, opened 2023-02-22
Reconstructed Sunrise 45 Source

thewonderidiot:
It's been a lot of work and a long time coming, but I've finally finished a first-pass disassembly of Sunrise 45, with lots of help from @indy91. Everything is named (some labels accurately, some certainly not), and it assembles to exactly match the dumps. A good enough starting point, at least.

As always, feel free to change up the headers and Main.annotation to your heart's content, if you don't like my wording.


rburkey2005:
@thewonderidiot In starting the process of writing this up for the website, I notice that in the previous writeup of just the module dumps, I didn't give any instructions for how to actually run any of the tests other than the electroluminescent test.  (Actually, I notice that the writeups for Sundial and Aurora don't have any such instructions either.)  Curious omission, given that the ability to perform tests is the what we're touting these things for.

If you get some spare time in which you're bored &mdash; assuming that's something that happens to you &mdash; perhaps you'd be willing to throw together some getting-started instructions that I could write into the descriptions of these three programs.

rburkey2005:
@thewonderidiot I did make a couple of minor changes &mdash; an extra &lt;br&gt; in the annotations, and rewrite of the program comments in MAIN.html to relate to Sunrise 45 rather than Solarium 55.  Not much, but maybe worth a pull from the repository.

thewonderidiot:
Okay, will pull! And I'll throw together some basic instructions for the tests I've figured out in Sunrise tonight.

There's one more file that needs to be updated here -- the README.md in the Sunrise45 folder. I didn't touch it from what you had before, and it's now at least partly outdated (I changed the name of the dump file from `Sunrise45.bin` to `Sunrise45-dump.bin`, since yaYUL wants to produce an output binary with the former name).

rburkey2005:
I hadn't noticed the README, but it's fixed up now, hopefully well enough.

thewonderidiot:
Okay, here's how to do various tests in Sunrise 45:

### AGC Self-Check

Sunrise 45 has a very early version of self-check that doesn't include erasable tests, fixed memory checks, etc. It essentially only has the control pulse tests. To kick it off:

- VERB 21 NOUN 27 ENTER XXXXX ENTER

where XXXXX is the number of times you want it to execute. If XXXXX is negative, it will instead loop forever until a new N27 is entered or an error is encountered. The high-level results can be monitored with:

- VERB 15 NOUN 01 ENTER 1570 ENTER

The first register will contain the address (within bank 5) of the first failure encountered. The next two registers are a double-precision count of successful test completions. When all is working, this will count up very quickly.

### Orbital Integration

The orbital integration section of Sunrise is essentially a test program that propagates a very low-altitude Earth orbit for just a bit more than one orbit. It can be started with verb 37 as follows:

- VERB 37 ENTER 11 ENTER
- KEY RLSE

The program will change to 02, and it will put up a V07N01 display of address 1315. In this display, registers 1 and 2 are a double-precision fractional simulation time in weeks. It will slowly count up until it reaches 0.009 weeks, i.e. 90 minutes, at which point the program will exit. While it's running, the double-precision vectors **RDISP** (position in kilometers scaled 2^-14, address 1431) and **VDISP** (velocity in meters per second, scaled sqrt(mu) * 2^6 * 10^-3, address 1437) can be examined one element at a time with a manual V07N01. I don't think there's a good way to display the whole vector. 

### Other Tests

The other tests available in Sunrise are Prelaunch Alignment, Gryo Drift Test, and PIPA Scale Factor Test. All three share two common traits: they're not terribly useful/interesting in a standalone emulator without a simulated inertial subsystem, and they're all very well-documented in the AGC Information Series, unlike self-test and orbital integration.

**Prelaunch Alignment**

- Documented in [AGCIS Issue 22, Prelaunch Alignment](https://www.ibiblio.org/apollo/Documents/agcis_22_prelaunch_alignment.pdf)
- Started with VERB 37 ENTER 01 ENTER

**Gyro Drift Test**

- Documented in [AGCIS Issue 21, System Test](https://www.ibiblio.org/apollo/Documents/agcis_21_system_test.pdf)
- Started with VERB 70 ENTER

**PIPA Scale Factor Test**

- Documented in [AGCIS Issue 21, System Test](https://www.ibiblio.org/apollo/Documents/agcis_21_system_test.pdf)
- Started with VERB 71 ENTER

rburkey2005:
@thewonderidiot Sweet!  I basically just cut-and-pasted it.  Thanks!

PR #1194, opened 2023-06-19
Solarium 55: Corrected names of a couple of log sections

thewonderidiot:
Tentative PR to fix the Solarium log section names, pending discovery of the reason they were changed to begin with.
image
image


rburkey2005:
@thewonderidiot It doesn't appear to me that the filenames were ever changed.  They disagreed with the log-section names from day 1.  What the comment for the 2016 commit meant is that the list of filenames in MAIN.agc didn't agree with the actual filenames, so the assembly failed, and MAIN.agc was changed to reflect the actual filenames so that the assembly could succeed.

As for why the change in 2016 was made to MAIN.agc rather than the filenames themselves, I'm sure there's no deep reason.  It was probably just less work, and the filenames were (after all) arbitrary.  

At any rate, changing them as you suggest is fine.

PR #1195, opened 2023-06-19
Add initial Corona 261 source reconstruction

thewonderidiot:
Here's the first pass at Corona 261 source. Lots of work is left to be done in reverse engineering and naming the Midcourse Navigation Game, but I'm probably going to table it for now and move on to getting initial Sunrise 69 and Sundial E source trees built up and submitted.

As part of this, I deleted the README, since it only talked about how the dump was tentative and subject to changes, and I am now significantly more confident in the accuracy of the dump (and there will be no further changes unless something dramatic happens).


rburkey2005:
:+1:

rburkey2005:
Since you're going to follow-on with these rather quickly, I won't write anything up about it until Sunrise 69 and Sundial E are merged.

PR #1198, opened 2023-07-01
Add source reconstruction of Sundial E

thewonderidiot:
Here's a preliminary source reconstruction of Sundial E! I didn't touch the README this time around; I'll leave that up to you if you want to change anything. 

I'm planning on following this up with a Sundial D reconstruction, because despite the lack of bugger words I'm quite confident that I know what the change was. For this Sundial E source, I have annotated said change at the bottom of the PRELAUNCH ALIGNMENT PROGRAM section. It's easy to back out -- I just need to test it in NASSP.

After that, I'll move on to Aurora 88.


rburkey2005:
@thewonderidiot Silly question.  In writing this up, I thought I'd give an example or two of the more-speculative portions of the reconstruction, and how to recognize them.  I couldn't find any.  Then I went back to Sunrise 69 (or at least the modules that were new relative to Sunrise 45) and couldn't find any there either.  My expectation, among other things, was to at least find symbolic labels that were obviously artificial, but I didn't see any.  Well, it wouldn't be the first time I've overlooked the obvious.  Hints please?

thewonderidiot:
I'd like to think that means I've done my job pretty well. :slightly_smiling_face: 

For Sundial, the most speculative parts are in the CSM & Saturn Tests, the Controls and Displays test, and (worst of all) the gyrocompassing.

- The [Saturn steering test](https://github.com/virtualagc/virtualagc/blob/master/SundialE/CSM_AND_SATURN_INTEGRATED_TESTS.agc#L528) is entirely my labels -- although it shares a lot of structure with the other tests that came more or less directly from Aurora's LEM FCS tests.
- The [Controls & Displays test](https://github.com/virtualagc/virtualagc/blob/master/SundialE/IMU_PERFORMANCE_TESTS_3.agc#L471) is also all made-up.
- This [big chunk in the middle of optimum gyrocompassing](https://github.com/virtualagc/virtualagc/blob/master/SundialE/OPTIMUM_PRELAUNCH_ALIGNMENT_CALIBRATION.agc#L784) (down to the next comment) is the weakest, and probably has the most room for improvement.

Sunrise doesn't really have many dubious labels in it, due to the sheer strength of the AGC Information Series documentation. The worst parts are:

- [Orbital integration](https://github.com/virtualagc/virtualagc/blob/master/Sunrise69/ORBITAL_INTEGRATION_PROGRAM.agc#L39), which is essentially the only part of Sunrise 45 that is not thoroughly documented.
- [The V20 leadins for the IMU performance tests](https://github.com/virtualagc/virtualagc/blob/master/Sunrise69/IMU_PERFORMANCE_TESTS_1.agc#L1039), which obviously aren't present in any other ropes because most ropes have dedicated verbs for this stuff.

Corona's flaws are the most obvious, in MIDCOURSE NAVIGATION GAME and B VECTOR ROUTINE -- it still retains many of my placeholder labels where I haven't even been able to slightly figure out what's going on yet. Sundial E does have just a couple of those [here](https://github.com/virtualagc/virtualagc/blob/master/SundialE/IMU_PERFORMANCE_TESTS_2.agc#L1101), for two functions that are totally unreferenced and so broken that I can't even really tell what they were trying to accomplish.

rburkey2005:
> I'd like to think that means I've done my job pretty well.

Well, yes, it's terrific work.  That's not even a question. 

The problem is that it leaves us with reconstructions in which nobody but you can tell where they may be questionable.  And in particular, nobody else can possibly improve them by having brainstorms or access to info you didn't have.  Because to everyone else, they're just solid blocks of apparent goodness.  Then too, these invented portions in Sundial E, Sunrise, Corona, and so on, are now going to propagate unquestioned into Sundial D, and at least potentially, any other future reconstructions.  And on a silly note, these picture-perfect reconstructions represent exactly what Apollo deniers might accuse us of (if they had any cognizance of Virtual AGC in the first place):  of making stuff up that can fool people into thinking it was originally created back in the 1960's.

In a word, I wonder if we haven't sacrificed too much traceability to achieve this improved readability.

I suppose it's far too late to make this observation, and I suppose it's my fault for not having paid enough attention before now, but I think the phased approach you used in Sundance 306 was better:  A Phase 1 reconstruction in which the uncertain portions (like placeholder labels) were used and the uncertain material was very clearly demarcated from the cut-and-pasted material, vs a Phase 2 reconstruction such as what you've posted now in which everything flows together wonderfully.

thewonderidiot:
Sundance's Phase 1 reconstruction was different -- I don't think it's a good comparison. I did the same thing for its phase 1 that that I did for these; it contains no placeholder labels and all of the "new" Sundance-only code was given names (e.g. see [P10 and P11](https://github.com/virtualagc/virtualagc/blob/master/SundanceXXX/P10%2CP11.agc) which is entirely Sundance-only-code with names entirely made up by me and Niklas). Sundance's Phase 2 reconstruction contains actual new code, written by my and Niklas, to bump up our older Sundance modules to be more 306-like.

Unlike Sundance 306ish, these source reconstructions _do_ assemble to be bit-for-bit identical to the dumps.

rburkey2005:
That's not what I meant about virtues of the Sundance reconstruction.  I meant that the "phases" proceeded from the more-certain to the more-invented, with that latter being traceable to the former.Sent on my Samsung Galaxy A51
-------- Original message --------From: Mike Stewart ***@***.***> Date: 7/1/23  12:19 PM  (GMT-06:00) To: virtualagc/virtualagc ***@***.***> Cc: Ronald Burkey ***@***.***>, State change ***@***.***> Subject: Re: [virtualagc/virtualagc] Add source reconstruction of Sundial E
  (PR #1198) 
Sundance's Phase 1 reconstruction was different -- I don't think it's a good comparison. I did the same thing for its phase 1 that that I did for these; it contains no placeholder labels and all of the "new" Sundance-only code was given names (e.g. see P10 and P11 which is entirely Sundance-only-code with names entirely made up by me and Niklas). Sundance's Phase 2 reconstruction contains actual new code, written by my and Niklas, to bump up our older Sundance modules to be more 306-like.
Unlike Sundance 306ish, these source reconstructions do assemble to be bit-for-bit identical to the dumps.

thewonderidiot:
Right, but my point is that the phase 1 is already guilty of what you're accusing the source reconstructions of here. So it would need at least three our four phases, instead of only two. Can you give examples of what you have in mind for the source reconstructions, specifically?

rburkey2005:
Just to be clear, I'm not asking for anything different on any of the reconstructions you've posted or are working on now, because I'm sure it's too late.  And I'm just now thinking about it for the first time, so I can't point to some perfect sample of one that was done "right".  They're _all_ "right", according to the criteria you were using.  So anything I can say on the matter has to be along "in a perfect world" lines of thought.

But think about it in terms of the reconstructions you did &mdash; yes, I know it's a completely different situation! &mdash; for Comanche 44, 45, and MANCHE45.  Everything in these came either from a cut-and-paste from a different but known-and-documented existing version, or else was made for some specific reason that (from your notes) I was able to document with an annotation.  The result is that at least in principle, you can go to any line of that source code and know immediately _where_ that code came from and _why_.   If something such as a program label was invented, you know that it was invented, and you know why it was invented, and you can use your own judgement to determine whether or not the invention was trustworthy or not.  Mainly, I'd presume, the mere fact that it's documented would lead you to believe that it was indeed trustworthy.

So in terms of the present question, it seems to me that in the reconstruction you have two different kinds of code:  That which was simply cut-and-pasted, unchanged, and that which has undergone complete or partial reimagining.  As a starting point, I'd say that cut-and-pasted code doesn't need justification or annotation and can just be used as-is.  It's anything that _didn't_ come directly (or wasn't derived in some obvious way) from cut-and-pasted code that needs to be made obvious in Phase 1.  That would seem to result in two principles that need to be applied in Phase 1:

1. Placeholder labels should look like placeholder labels, presumably because they're constructed according to some stated or obvious formula; and
2. Blocks of code derived from disassembly (vs cut-and-paste) should be explicitly annotated as having come from disassembly.

As far as an example is concerned, I think that [the example in Sundial E you pointed out](https://github.com/virtualagc/virtualagc/blob/master/SundialE/IMU_PERFORMANCE_TESTS_2.agc#L1101) is really close to what I'd want. The only change I'd make to it is to enhance the leading comment to something like:
<pre>
## Reconstruction:  The following two functions are derived from disassembly
## of the core-rope rather than taken from related programs.
## MAS 2023: They are unreferenced, and appear to be broken. It is unclear 
## what their original purpose was.
</pre>

thewonderidiot:
Okay -- I'll keep doing what I'm doing for Aurora. The reason why I opted to not add such comments throughout is that ropes like Sunrise, Corona, and Sundial, for which we have no real close analogs, would have hundreds or thousands of such comments throughout -- in many places to the point of making the code unreadable, in my opinion.

rburkey2005:
At first I had read that as "hundreds _of_ thousands" and was considering how to tone down a saucy comeback.   :smile: 

But yes, the readability of the Phase 1 code _would_ be much worse than the Phase 2 code, which would have none of the extra annotations and would be beautiful.

If the data were still available, as I imagine it's not, a dictionary of phase-1 placeholder labels vs the phase-2 labels and an inventory of the code blocks (giving just starting line number and block length in lines) derived solely by disassembly would be just as useful.  And since your placeholder labels are of the form Uxx,xxxx, the dictionary of placeholder labels could be a simple list of them.

If we had those and the Phase 2 reconstructions, a script could be written that regenerated Phase 1 reconstructions, rendering retention of an explicit phase-2 code base unnecessary.

thewonderidiot:
The data isn't really available, and in many cases never existed in the first place. I only use the Uxx,xxxx labels when I get really stumped. I think the easiest way to generate what you're going for would be to flesh out your disassemblerAGC tool and have it spit out source files.

rburkey2005:
I had forgotten its very existence.

PR #1199, opened 2023-07-03
Sundial E: disentangle illegally-placed patches

thewonderidiot:
While working on the Aurora 88 disassembly, I ran into what I briefly thought might have been a bug in yaYUL. In attempting to keep patches co-located with the code being patched, I created an inadvertent circular SETLOC chain, caused by the `SETLOC` for a patch in bank 13 preceding the first `BANK 13`  card in the listing. yaYUL gave up after 10 passes and produced no output.

I thought for a moment that perhaps the original Yul might be able to handle this situation, so I whipped up a punch card deck and threw it into pyul. It didn't even get as far as yaYUL did before barfing. The cuss it gave me, however, made me question some of what I had done in Sundial E.

So I put together a simple test program and fed it into both yaYUL and pyul:
```
            BANK    10
ZERO        OCT     0
END1        EQUALS

            SETLOC  END2
ONE         OCT     1

            SETLOC  END1
TWO         OCT     2
END2        EQUALS
```

It's a minimal reproducing case for how I placed some of the patches in Sundial E.  This should in theory allocate the constants in the order ZERO, TWO, ONE. yaYUL can assemble this chunk without any errors, and gets the order correct. No surprise there. pyul, however, cusses about it:
```
 0001                                  10,2000           BANK   10
 0002                        10,2000   00000 1  ZERO     OCT    0

 0003                                  10,2001  END1     EQUALS
 0004   REF   1                           ■■■■           SETLOC END2
E■■■■■■ # 1     "END2    " WAS UNDEFINED IN PASS 1
 0005                        10,2001   00001 0  ONE      OCT    1

 0006   REF   1                        10,2001           SETLOC END1
 0007                        10,2001   00002 0  TWO      OCT    2
E■■■■■■ # 2     CONFLICT IN USE OF THIS LOCATION
E■■■■■■ # 2     "TWO     " ASSOCIATED WITH CONFLICT
 0008                                  10,2002  END2     EQUALS
```

In other words, yaYUL is actually being a bit more powerful here than the original Yul, and let me do some things that should have been illegal. This MR unwinds that illegal SETLOCing, at the cost of patches moving to files that have nothing to do with the code being patched. We've seen that in some of the listings we have (e.g. Comanche places patches for the gravity model in RTB OPCODES), so this being a restriction is totally believable. I also fixed a few corrupted comments and corrected a bit of whitespace.

### DIFF
--- a/SundialE/AGC_BLOCK_TWO_SELF-CHECK.agc
+++ b/SundialE/AGC_BLOCK_TWO_SELF-CHECK.agc
@@ -12,6 +12,7 @@
 ## Website:     www.ibiblio.org/apollo/index.html
 ## Mod history: 2023-06-22 MAS  Created from Aurora 12.
 ##              2023-06-30 MAS  Updated for Sundial E.
+##              2023-07-03 MAS  Added note for patches introduced in Sundial D.
 
 
                 BANK    10
@@ -1218,7 +1219,9 @@ DV5--           EXTEND
                 INCR    SCOUNT +2
                 TC      SELFCHK         # START SELF-CHECK AGAIN
 
-
+## MAS 2023: The following two chunks of code were added as patches
+## in Sundial D. They were placed here at the end of the bank so
+## as to not change addresses of existing symbols.
 
 ERASLP1         EXTEND
                 INDEX   SKEEP7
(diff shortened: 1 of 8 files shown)


rburkey2005:
> yaYUL is actually being a bit more powerful here than the original Yul

I like the concept, but I suspect it's more like "a bit stupider".  Since I didn't know how to do it efficiently, I just threw lots and lots of CPU cycles at it, and it happened to work out okay.  :smile:

PR #1201, opened 2023-07-13
Add source reconstruction of Aurora 88

thewonderidiot:
Some particularly notable things about Aurora 88:

- It has a _lot_ of code for the LEM Optical Rendezvous System (LORS). We have no listings containing this code, and for a long time it only existed in offline assemblies like BATMAN. It appears to be fully functional -- you can put Aurora into LORS mode with VERB 66 ENTER. After that, you can exercise the three LORS verbs 70, 71, and 72. None of them do much without a LOTS simulation hooked up, but verb 71 (LOTS self-test) is the most interesting. Its operation is fully described in [XDE34-T-53, Preinstallation G&N Testing of the Lunar Excursion Module Optical Rendezvous System](https://www.ibiblio.org/apollo/Documents/xde-34-t-53.pdf) starting on PDF page 68. I left some big comments with useful relevant links, including this document, at the top of the tentatively-named LOTS_INTERFACE.agc and LOTS_TRACKING.agc files.
- Aurora 88 is _by far_ the patchiest AGC code we have. Nine banks have end-of-bank patches. A few of these have 6+ patches. In a couple of cases, they ran out of room in their end-of-bank patches and had to BANKCALL or POSTJUMP to the end of a _different_ bank so they could add more patch code. One of the most interesting "patches", however, is in IMU PERFORMANCE TESTS 2. All of the drift/PIPA positioning functions, POSN1 through POSN10, were "shortened" between Aurora 85 and 88. This is only true in terms of instructions executed; physically, due to cross-module references, the functions are all still the same length. The shorter versions of the functions were simply placed on top of the older, longer ones, leaving the tail ends of all of the old functions as dead code. In some places (presumably where this dead code would have caused assembly errors) random instructions were replaced by NOOPs. And in other places, where they would fit, patches that would have otherwise gone at the end of the bank were placed over top of the dead code. It's all quite messy and caused me a great deal of confusion while disassembling.


rburkey2005:
Excellent as usual, thanks!

PR #1205, opened 2023-07-16
yaLVDC: Assorted fixes for CDS, MPY/MPH, and debugging

thewonderidiot:
- The debug interface wasn't correctly accepting `SI` and `NI`
- CDS shouldn't require A8 to be 0:
image
- MPH was not automatically loading PQ into the accumulator. This is mechanized in the LVDC by forcing the instruction `CLA 775` to automatically occur 4 times after the `MPH` instruction. The last one grabs the final product.
image
- MPH and MPY were producing very bad results at first, and then mostly-correct-but-sometimes-off-by-one results once the gross error was fixed. I couldn't for the life of me find a reliable way to determine if the result was going to be off by one or not. But I did find that a naive implementation of the algorithm described in paragraph 2-319 of the laboratory maintenance instructions lead to results that match the simulator in every case I have thrown at it so far. It's less efficient than native multiplication, obviously, but it's only 6 iterations and the LVDC is a slow computer.
image
image

### DIFF
--- a/yaLVDC/gdbInterface.c
+++ b/yaLVDC/gdbInterface.c
@@ -36,6 +36,7 @@
  * Reference:   http://www.ibibio.org/apollo
  * Mods:        2020-04-30 RSB  Began.
  *              2020-05-21 RSB  Fixed arguments for DISASSEMBLE.
+ *              2023-07-16 MAS  Fixed handling of SI and NI.
  */
 
 #include <stdlib.h>
@@ -176,9 +177,9 @@ typedef struct
 commandAssociation_t commandAssociations[] =
       {
         { ctSTEP, "STEPI", "STEPI [n]", "Step n instructions, default n=1." },
-        { ctSTEP, "STEPI", "SI [n]", "Same as STEPI." },
+        { ctSTEP, "SI", "SI [n]", "Same as STEPI." },
         { ctNEXT, "NEXTI", "NEXTI [n]", "Next n instructions, w/o entry." },
-        { ctNEXT, "NEXTI", "NI [n]", "Same as NEXTI." },
+        { ctNEXT, "NI", "NI [n]", "Same as NEXTI." },
         { ctDELETE, "DELETE", "DELETE", "Delete all breakpoints." },
         { ctDELETE, "DELETE", "DELETE n", "Delete breakpoint n." },
         { ctCONTINUE, "CONTINUE", "CONTINUE", "Continue running emulation." },
--- a/yaLVDC/runOneInstruction.c
+++ b/yaLVDC/runOneInstruction.c
@@ -43,6 +43,13 @@
  *                                  ACC.
  *                              3)  Assigned "spare" CIO 175 as an input port
  *                                  rather than an output port.
+ *              2023-07-16 MAS  Fixes for various bugs affecting only the LVDC:
+ *                              1)  CDS should not require A8 to be 0.
+ *                              2)  MPH was not loading the accumulator with PQ.
+ *                              3)  MPH/MPY were producing incorrect results;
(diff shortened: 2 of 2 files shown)


rburkey2005:
> ... then mostly-correct-but-sometimes-off-by-one results once the gross error was fixed. I couldn't for the life of me find a reliable way to determine if the result was going to be off by one or not. But I did find that a naive implementation of the algorithm described in paragraph 2-319 of the laboratory maintenance instructions lead to results that match the simulator in every case I have thrown at it so far.

Good grief!

rburkey2005:
@thewonderidiot In the LVDC assembler, there are 6 remaining off-by-1 mismatches.  Two of those are actually identical errors in AS-512 and AS-513, so there are really only 5 _distinct_ mismatches, and 4 of those 5 mismatches are captured in the following code snippet:
<pre>
P.CSR  EQU     (4063.492)
       SSF1    (3031.4*P.CSR#000)B27
       SSF1    (4031.4*P.CSR#000)B27
       SSF1    (31320.0*P.CSR#000)B27
       SSF1    (31320.2*P.CSR#000)B27
</pre>
So I wonder if the LVDC assembler mightn't be mimicking LVDC multiplication?  Do you have any easy way to check those multiplications out?  (I mean, easier than me implementing a new multiplication algorithm in the assembler?)

The final error is 
<pre>
P.P1   EQU     (1/180)
       MPH     =(P.P1)B-7
</pre>
which obviously is a somewhat different problem.  But I can only suppose that if `MPY` is boogered up in the emulator, then `DIV` is likely to be boogered up in the emulator worse.  And if the assembler is mimicking the CPU (emulator) for multiplications, it's likely to be doing so for divisions as well.

thewonderidiot:
Hmmm -- I have an easy way to test multiplication, but it's not immediately obvious to me what the best way to convert the floating point inputs to integers for the computation. We'd have to pick some scale factors to get them into the right range, wouldn't we?

Here's my hack python test script -- the inputs are the raw "shifted-up-by-one" octals as would be printed in the listing, shifted to the right by two for the accumulator and one for the multiplicand.
```
acc = (0o134121730>>2) & (~1)
mcd = (0o275341220>>1) & (~3)

def delta1(a, m):
    b = (a >> 0) & 0o7
    if b in (0,7):
        return 0
    elif b in (1,2):
        return 2*m
    elif b == 3:
        return 4*m
    elif b == 4:
        return -4*m
    elif b in (5,6):
        return -2*m

def delta2(a, m):
    b = (a >> 2) & 0o7
    if b in (0,7):
        return 0
    elif b in (1,2):
        return 8*m
    elif b == 3:
        return 16*m
    elif b == 4:
        return -16*m
    elif b in (5,6):
        return -8*m

p = 0

for i in range(6):
    p += delta2(acc, mcd) + delta1(acc, mcd)
    p >>= 4
    acc >>= 4

print(oct(p & 0o377777777))
```

rburkey2005:
> Hmmm -- I have an easy way to test multiplication, but it's not immediately obvious to me what the best way to convert the floating point inputs to integers for the computation. We'd have to pick some scale factors to get them into the right range, wouldn't we?
>
> Here's my hack python test script -- the inputs are the raw "shifted-up-by-one" octals as would be printed in the listing, shifted to the right by two for the accumulator and one for the multiplicand.

Maybe, but I hope trial-and-error will see me through.  I'll let you know.  Tomorrow, hopefully.  Thanks!

rburkey2005:
@thewonderidiot Alas, that wasn't it (insofar as off-by-1 in the assembler was concerned), but it was worth a try.  While it's possible I didn't get the scaling just right, the discrepancy for large factors was much more than 1 bit, more than can be explained by poor scaling.

PR #1208, opened 2023-07-28
yaLVDC: Corrections for MPY partial products and DIV

thewonderidiot:
* MPY and DIV continuously update PQR, making it unsuitable for temporary storage while they're in progress. I added a countdown tracking their progress, and changed runOneInstruction() to update PQR at the end of every cycle with results. Ongoing MPYs and DIVs also inhibit interrupts from the LVDC side, which I don't think is happening yet, so I'm planning on dual-purposing this counter for interrupt checks when I get to fleshing out interrupt handling.
* The LVDC documentation claims that for MPY, the partial product of the multiplicand with the lower 12 bits of the accumulator is available to the 2nd instruction following the MPY. It turns out this is true, but not in the way we initially thought. This product is formed as though those lower 12 bits were the _most_ significant bits rather than the least significant bits.
image
Luckily, because I replaced yaLVDC's MPY implementation with the algorithm implemented in the hardware, we were already calculating the correct partial product and discarding it. It's not easy to replicate the garbage that shows up in PQR during the 1st and 3rd instructions following MPY, so what I've done for now is make this partial product appear for instructions 1-3, and the final product only appear for the 4th instruction.
* DIV scaling was off and has been corrected. This is giving good results and I haven't found any bad cases for it yet, so this may or may not be the final DIV implementation pending further self-tests. It has no such partial product requirement, so I made it put 0 into PQR for the first 7 instructions and only make the full result available for the 8th.

### DIFF
--- a/yaLVDC/runOneInstruction.c
+++ b/yaLVDC/runOneInstruction.c
@@ -50,6 +50,16 @@
  *                                  they have been replaced with a simple
  *                                  implementation of the algorithm used by the
  *                                  LVDC.
+ *              2023-07-28 MAS  More LVDC fixes:
+ *                              1)  MPY and DIV now continuously update PQR util
+ *                                  they are complete.
+ *                              2)  The partial product of the multiplicand with
+ *                                  the lower 12 bits of the accumulator available
+ *                                  the second instruction after MPY is scaled
+ *                                  differently from the final product. This
+ *                                  difference is now accounted for.
+ *                              3)  Scaling in DIV has been corrected to produce
+ *                                  correct results.
  */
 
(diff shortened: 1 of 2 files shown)


rburkey2005 (approved):

thewonderidiot:
I found a case for which yaLVDC's implementation of DIV is off of the correct value by 4:
```
       CLA     =(-.64)B0
       DIV     =(.9)B0
```
so unfortunately it's going to need to be replaced with a simulation of the hardware algorithm like MPY/MPH. I have said algorithm working in python, so hopefully it should be easy to translate to C.

PR #1209, opened 2023-07-29
yaLVDC: Simulate the hardware algorithm for DIV

thewonderidiot:
Division in the LVDC can give some slightly-off results, especially when negative signs are involved. For example, 00000000 / 377777777 = 377777774. In an attempt to achieve perfect agreement with the hardware simulation, I've replaced the implementation of DIV in yaLVDC with the modified nonrestoring division algorithm used in hardware.
image

This implementation is able to pass [all of the DIV self-tests I've put together so far](https://github.com/thewonderidiot/lvdc_simulation/blob/71cdd6f6606d72d1af834beb1d43bbedfdaf158a/self-test/self-test.lvdc#L1231).

### DIFF
--- a/yaLVDC/runOneInstruction.c
+++ b/yaLVDC/runOneInstruction.c
@@ -60,6 +60,11 @@
  *                                  difference is now accounted for.
  *                              3)  Scaling in DIV has been corrected to produce
  *                                  correct results.
+ *              2023-07-29 MAS  Replaced the implementation of DIV with a simulation
+ *                              of the non-restoring division algorithm implemented
+ *                              by the hardware. This was necessary to achieve
+ *                              consistent agreement with hardware simulation
+ *                              division results.
  */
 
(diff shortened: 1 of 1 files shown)


thewonderidiot:
As an aside, I accidentally re-learned while working on this that the automatic HOP save circuit uses the same delay line channels as some intermediate MPY/DIV calculations. If you try to `STO 776` or `STO 777` while there is a division going on, these intermediate calculations take precedence and what actually ends up in `776` or `777` is random garbage from these calculations. Subsequent attempts to HOP to them (to e.g. return from the subroutine you were calling while waiting for the DIV to finish...) causes you to have a bad day.

rburkey2005:
Fortunately, once upon a time, computer programmers were rocket scientists and never made mistakes like that.

rburkey2005 (approved):

PR #1212, opened 2023-08-02
yaLVDC: Add preliminary LVDA support, and refactor interrupts

thewonderidiot:
This PR contains a lot of tweaks, resulting from a pretty deep rabbit hole of adding a very simple (timers-only) LVDA implementation and getting interrupt behavior to match the hardware sim. Here's a rundown of what all has changed, and what I've learned about the LVDA.

Document sources are:
- [Laboratory Maintenance Instructions for LVDA, Volume 1: Description and Theory](https://www.ibiblio.org/apollo/Documents/19730064316_1973064316.pdf) - this document is frustratingly incomplete and lacks many theory-of-operation sections. Things not covered include interrupts, the RTC, and timers (i.e. almost everything that I care about in this PR).
- [Laboratory Maintenance Instructions for LVDA, Volume 2: Maintenance Data](https://www.ibiblio.org/apollo/Documents/LaboratoryMaintenanceInstructionsForLVDC-Volume2-MaintenanceData.pdf) - Despite the incompleteness of Volume 1, this volume includes essentially complete schematics for the LVDA. I've implemented 14 of the modules in the hardware simulator and all of it appears to be working without any changes. Much of what is covered in this PR was learned by poking at the LVDA in the hardware sim.
- [Astrionics System Handbook, Rev. November 15 1969](https://www.ibiblio.org/apollo/Documents/AstrionicsSystemsHandbook_Nov69.pdf) - The LVDA chapters in this document cover much of what Volume 1 of the maintenance instructions lacks. It has decently detailed descriptions of the timers and the interrupt processor.

### Interrupts
Interrupts now cause a real HOP instruction to occur. yaLVDC was formerly simply setting the current location to 400 immediately. To implement this, I generalized the `pendingEXM` logic to run any pending instruction, and made the interrupt logic queue up a pending `HOP 400`. Conveniently, the nextHOP field was easily dual-purposed to indicate the HOP of the interrupted instruction, although this did require some changes to the automatic HOP save (discussed below).

Interrupts are also now delayed by one cycle from when they were triggered. When the LVDA detects an interrupt and it is not inhibited, a bit is entered into a delay line and emerges one LVDC instruction cycle later. This bit sets INTC, the "master interrupt latch", which applies a discrete to the LVDC. The LVDC will then service this interrupt at its earliest convenience, pending local inhibits caused by various instructions.

### Interrupt Inhibit Cycles
To account for MPY inhibiting interrupts for 5 clock cycles and DIV inhibiting for 8, the `inhibitInterruptsOneCycle` variable has been changed to a counter. 

TRA, TMI, and TNZ no longer inhibit interrupts for one clock cycle. EXM _does_ inhibit interrupts for 1 cycle, but it no longer uses the counter because the current implementation forces both this cycle to occur in the same call to runOneInstruction() as the EXM itself.

MPH is a slightly weird case; it inhibits interrupts for one cycle longer than it lasts (it exits the fourth cycle after the initial MPY, but interrupts are inhibited for 5).

### Automatic HOP Save
The automatic HOP save circuit forms a mostly-complete HOP constant every cycle, lacking only the LOC field. If `STO 776` or `STO 777` are performed arbitrarily in the middle of a function, the current HOP is stored, but with the LOC entirely set to 000. Storage of LOC is enabled only by a HOP, TRA, TMI, or TNZ branch being taken. TMI and TNZ instructions that do not branch do not enable LOC storage.

Formerly, the hop save simulation was calculating its own LOC at the end of each runOneInstruction() call. It now always uses the contents of a new variable called `saveLOC`. This variable is set to 0 at the beginning of each runOneInstruction() call. If a branch occurs, the branching instruction copies `nextLOC` into `saveLOC` before changing `nextLOC` to the target location. This new behavior combines with `pendingEXM`'s setting of `nextLOC` in such a way that the HOP save correctly points to the interrupted instruction when an interrupt occurs.

### LVDA 
As far as I can tell there's not really any overlap between the LVDA and PTC IO channels. Rather than filling processInterruptsAndIO() with conditionals, I've opted to replace it with a simple stub that branches to either processInterruptsLVDAInterruptsAndIO() or processPTCInterruptsAndIO(), depending on the --ptc flag.

### PIO Channels
I'll discuss this more in the section on channel 072, but PIO channel bits 8 and 9 effectively have no meaning to the LVDA -- they only select the source of data (accumulator or memory) being sent from the LVDC to the LVDA. I.e. from the LVDA's perspective, PIO 072, PIO 272, and PIO 672 are all the same thing. I've changed the PIO code to mask off these bits for channel selection for the LVDA (but left them in place for the PTC because it appears they may be needed there).

#### 103 - Real-Time Clock (RTC)
This channel reads the 13-bit RTC that exists as part of the accelerometer processor in the LVDA. This clock increments precisely once every 246.09375 microseconds.... which at first blush is an extremely arbitrary time increment. But it works out to be exactly equal in length to 3 LVDC instruction cycles, so it's an easy frequency to simulate. 

The 13-bit RTC reading occupies the lower 13 bits of the 26-bit word read from the LVDA; the upper 13 bits are filled with garbage. I'm pretty sure this garbage is partly determined by accelerometer readings, but lacking accelerometers the hardware sim appears to have naturally defaulted to filling the top bits with all 1s (so the RTC counts from 377760000 to 377777777, and then wraps to back to 377760000). For starters, I've replicated this behavior in yaLVDC's LVDA.

#### 072 - Interrupt Inhibit
This LVDA register has essentially no special behavior; the value sent by the LVDC is directly written into the register, overwriting any previous contents. As mentioned above, PIO 072, PIO 272, and PIO 672 all directly refer to this same register, and all of them simply overwrite its contents. The only difference is if the LVDA sends the accumulator, the contents of address 272, or the contents of address 672, respectively. The fact that PIO 672 is regularly used to release inhibits appears to be a software convention with no basis in hardware.

#### 137 - Interrupt Latch
This register latches interrupts as they are appear. Interrupts that are inhibited by channel 072 are still latched here; the inhibit simply prevents the master interrupt latch INTC from being set. If an interrupt occurs while inhibited, it will remain in this latch and trigger an interrupt as soon as it becomes uninhibited.

**Important note:** The bits in channel 137 are shifted up by 1 from channels 072 and 032. This means that a minor loop timer interrupt will set channel 137 to 200000000, but if you want to inhibit or reset this same interrupt, you must write 100000000 to channels 032 or 072.

#### 032 - Interrupt Reset / Limiting
Any `1`s written to this register reset their associated interrupt bit in channel 137. Furthermore, if the interrupt in question is still being requested (i.e. if the interrupt was caused by a discrete going high, and that discrete is still high), an interrupt limiting bit begins circulating. This limiting bit acts as an inhibit, and persists until the source of the interrupt goes away. This stops interrupts from recurring from the same source.

#### 076 - Minor Loop Timer
The minor loop timer decrements at half the frequency of the RTC -- i.e. once every 6 LVDC instructions. When it hits 0, it generates an interrupt. The timer wraps from 0 to 17777.

#### 062 - Switch Selector Timer
This timer decrements at the same frequency as the minor loop timer, and also generates an interrupt upon reaching 0. However _unlike_ the minor loop timer, this timer stops at 0 and does not wrap.

### DIFF
--- a/yaLVDC/processInterruptsAndIO.c
+++ b/yaLVDC/processInterruptsAndIO.c
@@ -51,6 +51,9 @@
  *                              also that there were errors in the yaPTC.py
  *                              implementation of this that I fixed here.
  *              2020-06-05 RSB  Corrected CIO 065, 071.
+ *              2023-08-01 MAS  Split processInterruptsAndIO() into PTC and
+ *                              LVDA implementations. Added RTC, timer, and
+ *                              interrupt support for the LVDA.
  */
 
 #include <stdlib.h>
@@ -155,12 +158,11 @@ int typewriterMargin = 80;
 int typewriterTabStop = 5;
 int typewriterCharsInLine = 0;
 int
-processInterruptsAndIO(void)
+processPTCInterruptsAndIO(void)
 {
   int retVal = 1, channel, payload, lastInterruptLatch = 0;
 
-  if (ptc)
-    lastInterruptLatch = state.cio[0154];
+  lastInterruptLatch = state.cio[0154];
 
   // Some of these operations we want to performed entirely locally here,
   // in the yaLVDC program; some we want to be done entirely in the client,
@@ -175,10 +177,7 @@ processInterruptsAndIO(void)
 
       channel = state.pioChange;
       payload = state.pio[channel];
-      if (ptc)
-        interruptLatch = &state.cio[0154];
-      else
-        interruptLatch = &state.pio[0137];
+      interruptLatch = &state.cio[0154];
 
       //printf("here PIO channel %03o\n", channel);
 
@@ -480,7 +479,7 @@ processInterruptsAndIO(void)
       moreCIO: ;
     }
 
-  if (ptc && lastInterruptLatch != state.cio[0154])
+  if (lastInterruptLatch != state.cio[0154])
     state.prsParityDelayCount = 100;
 
   pendingVirtualWireActivity();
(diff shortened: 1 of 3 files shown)


rburkey2005 (approved):
Truly marvelous, thanks!

PR #1213, opened 2023-08-03
yaLVDC: Fix EXM such that its second cycle decrements counters

thewonderidiot:
This was a subtle bug that I found by pure luck. EXM's `goto reenterForEXM;` was bypassing both the PQR result and interrupt inhibit countdown timers. This means that if you have the instruction sequence
```
MPY   A
EXM   0,1,0
STO   B
CLA   775
```
before this fix, yaLVDC would not have the final product ready in 775 yet, even though it should.

It turns out that the AS-512 software uses precisely this sequence of instructions when constructing critical guidance matrices. After fixing this bug, yaLVDC is now capable of achieving a stable orbit in the AS-512 flight program's repeatable flight simulation mode, when started with this simple preflight program:
```
       ORGDD   0,0,0,000,0,0,0
E.PP1  CDS     D.LEID
       CLA     D.KB11
       STO     D.FMDI
       HOP     E.GP0
```
image
image

### DIFF
--- a/yaLVDC/runOneInstruction.c
+++ b/yaLVDC/runOneInstruction.c
@@ -90,6 +90,10 @@
  *                                  longer used in channel selection (they only
  *                                  determine if the accumulator or memory are the
  *                                  source of the data being sent to the LVDA).
+ *              2023-08-02 MAS  Moved PQR countdown to the start of runOneInstruction
+ *                              and positioned the reenterForEXM label such that both
+ *                              this countdown and the interrupt inhibit one are
+ *                              re-executed for EXM's second cycle.
  */
 
 #include <stdlib.h>
@@ -543,12 +547,22 @@ runOneInstruction(int *cyclesUsed)
     state.prsParityDelayCount++;
 
   // Set global variables providing background info on the emulation.
+  reenterForEXM: ;
   if (state.inhibitInterruptCycles > 0)
     state.inhibitInterruptCycles--;
   dataFromInstructionMemory = 0;
   instructionFromDataMemory = 0;
 
-  reenterForEXM: ;
+  if (state.mpyDivCount > 0)
+    {
+      // Copy ongoing MPY/DIV results into PQ
+      state.mpyDivCount--;
+      if (state.mpyDivCount > 0)
+        state.pq = state.pqPend1;
+      else
+        state.pq = state.pqPend2;
+    }
+
   if (state.pendingInstruction.pending)
     {
       state.pendingInstruction.pending = 0;
@@ -1144,15 +1158,6 @@ runOneInstruction(int *cyclesUsed)
       goto done;
     }
 
-  if (state.mpyDivCount > 0)
-    {
-      // Copy ongoing MPY/DIV results into PQ
-      state.mpyDivCount--;
-      if (state.mpyDivCount > 0)
-        state.pq = state.pqPend1;
-      else
-        state.pq = state.pqPend2;
-    }
   rawHopStructure.loc = nextLOC;
   rawHopStructure.s = nextS;
   if (!isHOP)


rburkey2005 (approved):
Yes, pretty dumb bug on my part.  But Awesome result.  (The capital-A in Awesome was a fumble-fingered typo because I just rolled out of bed and can barely even see the screen, but it's so _à propos_ that I'm leaving it in.)

rburkey2005:
@thewonderidiot What are the units on the axes of the graphs?

thewonderidiot:
Haha, sorry, those were pretty low-effort graphs made quickly. I can generate some better ones. The x axis in both is seconds; altitude y axis is km; velocity y axis is m/s.

I should also be able to generate a much less noisy altitude plot -- because the telemetry PIO isn't being simulated correctly yet, I made these plots by periodically reaching in and sampling the core memory values directly. But based on the data, it looks like a lot of times I sampled the position before it converged, or something. For the most part I ended up with multiple samples for the same internal LVDC timestep, and sometimes the first one is off:
```
time, alt, vel
687.556417, 172.159215, 7564.633399
689.159964, 166.649886, 7564.633399
689.159964, 172.131905, 7576.881413
689.975026, 172.131905, 7576.881413
689.975026, 172.131905, 7576.881413
691.597276, 166.603245, 7576.881413
691.597276, 172.105359, 7589.141153
691.597276, 172.105359, 7589.141153
```
I haven't done it yet, but I think the position chart would be significantly smoother if only the last position value for each timestep was used.

rburkey2005:
I was indeed wondering how it had been captured.  I was thinking of adding a feature to yaLVDC in which you could associate instruction addresses with variable addresses (and possibly a format specifier and a message string), and then every time the simulation hit one of the instruction addresses it would log the value of the variable (possibly formatted and with the message string) to a file, with a timestamp.  Though I'm not sure a feature like that would be used for anything other than this exact scenario, once!

thewonderidiot:
Hahaha yeah, probably better in the long run to get telemetry going -- these values are sent out over PIO. :smile: I'm working on adding the relevant modules to the hardware sim's LVDA now.

rburkey2005:
For AS-512 and AS-513, I've added the new Makefile target "repeatable.tsv".  It works exactly like a regular build of AS-512 or AS-513, except that it adds in a skeleton Preflight Program based on yours, and instead of creating the usual files creates an object file "repeatable.tsv" along with a listing file "repeatable.lst".  It's done in such a way that no memory locations for any existing program labels or variables are changed from the normal build, even though a HOP constant is transparently allocated in sector 2-17, and hence the listing shows just 5 mismatches (the 4 instructions and the HOP constant) vs. the transcribed octal listings.

I haven't emulated repeatable.tsv as of yet, since I don't have any way so far to capture the results, but I think the build works okay.

rburkey2005:
Perhaps the easiest way to capture the data would be to have an option in yaLVDC that simply logged all PIO instructions to a file.  A python script could be added afterward to analyze the logs.

rburkey2005:
It would look kind of like the SIM FLT MAL OAT report.

thewonderidiot:
Here's my [hacked-up processInterruptsAndIO()](https://gist.github.com/thewonderidiot/7be73ea63655aba3a1f6e290067ccd76) that was used to capture the data (search for "tbid"), if you want to reproduce the test quickly. Note that if you want to simulate AS-513, the `total_time` moved up from 0163 to 0161.

I think my statement on the last big MR that the LVDA doesn't care about bits 8 and 9 of the PIO channel are not entirely true -- they may be used as part of telemetry tag construction. I'm working on understanding it better now, but there will be changes along those lines coming shortly.

rburkey2005:
Thanks, that's helpful, and not completely dissimilar either.

rburkey2005:
I've pushed a change to yaLVDC that may step on your toes a little (sorry!).  There's now a CLI switch that creates a log of PIO outputs.  Use it like so:  `yaLVDC --log-pio=1 ...`.  That creates a file called yaLVDC.pio.

I've looked at the telemetry definitions (`TELD` in GFP_DATA_PACK) and the actual telemetry macro expansions (`TELM` in BOOST NAVIGATION), and I see that the following appear in the yaLVDC.pio log:

- `TELM D.VVS` (PIO 124) is the total space-fixed velocity.
- `TELM D.1OR` (PIO 054) is the inverse radius.
- `TELM V.TAS` (PIO 000) is the mission time at accelerometer read.

But there's something wrong with these numbers, because they flop around in an apparently-meaningless way.  I'm stopping for the day now, but I present it for whatever it's worth, which I presume is very little.

thewonderidiot:
There's a couple of problems you're probably running into:

- As I mentioned above, bits 8 and 9 matter for telemetry but need to be ignored for everything else. These bits are currently being indiscriminately discarded because I hadn't yet implemented anything needing them.
- PIOs are not unique. PIO 054 might give you D.V1OR, or it might give you V.TCI. This is why my test code is looking at raw addresses and not PIOs; just the PIO channel by itself is insufficient to know what you're getting. What should be unique is the resulting PCM tag, but that takes in other bits from the LVDA that I don't have a full understanding of yet.

These two problems are why I've switched gears to monkeying around with telemetry in the hardware sim, until I better understand how it's supposed to work. :slightly_smiling_face:

rburkey2005:
Discarding bits 8/9 from the channel numbers was inhibiting my ability to see PIOs with non-zero bits 8/9 in the log.  I haven't changed the discarding, but to get around that I've pushed a change to the way the logging was done, which again may step on your toes as far as your local changes are concerned.

rburkey2005:
I notice something wrong with the PIO logic in runOneInstruction(), in that it believes the PIO is an input if and only if the two least-significant bits of the channel number are 11.  However, in the MINOR LOOP I see code like this,
<pre>
       CLA     V.YAW+2 
       PIO     156                ISSUE YAW COMMAND
       CLA     V.ROL+2
       PIO     152                ISSUE ROLL COMMAND
</pre>
which are clearly outputs with the least-significant bits being 10.

rburkey2005:
I don't know if you had noticed these graphs from the AS-512 Flight Evaluation doc.  Pretty impressive agreement.

AS-512-altitude-flight-eval
AS-512-velocity-flight-eval

rburkey2005:
From the pretty-extensive discussion of telemetry [here](https://www.ibiblio.org/apollo/Documents/satinstunitibm_4.pdf#page=70), it appears to me that actual tag used for interpreting the telemetered data consists of the 9 bits of the PIO channel-number, plus 3 mode bits.  The actual value of the 3-bit mode-field appears to range from 0 through 4, and that presumably corresponds to the fact that the MODE field of TELD also ranges from 0 through 4.  

But it's hard to see how that MODE value gets incorporated into the telemetry, since what the TELM macro does with the MODE upon expansion seems to be merely (indirectly) to set the CPU's DS register differently to select one of the data sectors 2-11, 2-12, 2-13, 2-14, 2-15, or 2-16.  Is the DS register exposed externally by the CPU to the LVDA or elsewhere?

thewonderidiot:
Yeah, comparison with those flight evaluation report charts is how I knew it was finally working correctly. :smile: 

No, the DS register is not exposed to the LVDA. All of the connections between the LVDC and LVDA can be seen on PDF page 19 of Volume II of the maintenance instructions:
image

I don't have it working yet, but the real time register takes signals MODR1, MODR2, and MODR3 as inputs.
image

These appear to be bits from the mode register MODA, addressible with PIO 006 (or 206, or 606).
image

The bit numbering for the MOD register in the EDD does not match the bit numbering in the LVDA schematics.

rburkey2005:
That seems reasonable from the LVDC code.  And there's not a single PIO 006 in the PIO log.

rburkey2005:
I just got this velocity graph from the capture of PIO 124 in the PIO log, which for some reason is no longer flopping around madly.  Yes, it's scaled wrong vertically, but at least it's visually in the right ballpark.  PIO 124 doesn't seem to be one of those channels with an alternate interpretation.

image

rburkey2005:
Oh, I forgot that there was a reason I was posting the graph:  Besides not being scaled correctly vertically, there's a vertical offset, even at time 0.  Or actually at 27648 cycles, which is when the first data appears; the value PIO outputs at that point is about 00314201, and at time 0 it must be around 00314200.   Does that sound at all familiar?

thewonderidiot:
For the space-fixed velocity plot? Are you referring to earth's rotation rate?

rburkey2005:
I thought 124 was the space-fixed total velocity.  (I'm out for the afternoon and am speaking from memory.)Sent on my Samsung Galaxy A51
-------- Original message --------From: Mike Stewart ***@***.***> Date: 8/5/23  11:32 AM  (GMT-06:00) To: virtualagc/virtualagc ***@***.***> Cc: Ronald Burkey ***@***.***>, State change ***@***.***> Subject: Re: [virtualagc/virtualagc] yaLVDC: Fix EXM such that its second
  cycle decrements counters (PR #1213) 
For the space-fixed velocity plot? Are you referring to earth's rotation rate?

rburkey2005:
Okay, it actually turns out to be really easy to generate these graphs with the PIO log, without any choppiness, because none of the PIO's I needed had multiple interpretations and thus I didn't have to worry about any mode bits.  (The PIO log does have DS in it now, but I didn't use it.)  Those PIO's were:

- D.VVS (PIO 124), total space-fixed velocity.
- D.VXS (PIO 044), x component of space-fixed position.
- D.VYS (PIO 050), y component of space-fixed position.
- D.VZS (PIO 034), z component of space-fixed position.

I wrote [a little script](https://github.com/virtualagc/virtualagc/blob/master/yaLVDC/analyzeBoost.py) to just pull out the data from the log and do the minor mathematical massaging required, albeit with some confusion on my part about appropriate scaling.  And I assumed that the Earth remained at the origin of space-fixed coordinates, which doesn't seem right, but I didn't want to spend the time figuring it out more closely.  My velocity came out slightly high, and perhaps those factors have something to do with it.

Here are the graphs:

image
image

thewonderidiot:
Nice! Much smoother on the altitude.

You can pull the scaling out of the telemetry tags tables.
image

Just like the AGC, all LVDC numbers are expected to be fractional (i.e. between 0 and 1). `1.0` would theoretically be represented by `200000000`, if it were allowed to roll over into the sign bit. To arrive at the correct number, you multiply this fractional value by `2 ^ N`, where N is the scaling factor in the table. So in the case of total space-fixed velocity:
```
vel = (raw_value / 0o200000000) * (2**14)
```
which can be simplified to
```
vel = raw_value / 2048
```
You've got `raw_value / (2 * 1000)` in your script, which as you say is close but not quite.

rburkey2005:
Oh, I got confused and forgot I wasn't converting to kilometers.

rburkey2005:
@thewonderidiot I just ran it on AS-513.  The results look good, but just _after_ reaching orbit (_before_ I could do anything to escape from the simulation), LVDC execution aborted (while remaining in yaLVDC) with the following message:
<pre>
Fetching data from empty location 0-06-000

 HOP = 130212015 (ADR=2-03-0-050/0-06)   VAL = 00017   ACC = 202602320  P-Q = 202602320
(777)= ????????? (empty address      )  (776)= ????????? (empty address      )
 INT = 000000000  INB = 0
 RET = 130212515 (ADR=2-03-1-052/0-06)
Instructions: 7000732, Cycles: 7513286, Elapsed time: 616.324242 seconds
Source line: 3851
Code in memory (00017) has changed from load (03637).
From source:	2-03-0-050 2-03 03637   L.ESL0 CLA     D.KZRO       EXECUTE SYLLABLE 0 (STORE XXX)
Disassembled:	2-03-0-050 0-06 00017   L.ESL0  CLA     0-06-000, memory location is empty
</pre>

I haven't made any attempt to look into it further than that.

rburkey2005:
I've been looking for a bug in yaLVDC or in the AS-512 flight program, but I wonder if this isn't just something as simple as it expecting to find something in the Preflight Program that we haven't added to the skeleton preflight program.  The instructions at L.ESL0 and L.ESL1 are updated on an ongoing basis during telemetry.  Alas, I'm not sure exactly what that "something" would be, but it would seem to have something to do with telemetry, and something to do with terminating the flight-simulation mode.

rburkey2005:
As nearly as I can make out, though I'm still looking at it, it appears as though the subroutine `L.E305` is called when the Flight Program wants to begin transitioning from repeatable flt sim mode to normal flight mode.  In particular, it prevents further memory reads by `L.ESL0`.

By breakpointing it, I find that `L.E305` is indeed called in AS-512 (at 869.97 seconds), but in AS-513 the abort for an unallocated memory access (0-06-000) occurs before `L.E305` can be called.

If I alter the skeleton preflight program for AS-513 so that sector 0-06 is filled, the unallocated-memory abort does not occur, and `L.E305` is called at 657.7 seconds.

Although I haven't figured out a lot of the details here, such as exactly how `L.E305` gets called in the first place, or some extreme oddities in the telemetered data after orbit insertion but before `L.E305` is called, or why AS-513 is trying to telemeter sector 0-06, I'd tentatively infer that the unallocated-memory access is a bug in the skeleton preflight program rather than in yaLVDC or the AS-513 flight program, and that filling sector 0-06 in the skeleton preflight is the appropriate fix.

rburkey2005:
As far as the "extreme oddities" in altitude I mentioned are concerned, here's what the data looks like if you let the simulation run beyond orbit insertion through `L.E305`.  The oddities appear only in AS-512, with the big negative spike being at 825 seconds.  Each of the simulations stops telemetering the PIOs I need several sample times before reaching `L.E305`.

image

image

thewonderidiot:
L.E305 is the repeatable simulation termination function. It doesn't really do anything other than zero a whole bunch of memory, telemeter some final things, and then enter the one-instruction loop. It is triggered through the `G.TERM` entry of the time base 5 event time table. G.TERM is set to L.E305 for repeatable simulations only, and L.EP01 otherwise. For AS-512 it is scheduled to occur at TB5+145s.

I tried a couple of days ago to make it past TB5+145 in AS-512 by setting G.TERM to L.EP01 for repeatable simulations as well, but hit the same thing where telemetry code was trying to send down some uninitialized memory from the preflight program memory area. I think it was a different location though.

I think I have now implemented enough of the LVDA circuitry that LVDC telemetry should just about fully work (LVDA telemetry is still several modules away). I also managed to program the LVDC into an FPGA and get it working-enough that it passes most (all?) of the self-tests -- so I am nearing the point where I can run these repeatable simulations on hardware in real-time, to try to determine if there are any remaining yaLVDC inaccuracies at play.

rburkey2005:
Well, yes, I could see that `G.TERM` got loaded.  But I couldn't see any code that made use of `G.TERM` to actually do anything with its value.

rburkey2005:
@thewonderidiot I'm very confused about this business of dropping bits A8, A9 from the PIO channel number.  Specifically, the case in which A8=1.  And even more specifically, about `PIO 206` vs `PIO 006`.  `PIO 206` appears quite often in the log.  The way `runOneInstruction()` in yaLVDC.c is coded now, `PIO 2xx` is going to be treated like,
<pre>
       CLA     0xx        (i.e., get data from DM-DS-0xx)
       PIO     0xx
</pre>
(Without actually changing the value in the accumulator, of course.)  

Is that really the intended action of `PIO 2xx`?

thewonderidiot:
Yes, that's really the intended action. PIO 006, 206, and 606 all hit the mode register MODA in the LVDA. PIO 006 sends the data in the accumulator, while PIO 206 and PIO 606 send the data from address 206 in the current sector, or address 606 from residual memory, respectively. The data goes to the LVDA through either AI3V (accumulator) or TRSV (memory). The LVDA, for this address group, uses bits 1-2 and 8-9 to determine which of the two lines it should be listening to for the current PIO instruction.

thewonderidiot:
This obviously isn't globally true, of course, since bits 8-9 *do* matter for telemetry PIOs. I think they might only matter when bit 2 is set (i.e. PIO address xxxxxxx1x)? But it's a bit messy. Honestly what I had been contemplating was separating out telemetry PIOs from internal PIOs, and presenting two different "streams" of data to clients -- a telemetry stream that sends telemetry data along with the yaLVDC-constructed PCM tag, and a PIO interface where the client can receive/respond to non-telemetry PIO commands as currently designed (which would be all subject to bit 8-9 droppage).

rburkey2005:
Okay, so even if that is the intended behavior for 006/206/606, it's not the intended behavior for lots of other PIO channels, and thus the current coding that immediately strips bits A8/A9 out in runOneInstruction() is wrong. 

I don't see why two separate streams of PIOs are needed.  Why can't each individual client just interpret the full, unfiltered PIOs in the way it needs, ignoring them and/or stripping/not-stripping them as appropriate?  That has always been the way the virtual wiring has worked, in which clients are assumed to be smart enough to understand their own needs.

The only "extra" needed would seem to be that when the PIO is sent out over the virtual wiring, it needs to include not only the channel number and the data word, but also the values of some number of bits (3?) from the current mode register.  (And it only needs that to cover the possibility that the client may have connected to the host _after_ the PIO 006/206/606 occurred.)  At any rate, the virtual-wire protocol in virtualWire.c needs to be redefined from 6-byte packets to add a 7th byte to handle the mode-register bits.

I'm going to remove the bit-stripping from runOneInstruction().  As far as the changes to virtualWire.c are concerned, I've been trying to figure out how the 26 bits output by PIO 006/206/606 map to the 3 bits seemingly required by clients, and haven't been able to do so.  Until I understand that, I'm not sure it's worthwhile yet to change virtualWire.

rburkey2005:
It also turns out that yaLVDC's virtual-wire output had been broken for LVDC, though it was still working for PTC, so I've restored operation for LVDC.  The yaPTC.py program, which provides the PTC panel UI has also been modified to have a --terminal CLI option, allowing it to spy on yaLVDC's virtual-wire output.  The virtual-wire output does seem to be working now, in the sense that it reports the same channels and values as yaLVDC's PIO logging, but since it doesn't have the timestamping which yaLVDC's PIO logging has, it's not as good for trying to capture results from yaLVDC in a way that's plottable vs time.

That puts me in a position to start adding the mode-register values to the virtual-wire protocol, but I'll leave that for another day.

rburkey2005:
I realized belatedly that the mode does _not_ have to be included in each virtual-wire PIO packet.  All that needs to happen is that yaLVDC needs to transmit an initial PIO 006 packet to the client as soon as a client connects.  yaLVDC has now been fixed up to do that, with the virtual-wire protocol remaining the same as it has been all along.

However, I haven't been able to guarantee that the PIO 006 is the very _first_ thing transmitted &mdash; in fact, it always seems to be the 2nd packet transmitted &mdash; so client devices receiving virtual-wire transmissions need to assume that the mode is unknown until the first PIO 006/206/406/606 is received.

Off the top of my head, that would seem to conclude all of the work I felt I needed to do right now on yaLVDC, and everything I personally see as necessary right now concerning the virtual wiring.

rburkey2005:
@thewonderidiot Oops, I was so caught up in my own thinking that I misunderstood what you were saying about a separate telemetry stream!  You may be right.  It wasn't really necessary for yaAGC, since only a tiny specific set of channel numbers carried telemetry data, so they were very easy to pick off of the virtual wiring.  Whereas with yaLVDC there's a whole swath of PIO numbers carrying telemetry data, sometimes in an obscure fashion.

Still, it's not entirely clear that providing a separate data stream is superior to just providing a library function (say, in C and Python) that can can be used by client software to decode the virtual wiring properly and pull the telemetry out of it.  For example, I suspect that production of these client/server data streams may be the primary obstacle in clock multipliers for the emulator.  So as you say, it does need some thought.

rburkey2005:
@thewonderidiot I don't know if you've noticed that p. 3 of the AS-206RAM listing (MAIN.lvdc8) contains a description of the 40-bit telemetry word which is largely consistent with [the document we've talked about before](https://www.ibiblio.org/apollo/Documents/satinstunitibm_4.pdf#page=70), though I'm not sure it's 100% so.

One thing it notes is that "BITS 20, 1, AND 2 [of the 40-bit telemetry word] ARE THE 3 LOW ORDER BITS OF THE MODE REGISTER", which from my reading would be the least-significant 3 bits output by PIO 006.  Which is most-definitely _not_ the particular 3 bits I would have expected.  Though it's possible with the funky bit-numbering system used that it may have meant the most-significant 3 bits of PIO 006 instead.

Have you been able to determine how to extract the 3 mode-register bits from PIO 006?

thewonderidiot:
> Oops, I was so caught up in my own thinking that I misunderstood what you were saying about a separate telemetry stream! You may be right. It wasn't really necessary for yaAGC, since only a tiny specific set of channel numbers carried telemetry data, so they were very easy to pick off of the virtual wiring. Whereas with yaLVDC there's a whole swath of PIO numbers carrying telemetry data, sometimes in an obscure fashion.

Part of my reasoning there was that we might also want to support LVDA telemetry. That telemetry is _usually_ triggered when the LVDC does a PIO, but the data is sourced from within the LVDA, and is prepended with some bits from the real-time clock. It felt to me like it might make sense to handle all of this tag and clock construction in the local LVDA implementation where it is all readily available.

> @thewonderidiot I don't know if you've noticed that p. 3 of the AS-206RAM listing (MAIN.lvdc8) contains a description of the 40-bit telemetry word which is largely consistent with the document we've talked about before, though I'm not sure it's 100% so.

I hadn't, and will have to check it out, thanks!

> Have you been able to determine how to extract the 3 mode-register bits from PIO 006?

It's using the low-order 3 bits of the 6-bit mode register, which gets filled with the 6 high-order bits of the PIO word. So for example, if I run
```
L13P1  CLA     =O35
       STO     206
       CLA     =O21
       PIO     006
       PIO     206
       TRA     *
```
in the hardware sim, I get
image

where MODR1, MODR2, and MODR3 are the three mode register bits used in tag construction. They appear to correspond directly with M2, M1, and M3 in that figure in the EDD.

rburkey2005:
> It's using the low-order 3 bits of the 6-bit mode register, which gets filled with the 6 high-order bits of the PIO word. So for example, if I run ...

Or in other words, 5?  That had occurred to me afterward as an alternate reading of what it says in AS-206RAM, though I didn't know that the mode register was 6 bits and I didn't like to just keep theorizing ignorantly.  That does seem to be the most-consistent interpretation of the comments in AS-206RAM at the points where the `PIO 006` instructions appear.

Sounds right, thanks!

rburkey2005:
In case it's of any value, I've thrown together a first cut at a Python module for pulling telemetry out of a stream of PIOs, maintaining the 3 mode bits as an internal state.  It's the new file yaLVDC/lvdcTelemetryDecoder.py, and should be pretty self-explanatory to use.  For convenience in testing, I've also integrated it into yaLVDC/analyzeBoost.py, the program I use for pulling altitude/velocity curves out of PIO logs from yaLVDC, so that analyzeBoost.py could alternately be used for pulling general telemetry of a PIO log.  To analyze general telemetry, the usage is

      analyzeBoost.py --telem=VERSION <yaLVDC.pio

where `VERSION` is an integer:  1 for AS-206RAM, 2 for AS-512, 3 for AS-513.  The `--telem=VERSION` switch is omitted to instead continue pulling out inertial velocity and altitude curves.

Here's a brief excerpt from the AS-512 BOOST log:
<pre>
...
4.356   V.DCZT  0.000000        B0      PIRADS  DELTA CHI Z
4.380   V.THXT  0.000000        B0      PIRADS  ROLL GIMBAL ANGLE
4.385   V.THYT  0.000000        B0      PIRADS  PITCH GIMBAL ANGLE
4.390   V.THZT  0.000000        B0      PIRADS  YAW GIMBAL ANGLE
4.395   V.CCXT  0.000000        B0      PIRADS  MINOR LOOP RATE-LIMITED CHI X
4.422   V.CCYT  0.000000        B0      PIRADS  MINOR LOOP RATE-LIMITED CHI Y
4.427   V.CCZT  0.000000        B0      PIRADS  MINOR LOOP RATE-LIMITED CHI Z
4.432   V.EARX  0.000000        B0      .05 M/SEC       EXTRA X ACCELEROMETER READING
4.437   V.EARY  0.000000        B0      .05 M/SEC       EXTRA Y ACCELEROMETER READING
4.463   V.EARZ  0.000000        B0      .05 M/SEC       EXTRA Z ACCELEROMETER READING
4.468   V.EART  O377763121      B0      RTC PULSES      REAL TIME CLOCK READING ASSOCIATED WITH EXTRA ACCEL READ
4.474   V.EMR   O000000000      B25     OCT     ERROR MONITOR REGISTER STATUS WORD (EB)
4.478   V.LOTW  O000000000      B0      OCT     MODE CODE 24 (EB)
...
</pre>

In no way do I imagine this is complete and perfect as it stands, but it's a start.

rburkey2005:
I've pepped Python PIO-to-telemetry parser described above a bit, and added a C port of it as well (lvdcTelemetryDecoder.c,h).  Using the C API to provide LVDC telemetry parsing in another program should be straightforward, though it's also possible to build it as a stand-alone test program.  The test program parses the PIO log files from yaLVDC, producing identical results to the Python version, if we overlook differences in the least-significant digit of the timestamp due to Python's non-conformist rounding algorithm.

thewonderidiot:
@rburkey2005 Something else that you may not have found yet, but that you may like to include in your telemetry display -- the [Universal Saturn Launch Vehicle Systems Handbook for AS-508 and Subsequent](https://www.ibiblio.org/apollo/Documents/GH-36-MSC-01829-19700220.pdf) includes the meanings of all of the bits in the mode code registers starting on PDF page 238. As far as I can tell, the definitions are mostly (alll?) still valid for AS-512. Mode code 25 in particular is fun to watch tick along during a repeatable flight simulation.

rburkey2005:
I know what you mean about MC25.  :smile:  The bits for all of the mode code words are listed in the source code too, for AS-512 with MC25 being on PAGE 019.  I was thinking that eventually there might be some less crude reporting on some of the telemetered items, including the mode code words, but I hadn't thought to do anything about it right now.

At the moment I'm wrestling with the opposite direction:  Digital Command System (DCS) commands sent to the LVDC.  It turn out that while [the documentation of it in the EDD](http://www.ibiblio.org/apollo/Documents/satinstunitibm_4.pdf) looks pretty authoritative, its really severely lacking because it doesn't cover most of the commands, and some of the ones it does cover aren't applicable to AS-512.  The formats of almost all commands sent to the LVDC will have to be reverse engineered from the LVDC source code.  And that may turn out to be harder than is warranted.

indy91:
@rburkey2005 We have some information about the Saturn V uplink format. In the document with the translated LVDC code (NTRS ID 19720013551) there are these two tables:

image
image

And in this document on the Honeysuckle Creek website (https://www.honeysucklecreek.net/msfn_missions/Apollo_11_mission/AS-506_NOD_Supplement.html) there are even some octals for uplinks. But I am not sure if that also applies to the AS-512 software and if it is exactly the data the LVDC wants to receive. It starts on PDF page 235.

PR #1215, opened 2023-08-15
yaDSKYb1 Fixes

CaptainSwag101:
This fixes several inaccuracies and other issues with yaDSKYb1, found when developing my own Block I AGC emulator:
- The COMP FAIL light illuminates instead of the COMP ACTY light
- The SCALER FAIL light illuminates instead of the TM FAIL light
- The flashing pace of the Verb/Noun display was too slow, documentation stating "roughly once per second" was exaggerating: it's actually the FS17 timing signal from the scaler, which has a frequency of 0.78125 Hz, or a period of 1.28 seconds (640 ms on, 640 ms off).
- Launching the program would display a warning message about an incorrect grid sizer alignment object.
- There was an unnecessary 200 ms delay in part of the packet-sending code (recommended to remove by @thewonderidiot )

### DIFF
--- a/yaDSKYb1/Makefile
+++ b/yaDSKYb1/Makefile
@@ -20,7 +20,7 @@ endif
 all default:	yaDSKYb1
 
 yaDSKYb1: yaDSKYb1-widgetized.cpp yaDSKYb1-widgetized.h ${EXTRASOURCE} ${EXTRAHEADERS} Makefile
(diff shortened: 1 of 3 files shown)


CaptainSwag101:
Thanks for the mention in the Virtual AGC changelog! :smile: 
@rburkey2005  Two small notes, though: My first name is "James", and the emulator I'm developing for the Block I AGC is a software-based emulator, but I have been referencing the hardware schematics to sort out inconsistencies or uncertainties where needed, such as for the DSKY flashing and other things like the hardware alarm logic. However, the emulator itself operates at a lower level than others I've seen out there because it functions at the "control pulse" or sub-instruction microcode level. So while not a simulation of the exact hardware logic by any means, it is definitely an interesting middle point between that and a higher-level, instruction-logic emulator like I think yaAGC is.

PR #1218, opened 2024-01-29
yaAGC: Implement internally-timed RADARUPTs

thewonderidiot:
RADARUPTs are somewhat unique in the AGC in that unlike other interrupts for external hardware, all of the timing for the interrupt and the actual generation of the interrupt occurs entirely within the AGC. Strictly speaking, no hardware even needs to be attached to the computer for a radar cycle to complete and a RADARUPT to be generated.

It turns out that certain ropes, especially Skylark, are very sensitive to the exact timing of this interrupt. If they appear too early or too late, it's possible to get program alarms, or even for running programs to silently break and become completely non-functional. As such, I think it makes sense to add accurate RADARUPT simulation to yaAGC.

The timing for the interrupt presented in this PR is as accurate as I can get it without deeper changes to yaAGC. The real conditions for RADARUPTs are:
1. RadarGateCounter == 9
2. F05B
3. ~FS06 && FS07 && FS08 && FS09 && FS10

yaAGC currently only increments its scaler at 3200Hz, but to see F05B we'd have to double it to 6400Hz. So instead I've placed it very slightly (~10 MCT) early, which shouldn't be have any bad effects.

I decided to implement the function that actually gets radar data as a callback for integrators like NASSP to provide their own implementation for. The existing fictitious channel stuff didn't seem applicable since fictitious input channels are processed asynchronously from the computer logic, and fictitious output values aren't going to return a value. Also, it didn't appear that there was currently any way at all for external users to actually generate RADARUPTs, so I went ahead and just made that always happen. If we want to provide for the possibility of it not happening, we can instead push some or all of the RADARUPT generation/side effect code into the callback. I'm open to other ways of doing this, though, if you have better ideas.

@indy91 has flown a Skylab rendezvous with Skylark using these changes, and they have greatly improved the experience (i.e. Skylark no longer totally breaks). He's also flown a landing and everything works as expected with the landing radar.

### DIFF
--- a/yaAGC/NullAPI.c
+++ b/yaAGC/NullAPI.c
@@ -41,6 +41,9 @@
 				the GPL, for linking to Orbiter SDK libraries.
 		05/14/05 RSB	Corrected website references.
 		05/31/05 RSB	Added ShiftToDeda.
+		01/29/24 MAS	Added a stub RequestRadarData function for use
+				by integrators of yaAGC into spacecraft
+				simulations.
 */
 
 #ifdef WIN32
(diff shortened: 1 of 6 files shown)


rburkey2005:
:+1:

PR #1223, opened 2024-05-14
Add source reconstruction of Comanche 67

thewonderidiot:
Much easier than Skylark! My overall impression from looking at the dump is that the parts we were sure about in the reconstruction attempt, we indeed got correct.... but the parts we were unsure of, we were mostly wrong and pretty much didn't stand a chance.


rburkey2005:
:+1:

Issue #1179, opened 2022-08-26
Reconstruct Comanche 72 from Comanche 67


thewonderidiot:
### Changed Bugger Words from Closest Program (Comanche 67)
```
Bugger word mismatch in bank 01; actual 42616 != expected 42675 (diff = 00057)
Bugger word mismatch in bank 04; actual 71744 != expected 71330 (diff = 00414)
Bugger word mismatch in bank 06; actual 56656 != expected 56652 (diff = 00004)
Bugger word mismatch in bank 07; actual 46502 != expected 46500 (diff = 00002)
Bugger word mismatch in bank 10; actual 47263 != expected 47257 (diff = 00004)
Bugger word mismatch in bank 11; actual 73456 != expected 73473 (diff = 00015)
Bugger word mismatch in bank 12; actual 46261 != expected 53020 (diff = 04537)
Bugger word mismatch in bank 13; actual 52653 != expected 46536 (diff = 04115)
Bugger word mismatch in bank 14; actual 57433 != expected 64436 (diff = 05003)
Bugger word mismatch in bank 15; actual 71421 != expected 71377 (diff = 00022)
Bugger word mismatch in bank 16; actual 67526 != expected 67540 (diff = 00012)
Bugger word mismatch in bank 17; actual 65757 != expected 65762 (diff = 00003)
Bugger word mismatch in bank 20; actual 66674 != expected 40652 (diff = 26022)
Bugger word mismatch in bank 21; actual 73176 != expected 73202 (diff = 00004)
Bugger word mismatch in bank 22; actual 64526 != expected 64530 (diff = 00002)
Bugger word mismatch in bank 23; actual 21333 != expected 21331 (diff = 00002)
Bugger word mismatch in bank 24; actual 76767 != expected 65161 (diff = 11606)
Bugger word mismatch in bank 25; actual 72422 != expected 72424 (diff = 00002)
Bugger word mismatch in bank 26; actual 74406 != expected 74423 (diff = 00015)
Bugger word mismatch in bank 27; actual 61634 != expected 61646 (diff = 00012)
Bugger word mismatch in bank 30; actual 77025 != expected 77037 (diff = 00012)
Bugger word mismatch in bank 31; actual 65305 != expected 55030 (diff = 10255)
Bugger word mismatch in bank 32; actual 45660 != expected 45661 (diff = 00001)
Bugger word mismatch in bank 34; actual 50336 != expected 50337 (diff = 00001)
[...]

PR #1224, opened 2024-05-19
Reconstructed Comanche 72

thewonderidiot:
As described in #1179, here's my reconstruction of Comanche 72 from the Comanche 67 dumps. Unfortunately since it is forward-looking and Artemis is so different, there are three areas of ambiguity in this one where different implementations may result in the same banksum:

- **PCR-963, R52 - Delete 407 alarm and drive trunnion to 50°** - we don't know the exact value of the new constant `TRUNLIM` introduced in this. I found a solution assuming octal 25252 for the constant, but if it was instead 25253, then a different implementation would be required. I couldn't find one in my experimentation but that doesn't mean there's not one.
- **COM-26 - Extended verb 92 flag 6 is changed while job is not in INHINT** - This INHINT can be placed in a total of 4 places, and each of them result in the same checksum. I picked the one I think makes the most sense, but we have no way of knowing where it really was at the moment.
- **COM-29 - N70 instead of N71 display in P23** - The arguments for the bit-checking portion of this change can appear in the opposite order and still have the same checksum. I.e. instead of `CA FLAGWRD9; MASK SAVECBIT`, it could have been `CA SAVECBIT; MASK FLAGWRD9`.

The remaining changes are (in my opinion) pretty unambiguous, so if this isn't exactly right it should be very close. This also hasn't been tested in NASSP yet, so there is some (small) chance an error will be found and need to be fixed when that happens.


rburkey2005:
@thewonderidiot Excellent, thanks!  Given the uncertainties involved, I'll let it sit a few days before writing it up.

Issue #1180, opened 2022-08-29
Reconstruct MANCHE72 rev 3 from Comanche 72


thewonderidiot:
### Branch
TBD

### Changed Bugger Words from Closest Program (Comanche 72)
```
Bugger word mismatch in bank 06; actual 56652 != expected 50373 (diff = 06257)
Bugger word mismatch in bank 11; actual 73473 != expected 51224 (diff = 22247)
make: *** [Makefile:21: diffComanche072sums] Error 2
```

### Known Differences (from Apollo 13 FSRR slides)
* PCR 984 -- Avoid Coarse Align during Saturn

### Resources
* [PCN 1041](https://www.ibiblio.org/apollo/Documents/PCN-1041.pdf)
* [Colossus Memo 226 - Revisions 102 through 104](https://www.ibiblio.org/apollo/Documents/COL-unnumbered-042770.pdf)
* [Apollo 13 FSRR Material](https://www.ibiblio.org/apollo/Documents/apollo_13_fsrr_slides.pdf)
* [Artemis 72 program listing](https://archive.org/details/Artemis072J2k60)

### Notes
According to the FSRR slides, MANCHE72 rev 3 differs from Comanche 72 by only a single change -- PCR 984. We don't have a copy of PCR 984 itself, but we _do_ have a copy of PCN 1041, which was a modification of PCR 984. 

Artemis 72 includes the code for both PCR 984 and PCN 1041 in the [T4RUPT PROGRAM](https://github.com/virtualagc/virtualagc/blob/master/Artemis072/T4RUPT_PROGRAM.agc#L596) section. It's easy to identify as such because [Colossus Memo 226](https://www.ibiblio.org/apollo/Documents/COL-unnumbered-042770.pdf) details the implementation of PCR 984 and PCN 1041 in Comanche 102-104.
```AGC
		EBANK=	DAPDATR1
		CAF	EBANK6
		TS	EBANK
		CS	DAPDATR1
		MASK	PRIO30
		CCS	A
		TCF	CALLCRS			# NO

		CAF	AVEGBIT			# YES, IS AVEG ON
		MASK	FLAGWRD1
		CCS	A
		TCF	NOGIMRUN		# YES, DON'T COARSE ALIGN
```

The second chunk of this code is from PCN 1041, which was not incorporated into MANCHE72. So, the trick with this one is to figure out (a) how the above code looks different without the second chunk, and (b) how they managed to fit this into one module without breaking other modules.

[...]

PR #1225, opened 2024-05-21
Reconstruction of Manche72 revision 3

thewonderidiot:
Here's a reconstruction of Manche72 revision 3, with the changes described in #1180. I think this solution is a likely one, but the addition to bank 11 is so large (12 words!) that I would be a bit surprised if there weren't alternate implementations that lead to the same banksum. So, as with Comanche 72, I think there is ambiguity here that we won't be able to rid ourselves of until we find more sources.


thewonderidiot:
@indy91 has done a little bit of testing of this one in NASSP, with more to come soon. The parts he's checked out so far have worked. :crossed_fingers:

rburkey2005:
Okay, great!  I'll write up both Comanche 72 and Mache72 rev 3 with those provisions.

I'm wondering, though ... are you so confident that there weren't workable alternate instruction-sequences on _prior_ revisions?  It always seemed to be a possibility before, but I don't recall it being significant enough to have extra attention called to it.  So is the risk of it really bigger now with 72 and 72/3, or is it that you've become more sensitized to the possibility?

thewonderidiot:
My general feeling is that the more code we have to write ourselves, the more likely we are to get it wrong. A _lot_ of the reconstructions we've done in the past have involved writing essentially no new code whatsoever -- for those, we essentially just selectively copy-pasted original code from other revisions. Some of them have required new code, but only a small handful of instructions, in which the number of possible solutions that achieve the same banksum is very small. Comanche 72 required adding new code where e.g. we didn't know the placement of an INHINT from a later rope, and didn't know the exact value of a constant we had to add. And this one just requires so many new words that the solution space is very large.

rburkey2005:
More now, got it!  Thanks.

PR #1226, opened 2024-05-21
update README.md with correct paths and instructions to build for linux

erentar:
I was confused how to run the simulator and had to fiddle with the VirtualAGC/VirtualAGC binary and paths to make it work, hopefully this will prevent others from going through the same process.


rburkey2005:
@erentar Thanks!  As the README itself notes, its instructions are not up-to-date ... but as long as they're there, corrections are good.

PR #1229, opened 2024-07-15
add (void) to unprototyped function declarations ..

gavineadie:
For yaAGC only: Add the "void" to "function declarations without a prototype is deprecated in all versions of C" ..


rburkey2005 (approved):
@gavineadie Indeed, these were all lines of code I didn't write personally, so at least that makes _me_ feel a little better about them.  :smile:  Thanks!

PR #1230, opened 2024-07-16
Add Alarm Messages

NeilFraser:
Prints messages when the master alarm is triggered, and indicates what the type of the alarm.

Several alarms are reboots are silently being triggered in [example code](https://www.ibiblio.org/apollo/DIY.html#Bare-Metal_Programming).  Raising visibility of them is a prerequisite in either fixing the example code, or else fixing the AGC if the alarms are not correctly firing.

### DIFF
--- a/yaAGC/agc_engine.c
+++ b/yaAGC/agc_engine.c
@@ -2027,6 +2027,7 @@ agc_engine (agc_t * State)
                   State->SbyStillPressed = 1;
 
                   // While this isn't technically an alarm, it causes GOJAM just like all the rest
+                  printf("Alarm: Standby\n");
                   TriggeredAlarm = 1;
 
                   // Turn on the STBY light, and switch off the EL segments
@@ -2054,6 +2055,7 @@ agc_engine (agc_t * State)
           if (!State->Standby && State->NightWatchman)
             {
               // NEWJOB wasn't checked before 0.64s elapsed. Sound the alarm!
+              printf("Alarm: NightWatchman\n");
               TriggeredAlarm = 1;
 
               // Set the NIGHT WATCHMAN bit in channel 77. Don't go through CpuWriteIO() because
@@ -2102,6 +2104,7 @@ agc_engine (agc_t * State)
           else if ((State->RuptLock || State->NoRupt) && 0300 == (0777 & State->InputChannel[ChanSCALER1]))
             {
               // We've either had no interrupts, or stuck in one, for 140ms. Sound the alarm!
+              printf("Alarm: Rupt (%i %i)\n", State->RuptLock, State->NoRupt);
               TriggeredAlarm = 1;
 
               // Set the RUPT LOCK bit in channel 77.
@@ -2117,6 +2120,7 @@ agc_engine (agc_t * State)
           else if ((State->TCTrap || State->NoTC) && 000 == (037 & State->InputChannel[ChanSCALER1]))
             {
               // We've either executed no TC at all, or only TCs, for the past 5ms. Sound the alarm!
+              printf("Alarm: TC (%i %i)\n", State->TCTrap, State->NoTC);
               TriggeredAlarm = 1;
 
               // Set the TC TRAP bit in channel 77.


rburkey2005 (approved):

rburkey2005:
Sure, why not?  Thanks!

rburkey2005:
> We ran into some confusing alarms when running the bare metal code on this page:
https://www.ibiblio.org/apollo/DIY.html#Bare-Metal_Programming
>
>The night watchman is fine, it's being tickled in two different places.  However, there are two separate GOJAM restarts occurring.
>
>There's a TC Trap alarm caused by the ALLDONE loop.  My understanding is that this should not happen since the TC Trap is only supposed to catch one-line loops.  Adding a dozen NOOP statements doesn't resolve this alarm.
>
>There's also a NoRupt alarm being fired at a lesser frequency than the TC Trap.
>
>Most interestingly, changing the "CS NEWJOB" line to "CA NEWJOB" (in the ALLDONE loop) resolves both issues.  We are at a complete loss to explain this behaviour.
>
>It's unclear whether the bare metal code and the documentation is wrong, or if the virtual AGC is wrong.  My suspicion is the latter.

I'm forced to admit that the "bare-metal" documentation probably is very much out of date with respect to these issues, not least of which because I doubt anybody has ever used it before.  I wonder if @thewonderidiot  mightn't have more insight into the problems you're addressing than (unfortunately!) I do, since they sound similar to problems I know that he has had to address in the past.

thewonderidiot:
> Sure, why not? Thanks!

@rburkey2005  I think they "why not" is that this sort of thing isn't desirable for simulator integrators like NASSP. I think this would be better be hidden by a debug compiler flag, or something like that.

> My understanding is that this should not happen since the TC Trap is only supposed to catch one-line loops. Adding a dozen NOOP statements doesn't resolve this alarm.

That's not really true. There are two cases of TC trap, one of which is bugged in the original hardware.
1. _Only_ TC or TCF are executed in a given interval. This can be any number of different TCs or TCFs. As you've discovered, even code like
```
LOOP  NOOP
      NOOP
      NOOP
      NOOP
      NOOP
      NOOP
      TC      LOOP
```
will trigger this, since that whole loop is composed of only TCs and TCFs.

Adding NOOP isn't doing anything for you because NOOP is not a real instruction -- if located in fixed memory, it is assembled as `TCF +1`. So by adding NOOPs, you've actually just added more TCFs to your loop.

This half of the hardware alarm works by checking to see if TC or TCF is the active instruction during time 4. This works exactly how they intended.

2. _No_ TCs are TCFs are executed in a given interval. In theory, this was meant to catch loops like
```
      CAF     ZERO
LOOP  EXTEND
      BZF     LOOP
```
However, it has a very serious design flaw in that it doesn't sample the `TC0` and `TCF0` nets at a specific time, unlike the other half of the alarm. This means that brief transients on those two nets, which frequently happen during instruction changeover, are sufficient to trick this alarm into thinking that TCs or TCFs have in fact been executed, when really they haven't. The most simple case (shown above) does _not_ generate any such transients, which is why they never caught this (at least during the design process).

> Most interestingly, changing the "CS NEWJOB" line to "CA NEWJOB" (in the ALLDONE loop) resolves both issues. We are at a complete loss to explain this behaviour.

This is a bug related to our attempted simulation of the somewhat-broken "No TCs" portion of the alarm. The way I implemented this originally treats transients the same way it treats real TCs/TCFs. This achieves proper simulation of the broken half of TC Trap, but unintentionally breaks the "working" half, because it's now thinking that things that generate transients are real TC/TCF instructions.

The specific behavior you're seeing is because selection of the `CS` instruction always generates a transient on `TC0`, while selection of `CA ` generates no such transients.

The fix for this is to treat transients differently from real TC/TCFs -- it should be pretty straightforward. I'll put something together tonight.

rburkey2005:
> I think the "why not" is that this sort of thing isn't desirable for simulator integrators like NASSP. I think this would be better be hidden by a debug compiler flag, or something like that.

That was my first thought as well, but I checked and saw that `agc_engine` (where these messages were added)  was already printing other messages for `PINC` and `MINC` and so on, so it seemed to me that whatever mechanism was swallowing those messages would work identically for these newer, presumably much less frequent, messages.

thewonderidiot:
Those print statements can never occur and almost certainly get optimized out, unless code changes are made. All of them are gated behind `TrapPIPA`, which is set to 0 by default and can never be set to 1 without uncommenting some lines. Those prints will never be encountered by simulators. These new ones will, several times per flight at least.

rburkey2005:
Apparently I didn't check it thoroughly enough.  A corollary of that sloppy checking is that if I tried to enable the messages via a command-line parameter, I'd probably botch it the first few times, and I'm not in the mood for that.  I've done the next-best thing, which is to put them into conditional compilations, _à la_
<pre>
#ifdef TRAP_ALARMS
    printf(...);
#endif
</pre>
So they'll be entirely absent unless compiled with `-DTRAP_ALARMS` on the compiler command line.  Is that a good enough compromise for everybody?

thewonderidiot:
Works for me!

PR #1256, opened 2025-06-11
Proof Luminary099/UPDATE_PROGRAM (pg: 1386-1396)

Hexman768:
### Description 
* These changes are whitespace corrections for Luminary099/UPDATE_PROGRAM. 
* These changes are based on reduced quality scans for the UPDATE_PROGRAM in Luminary099 [here](https://www.ibiblio.org/apollo/ScansForConversion/Luminary099/)


rburkey2005:
@Hexman768 The changes look fine, I think, but something has occurred to me that may require additional thought before proceeding.  The issue is the overall alignment &mdash; i.e., the left margin &mdash; of full-line comments _vs_ end-of-line comments.

The original AGC software developers would have had the expectation that after assembly, the listing produced by the assembler would have aligned the full-line comments with the end-of-line comments in a mutually-consistent manner.  For example, looking at the very end of Luminary099 UPDATE_PROGRAM, the original assembly listing looks like so:
image
The mutually-consistent alignment that I'm talking about is that the final L in "... ALL EXITS FROM ..." aligns with the T in "TURN OFF ..." and the leading parenthesis of "(BIT 3 OF CHANNEL 11)".

But the "modern" assembler doesn't produce the same alignment, so that it ends up looking like this instead:
image

Now that I've written all of that out, I suppose the fix for this doesn't have anything to do with the corrections you're making, but instead means that the assembler ought to align the full-line comments 63 columns to the left of where it puts them now.  I guess I didn't do that originally simply because I didn't think the whitespace was going to be transcribed correctly anyway, but it's too long ago to really know what I was thinking.  I'm afraid that the reason I didn't do it before was that it breaks something else.

Any thoughts?

Hexman768:
I will also note that I am not super familiar with the development process of AGC source code, so I'm not sure how helpful I can be in the original assembler vs. modern assembler situation. I think as long as historical accuracy is preserved, we can go whichever way you feel would be best.

rburkey2005:
Upon reflection, it appears to me that you should just continue doing as you have done, and I'll worry about how (or if) to modify the behavior of the modern assembler.  So I'm going to merge this request right now.  The comments below replying to your comments can probably be ignored without consequence.

> Are there instances where we want to develop these pieces of code with the modern assembler, or are we attempting to preserve the historical accuracy of the source code/code comments as it originally was?

> I will also note that I am not super familiar with the development process of AGC source code, so I'm not sure how helpful I can be in the original assembler vs. modern assembler situation. I think as long as historical accuracy is preserved, we can go whichever way you feel would be best.

In that case, I suppose I should make it clear that in no case whatsoever do we have the original source code.  That was on computer punch cards.  All we have (in the <i>best</i> case, which fortunately includes Luminary099) is the printout from the original assembler, and that printout itself contains only a reformatted version of the original source code.  Nor did the punch cards have anything like the leading '#' character that we use in front of comments, but did have certain columnar requirements which we don't have in our files.  So in the end, I'm afraid the best we can hope for is consistent columnar alignment among the various comments.

> ... we definitely want to eliminate tabs in comments ...

Tabs are uniformly treated by the modern assembler &mdash; they didn't exist at all in the days of computer punch cards &mdash; as advancing to the next group of 8 columns.  So I don't think they introduce any particular problem if present, but I think you're right that for consistency it's nicer to use spaces instead.

PR #1257, opened 2025-06-19
Proof Comanche055/PINBALL_GAME__BUTTONS_AND_LIGHTS (pg: 307-317)

Hexman768:
### Description 
* These changes are whitespace corrections for the PINBALL_GAME_BUTTONS_AND_LIGHTS program. 
* These changes are based on reduced quality scans [here](https://www.ibiblio.org/apollo/ScansForConversion/Comanche055/)
* This is the first PR of 7 to address the whitespace issues in this program, after that is done, I can begin making sure that the complete changes in the program get patched over this program anywhere else required. 

### Special Note
* I noticed that the PINBALL program actually seems to be named incorrectly (unless I am mistaken). I thought it was supposed to be PINBALL_GAME_BUTTONS_AND_LIGHTS. But there seem to be two underscore symbols after GAME in the program name in the repo. Is this correct?


Hexman768:
I now see that there are two spaces after GAME at the top of page 307 in the scans. I am guessing this is the reasoning behind the naming convention. Please disregard the note in the PR message.

rburkey2005:
> I now see that there are two spaces after GAME at the top of page 307 in the scans. I am guessing this is the reasoning behind the naming convention. Please disregard the note in the PR message.

I don't actually know the rationale.  I originally named the source-code files in a more-simple fashion, and the filenames were updated later by other people to more-closely reflect the original names of the software modules.  In other words, the filenames were corrected in much the same fashion as you are now correcting whitespace in comments.  I spot-checked the changes to convince myself that they made sense, but I did not question them in detail.

PR #1259, opened 2025-06-25
Proof Comanche055/PINBALL_GAME__BUTTONS_AND_LIGHTS (pg: 329-349)

Hexman768:
### Description 
* These changes are whitespace corrections for the PINBALL_GAME_BUTTONS_AND_LIGHTS program. 
* These changes are based on reduced quality scans [here](https://www.ibiblio.org/apollo/ScansForConversion/Comanche055/) 
* This is the 3rd of 7 PRs created to address the whitespace issues in this program.


rburkey2005:
> Also just a note: If you did want a build to run for every PR, I know there are GitHub workflows that I have had great success with in the past for other projects. It should work here as well.

The deal is that the build service had been provided by something called Travis CI, which _used to_ have a free tier for open source projects, but now no longer has one.  So the automated-build workflows we had been using &mdash; not just for pull requests, but for all commits of any kind &mdash; stopped being performed when the free tier was discontinued.  I simply didn't know that that had happened, since there seems to have been no notification of any kind.

If you know of an alternative workflow that uses something free &mdash; as opposed to Travis CI, which I have no intention of paying for &mdash; that would be welcome news.

PR #1264, opened 2025-11-09
add Continuous Integration testing with CMake across OS and compilers

scivision:
Using free GitHub Actions, add continuous integrations self-tests across Windows, macOS and Linux and Clang, GCC compilers.

Note: merge #1263 first as there was a glitch in the CMake script.


rburkey2005:
Thanks!

PR #1267, opened 2025-11-12
CI: Windows: use only the preinstalled GCC

scivision:
This avoids need for third-party actions as per https://github.com/virtualagc/virtualagc/pull/1264#issuecomment-3511456563


rburkey2005:
Thanks for working through this.

PR #1286, opened 2026-04-04
Add an initial working docker version

changux:
Adding a working docker deployment for VirtualAGC


rburkey2005:
Cool, thanks!  I'm accepting this in advance of actually trying it out, since it's impossible for it to break anything, but I'm looking forward to doing so.

changux:
Glad to hear from you!

PR #1287, opened 2026-04-04
Update README.md with the docker deployment option

changux:


rburkey2005:
@changux For what it's worth, I added a short write-up about this to the Virtual AGC website.  The write-up is very short, since it says little other than to read your README.  It's more of a chronicle of my very short acquaintanceship with it.  At any rate, if you're interested in seeing what was said, [start with the change log](https://www.ibiblio.org/apollo/changes.html) and proceed from there along the link it provides you.

changux:
Happy to see it. I'm Polkan García, a curious engineer from Colombia.

rburkey2005:
Cool, Polkan, thanks again!  I've updated my write-up to credit you more directly than before.

changux:
Awesome! Thank you so much. Regarding the Docker environment, I tested the scripts on Linux (Ubuntu 22, 24 and 25), Windows 11 with WSL and macOS Tahoe (arm64 and x86) with excellent results. I will improve soon the documentation and other kind of validation scripts.

rburkey2005:
Oh, good.  I'll update my own write-up with the info about Windows and Mac.  I had meant to ask, but had forgotten to do so.  Another question that I hadn't gotten around to asking:  Is there any way to adjust the resolution of the image?  I gather that it's 1600&times;900, but that ends up looking pretty crummy when it scales up on my own display.  Probably it varies for everybody, so it would be nice if it could be customized.

changux:
Great! The resolution depends on each setup and screen. You can try different resolutions by modifying line 7 of the `start.sh` [script](https://github.com/virtualagc/virtualagc/blob/master/Docker/start.sh).

`Xvfb :1 -screen 0 1600x900x24`

And next, try again with: `docker-compose up`

rburkey2005:
"Windows with WSL":  Does that mean that docker was running in Windows, or that docker was running in Linux?

changux:
Sorry for the imprecision. This applies to Windows with the Windows Subsystem for Linux, as well as to the native Docker Desktop for Windows.

rburkey2005:
:+1:

rburkey2005:
I like this docker-container concept for running Virtual AGC on Windows and Mac (and Linux) very much, but never pursued it because back in the time I was thinking about it, it seemed to me as though supporting graphical applications was either impossible or very difficult.  So when I grew tired of having to support Windows and Mac OS, I chose to create a virtual machine rather than a docker container.  But the virtual machine is a _big_ download, and expands to a _much bigger_ image in operation, as well as chewing up CPU and memory.  So a docker container is a much lighter way of going about it.

I've experimented enough with your docker container to see that I can call up a terminal window and run command-line programs in it like yaYUL, yaAGC, and so forth.  Right now the container is limited in that it must run the VirtualAGC GUI program, and stops when the VirtualAGC GUI is closed, but if instead it would act more like the Virtual AGC VM, it seems to me that it could be a superior replacement for the VM.  I'm not sure it could provide _all_ of the VM's capabilities, but certainly most of them.

I can see that the container already presents a desktop-like environment, even though there are no icons on it.  Is it possible to modify start.sh (or whatever) so that instead of running the VirtualBox GUI program, it instead provides an environment more like the VM?  (Of course, it doesn't have to be a change to apollo11-demo.  It could be a new type of container, called say "virtualagc-vm".)

I'm not asking you to _do_ that, of course, just trying to understand the possibilities.

changux:
Ronald,

Today, the limits of container technology are infinite. Containerisation itself (with Podman, Containerd, Docker or Kubernetes, for example) is revolutionary. Virtual machines are still needed, but they are hard to maintain and not as modular as containers are by nature. A container is not a virtual machine, but it can mimic almost 100% of functions. The idea is that we want small, portable pieces of software that can work with other containers and scale easily (using _Infrastructure as Code_ for example). 

I configured the `VirtualAGC` program as a starting point, which is why, when you close it, the container closes too. That is its purpose. Be volatile. So, I suggest creating a development container in which to experiment with potential new Docker Compose scripts and containers. Remember that the idea here is to make your user data persistent (user and configuration files) and make the 'static' part disposable. In conclusion, I can work on that and help you in that direction if you are OK with it.

BTW, I wrote this brief note about the port — please take a look.

https://www.linkedin.com/pulse/from-4kb-moon-back-your-terminal-polkan-garcia-lazte?trk=public_post_feed-article-content

rburkey2005:
It won't let me read it unless I rejoin linkedin ... the membership of which I dumped just last week because of the increasing amount of annoying spam I had been getting from them.  Perhaps you can send me a PDF or whatever.  :smile:

But the idea of fleshing out a docker container to perhaps ultimately replace or at least provide an alternative to the VM is pretty compelling, and I'm sure I'll want to get to it at some point.  The problem right _now_ is I have some more-urgent problems it would be easy to distract me from. Since I can't afford to be distracted, I'll have to put off the docker container idea until later, and perhaps just do some experimenting in the meantime.  There's also a longer-range problem in that if I create the thing then I have to maintain the thing, and as you can probably tell, all I know how to do with docker right now is to slavishly follow somebody else's instructions.  Trouble-shooting other people's problems with it?  Not so much.

changux:
Understood. I can help in the meantime, don't worry.

rburkey2005:
> It won't let me read it unless I rejoin linkedin ... the membership of which I dumped just last week because of the increasing amount of annoying spam I had been getting from them. Perhaps you can send me a PDF or whatever. 😄

Here's a fun fact:  Even though I had gone through the rigmarole of deleting my linkedin account to avoid their spam, the spam has just resumed.  So I checked, and my account is back!  Apparently, it is not possible to get out of linkedin, except possibly via Hawking radiation.  So as long as I was there, I went ahead and looked at your note.  Nice!

changux:
Easy to get in, hard to get out. Thank you and again, count on me.

PR #1294, opened 2026-04-21
Clean up gitignore

ThymoNL:
The repo had some leftover files that should've been ignored.
I have:

- Removed the leftovers (types were part of the gitignore already)
- Ignored all core files, not just the ones in the yaAGC directory
- Simplified the gitignore


rburkey2005 (commented):
@ThymoNL A digression:  Even when I can't immediately see why a file might be present, I'm leery in general about concluding that such a file needs to be deleted.  Back in the Windows 95 days, I used to have a neighbor who would call upon me to get her PC up and running after it had suddenly become inoperative somehow.  A mystery!  I almost invariably found that she had intentionally deleted some of the system files, explaining that she needed the extra space and didn't know what those files were, thus concluding that it was safe to get rid of them.  Which left me scrambling to figure out what those files had been, and to find copies of them somewhere.  Remarkably, I always did succeed in the end, albeit after a lot of wasted time.  :smile:  Okay, end of digression.

In the case of all these *.cbp files, those are configuration files for the Code::Blocks debugger.  We only have them for a few AGC versions because they were all manually created a long time ago.  A couple of months ago, I was trying to work out how to automate their creation, so that we could have them for all of the AGC versions.  I got distracted, and haven't yet finished that up.  At least, I'm pretty sure I never finished it, or else there would be a lot more of them!  At any rate, if they were somehow in .gitignore, perhaps that was a mistake I made at some point but never noticed because the *.cbp files haven't been changed in so long.  Regardless, few if any of them should be deleted.  

Is presence in .gitignore the only criterion you used for determining that a file could be safely deleted?

ThymoNL:
I used `git ls-files -c -i --exclude-standard` to find all files in the index that are in the gitignore. The gitignore was the primary factor in my reasoning to delete them. IMO no files in the gitignore should be checked in unless a comment in the gitignore is added to state a reason otherwise.

Regarding the cbp files I assumed they were generated automatically. I would suggest removing the cbp files from the gitignore.

ThymoNL:
Actually I do see that the Makefile produces some cbp artifacts. Those should likely not be checked in. How would you like to handle those?

https://github.com/virtualagc/virtualagc/blob/master/Makefile#L705

YUL/aurora12_memory_dump.lst
rburkey2005, line 1: It doesn't appear to me that this file was manually created, nor that it can be automatically generated.  The fact that it just happens to be named *.lst seems to me to be accidental.  The reason *.lst files are normally ignored is that .lst is a filename extension I personally use for capturing reports generated by yaYUL, but that's not what this file is.  It's not anything I ever worked on personally so I don't know for sure, but I suspect that the file is here for the purpose of validating the operation of the ARGUS version of YUL, if work on that is ever completed.  Since this area of the repo isn't being actively worked on, if this file were to be mistakenly deleted it mightn't be discovered for many years, at which point recovery would be irritating to whoever has to do it ... assuming anyone remembered that it existed.  I don't think it should be deleted.

XCOM-I/RONLIB.xpl
rburkey2005, line 1: This file wasn't in .gitignore, so I don't see a rationale for deleting it.  I admit that I don't think it's doing anything at present, but the comments in it deal with a subject I find very difficult, so I'd prefer to keep the file, even if it's just for the comments alone.

XCOM-I/XPLI.LIBRARY.xpl
rburkey2005, line 1: This file wasn't in .gitignore, so I don't see a rationale for deleting it.  I admit that I don't think it's doing anything at present, but the comments in it deal with a subject I find very difficult, so I'd prefer to keep the file, even if it's just for the comments alone.

templateAGC-bottom.cbp
rburkey2005, line 1: This file is used in the Makefile in order to generate the AGC version-by-version .cbp files.  It should not be deleted.

templateAGC-top.cbp
rburkey2005, line 1: This file is used in the Makefile in order to generate the AGC version-by-version .cbp files.  It should not be deleted.

rburkey2005:
Regarding the .cbp files _other_ than the templateAGC*.cbp ones, I think you're probably right and I was confused about them.

XCOM-I/XPLI.LIBRARY.xpl
ThymoNL: It is being explicitly ignored by name in XCOM-I/.gitignore:67

XCOM-I/RONLIB.xpl
ThymoNL: It is being explicitly ignored by name in XCOM-I/.gitignore:66

ThymoNL:
I kept the template cbp files and also the .lst in Pultoraks block 1 simulator. That was used as input for the program and not generated from anything. I also added a negate for all binary files in the rope dump folder as those do need to be tracked.

YUL/aurora12_memory_dump.lst
ThymoNL: Agreed. Keeping it.

templateAGC-bottom.cbp
ThymoNL: Retained

templateAGC-top.cbp
ThymoNL: Retained

XCOM-I/RONLIB.xpl
rburkey2005: Oops!  I was looking a the changes and not the full .gitignore.  Okay, let's get rid of it.

XCOM-I/XPLI.LIBRARY.xpl
rburkey2005: Yes, let's delete it.

rburkey2005:
Okay, let's let it simmer overnight (for me, anyway), and if there's no further movement in the meantime I'll merge it tomorrow morning.  Thanks!

PR #1302, opened 2026-05-01
Issue #1296: XCOM-I MONITOR(9,10,12): bit-exact string<->IBM hex DP conversion

ColanderCombo:
Replaces the IEEE-754 round-trip in MONITOR(10) (HAL/S literal parsing) and MONITOR(12) (HAL/S runtime print) with direct IBM-hex DP arithmetic, ref issue #1296

This patch replaces only the conversion paths.  ibmFloat.[ch] provides implementations of IBM DP float arithmetic that can be used to replace the MONITOR(9) float functions and instruction emulation in inline360 when desired.

  - New XCOM-I/ibmFloat.[ch]: library of bit-exact IBM hex DP primitives -- ibm_dp_{add,sub,mul,div,addsub}

    Based on the FP logic in the Hercules/Hyperion 390 emulator, implemented here:

	  https://github.com/hercules-390/hyperion/blob/master/float.c

    Although not identical (no exception handling, f.e.) we follow the logic of the functions closely.

  - Also in XCOM-I/ibmFloat.[ch]:
    - ibm_dp_from_string: based on MONITOR.ASM/XXXTOD.bal, should match the behavior of the 360 compiler exactly
    - ibm_dp_to_double: closest double approximation of IBM hex value
    - ibm_dp_to_hal_string

  - runtimeC updated to use ibmFloat for MONITOR(10) and MONITOR(12)
    - MONITOR(10): parses literals to IBM hex floats for storage in the literal table.  We also leave the converted IBM hex -> double value in FR[0]
    - MONITOR(12): formats IBM hex floats into strings

### DIFF
--- a/XCOM-I/XCOM-I.py
+++ b/XCOM-I/XCOM-I.py
@@ -711,6 +711,7 @@ if targetLanguage == "C":
         except:
             pass
         for name in ["runtimeC.c", "runtimeC.h", "inline360.c", "inline360.h",
+                  "ibmFloat.c", "ibmFloat.h",
                   "debuggingAid.c", "Makefile.template"]:
             shutil.copy(basePath + name, outputFolder + os.sep + \
                         name.replace(".template", ""))
--- a/XCOM-I/ibmFloat.c
+++ b/XCOM-I/ibmFloat.c
@@ -1,0 +1,574 @@
+/*
+ * License:     The author (Donald Schmidt) declares that this program
+ *              is in the Public Domain (U.S. law) and may be used or
+ *              modified for any purpose whatever without licensing.
+ * Filename:    ibmFloat.c
+ * Purpose:     IBM hex floating-point arithmetic and conversions.
+ *              Based on the floating point logic in the 
+ *              Hyperion/Hercules IBM 390 & z/Series emulator:
+ *                  https://github.com/hercules-390/hyperion/blob/master/float.c
+ * Reference:   http://www.ibibio.org/apollo/Shuttle.html
+ */
+
+#include <assert.h>
+#include <math.h>
+#include <stddef.h>
+#include <stdio.h>
+#include "ibmFloat.h"
+
+#define twoTo56 (1LL << IBM_DP_MANT_BITS)
+#define twoTo52 (1LL << 52)
+
+static uint64_t ibm_dp_from_uint(uint64_t v);
+
+// Convert a C double to IBM double-precision float, returning the (msw, lsw)
+// pair.  On overflow returns (IBM_DP_OVERFLOW_MSW, 0). 
+// NOTE: IBM float has 56 bit mantissa, IEEE754 double is 53 bits,
+//       so this loses precision.
+void __attribute__ ((no_instrument_function))
+ibm_dp_from_double(uint32_t *msw, uint32_t *lsw, double d) {
+    assert(msw != NULL && lsw != NULL);
+    int s; // The sign: 1 if negative, 0 if non-negative.
+    int e; // Exponent.
+    uint64_t f;
+    if (d == 0) {
+        *msw = 0x00000000;
+        *lsw = 0x00000000;
+        return;
+    }
+    // Make x positive but preserve the sign as a bit flag.
+    if (d < 0) {
+        s = 1;
+        d = -d;
+    } else
+        s = 0;
+    // Shift left by 24 bits.
+    d *= twoTo56;
+    // Find the exponent (biased by IBM_DP_EXP_BIAS) as a power of 16:
+    e = IBM_DP_EXP_BIAS;
+    while (d < twoTo52) {
+        e -= 1;
+        d *= 16;
+    }
+    while (d >= twoTo56) {
+        e += 1;
+        d /= 16;
+    }
+    if (e < 0)
+        e = 0;
+    if (e > IBM_DP_EXP_MAX) {
+        *msw = IBM_DP_OVERFLOW_MSW;
+        *lsw = 0x00000000;
+        return;
+    }
+    f = llround(d);
+    *msw = (s << 31) | (e << 24) | ((f >> 32) & 0xffffffff);
+    *lsw = f & 0xffffffff;
+}
+
+
+double __attribute__ ((no_instrument_function))
+ibm_dp_to_double(uint32_t msw, uint32_t lsw) {
+    int s; // sign
+    int e; // exponent
+    long long int f;
+    double x;
+    s = (msw >> 31) & 1;
+    e = ((msw >> 24) & IBM_DP_EXP_MAX) - IBM_DP_EXP_BIAS;
+    f = ((msw & 0x00ffffffLL) << 32) | (lsw & 0xffffffff);
+    x = f * pow(16, e) / twoTo56;
+    if (s != 0)
+        x = -x;
+    return x;
+}
+
+
+// Convert a small unsigned integer to IBM DP packed format.  
+static uint64_t
+ibm_dp_from_uint(uint64_t v) {
+    if (v == 0) return 0;
+    int exp = IBM_DP_EXP_BIAS + IBM_DP_MANT_HEXDIGITS;
+    while (v >= ((uint64_t)1 << IBM_DP_MANT_BITS)) { v >>= 4; exp += 1; }
+    IBM_DP_RENORMALIZE_56(v, exp);
+    if (exp < 0) return 0;
+    if (exp > IBM_DP_EXP_MAX) return IBM_DP_OVERFLOW_PACKED;
+    return IBM_DP_PACK(0, exp, v);
+}
+
+
+// ibm_dp_from_string():
+// convert a HAL/S decimal-literal string to IBM DP hex.  Faithful
+// reproduction of MONITOR.ASM/XXXTOD.bal — the routine the HAL/S-FC
+// compiler invokes via MONITOR(10) at PASS1.PROCS/SCAN.xpl:690 to
+// parse every numeric literal into the literal table.
+//
+// Final scaling follows XXXTOD.bal:1530-1700:
+//   TIMES10A  while |dec_exp| >= 23: MD/DD F0,=D'1E23'
+//   TIMES10B  build F4 = 10^(0..22 remainder) via repeated squaring
+//   TIMES10C  MDR/DDR F0,F4
+//
+// =D'1E23' is IBM Assembler-H's evaluation of the literal `1E23`.
+// True 10^23 × 16^14 / 16^20 = 5,960,464,477,539,062.5 — exactly
+// halfway between two DP mantissas — and Assembler-H rounds half-up,
+// so the constant lives at 0x54152D02C7E14AF7
+//
+static const uint64_t IBM_DP_1E23 = 0x54152D02C7E14AF7ULL;
+
+void __attribute__ ((no_instrument_function))
+ibm_dp_from_string(const char *s, uint32_t *msw, uint32_t *lsw) {
+    assert(s != NULL && msw != NULL && lsw != NULL);
+    uint8_t sign = 0;
+    if (*s == '+') s++;
+    else if (*s == '-') { sign = 1; s++; }
+
+    uint64_t F0 = 0;
+    uint64_t TEN = ibm_dp_from_uint(10);
+    int frac_digits = 0, seen_dot = 0, ignore = 0;
+    while (*s) {
+        if (*s >= '0' && *s <= '9') {
+            if (!ignore) {
+                F0 = ibm_dp_mul(F0, TEN);
+                if (*s != '0') {
+                    F0 = ibm_dp_addsub(F0, ibm_dp_from_uint((uint64_t)(*s - '0')), 0, 1);
+                }
+                if (seen_dot) frac_digits++;
+                // XXXTOD precision guard (MONITOR.ASM/XXXTOD.bal lines
+                // 26-32): once F0 reaches the SP equivalent of ~7.2e15
+                // (=X'4E19999A') and we're past the decimal point,
+                // ignore all remaining digits.  Without this, every
+                // additional digit costs another truncating *=10
+                // multiply, accumulating LSB error past the 56-bit
+                // mantissa's natural ~17-decimal-digit precision.  Note
+                // pre-decimal digits never trigger ignore — they affect
+                // magnitude and must always be processed.
+                if (seen_dot
+                    && ((uint32_t)(F0 >> 32)) >= 0x4E19999AU) {
+                    ignore = 1;
+                }
+            }
+            s++;
+        } else if (*s == '.' && !seen_dot) {
+            seen_dot = 1;
+            s++;
+        } else {
+            break;
+        }
+    }
+
+    int dec_exp = 0;
+    if (*s == 'E' || *s == 'e') {
+        s++;
+        int es = 1;
+        if (*s == '+') s++;
+        else if (*s == '-') { es = -1; s++; }
+        int v = 0;
+        while (*s >= '0' && *s <= '9') { v = v * 10 + (*s - '0'); s++; }
+        dec_exp = es * v;
+    }
+    dec_exp -= frac_digits;
+
+    if (IBM_DP_MANT(F0) == 0) { *msw = 0; *lsw = 0; return; }
+    if (sign) F0 |= IBM_DP_SIGN_BIT;
+
+    if (dec_exp != 0) {
+        int absexp = dec_exp < 0 ? -dec_exp : dec_exp;
+        // TIMES10A: chunk by 23 using the Assembler-H 1E23 constant.
+        while (absexp >= 23) {
+            F0 = (dec_exp > 0)
+                 ? ibm_dp_mul(F0, IBM_DP_1E23)
+                 : ibm_dp_div(F0, IBM_DP_1E23);
+            absexp -= 23;
+        }
+        // TIMES10B / TIMES10C: build F4 = 10^(0..22) via squaring,
+        // then MDR/DDR.
+        if (absexp > 0) {
+            uint64_t F4 = ibm_dp_from_uint(1);
+            uint64_t F2 = TEN;
+            while (absexp > 0) {
+                if (absexp & 1) F4 = ibm_dp_mul(F4, F2);
+                absexp >>= 1;
+                if (absexp) F2 = ibm_dp_mul(F2, F2);
+            }
+            F0 = (dec_exp > 0) ? ibm_dp_mul(F0, F4) : ibm_dp_div(F0, F4);
+        }
+    }
+
+    if (F0 == IBM_DP_OVERFLOW_PACKED) { *msw = IBM_DP_OVERFLOW_MSW; *lsw = 0; return; }
+    if (IBM_DP_MANT(F0) == 0) { *msw = 0; *lsw = 0; return; }
+    *msw = (uint32_t)(F0 >> 32);
+    *lsw = (uint32_t)F0;
+}
+
+
+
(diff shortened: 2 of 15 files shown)


rburkey2005:
I think that if I merge this it will break "HALSFC --test ... source.hal", which works by comparing output files from HAL_S_FC.py vs HALSFC-PASS1, because HAL_S_FC.py (or at least its xplBuiltins.py) will still be using the old IBM hex conversions for its MONITOR(10) and MONITOR(12) functions.

rburkey2005:
Actually, I'm not sure the comparison would fail.  The most obvious place for it to fail would be in comparison of the literal files, but the literal files aren't quite compatible and so "HALSFC --test" doesn't do that particular comparison.  But comparison is possible with slightly more effort (with intervening processing by unLitfile.py) and probably will be implemented soon.  But at any rate it's inconsistent even if the test might not currently detect the difference.

rburkey2005:
Actually, that makes me keen now that it has occurred to me to add the literal-file comparison to "HALSFC --test" ....

rburkey2005:
I was right to begin with.  It would break "HALSFC --test" if "DEBUG &cent;E" used in the HAL/S source code, the entire literal table is printed out in the compiler report.  The hex values would then not match and the test would fail.

rburkey2005:
> Actually, that makes me keen now that it has occurred to me to add the literal-file comparison to "HALSFC --test" ....

I couldn't resist the temptation, and have just done so.  Note that the new commits are entirely independent of the changes you're looking at, and so should have no conflicts.

rburkey2005:
@ColanderCombo 

No thoughts on this?  Perhaps the `--test` feature seems like minor functionality, if you don't happen to use it. But I can't afford to break it, because it's the principle and invaluable tool I personally use in detecting and isolating compiler bugs.

ColanderCombo:
@rburkey2005  Since the diffs are presumably the result of the new implementation being more correct, it seems likely we'll have to fix the python implementation to match the behavior of ibmFloat.c.  I see two options:

- I can mechanically translate the ibmFloat.c implementation into an equivalent python.  That should satisfy the immediate need, but means we'll have to carefully keep the two implementations in sync.
- We could use a single shared IBM FP implementation with python bindings

I need IBM FP in multiple places;
  - XCOM-I for HAL/S-FC, which we just did (C)
  - nsts-sim-gpc, for the AP-101 implementation (nodejs)
  - future AP-101 C/C++ simulator (C/C++)
  - (and now) python HAL/S-FC port (python)

Instead of keeping multiple implementations in sync I've been considering making a single small library that I can use in all three places.  I've done a prototype 'libibmhexfp' library in C with bindings generated for node and python.  If that sounds like a possibility for you, we could adjust both XCOM-I and the python port to use that library.  (We could do a real library dependency or just copy the source directly into the virtualagc tree, although that does make keeping everything in sync more inconvenient.)

[note: 360 and AP-101 floating point arithmetic do have slightly different semantics.  The library I put together is aware of this, and has different entry points for 360/AP-101 where necessary.]

rburkey2005:
> * We could use a single shared IBM FP implementation with python bindings

While this makes sense efficiency-wise, I don't much like the idea, because there are use cases for HAL_S_FC.py as a very lean standalone program that's cross-platform with no installation requirements, and I don't want to burden it with non-Python stuff that has to be built in some manner.

> * I can mechanically translate the ibmFloat.c implementation into an equivalent python.  That should satisfy the immediate need, but means we'll have to carefully keep the two implementations in sync.

I can shoulder a little of the burden by handling creation and maintenance of an ibmFloat.py that mirrors ibmFloat.c. 

_On a separate matter,_ though, I realize I've badly overreached in insisting on a narrow specialization to `MONITOR(10)` and `MONITOR(12)`.  `MONITOR(9)` is still going to be needed if we expect compile-time stuff like `INITIAL(2 (1 + 3))` to produce correct literals.  That still seems to me like quite a "small" pull request rather than the big one that freaked me out originally.

Why was the original pull request so big?  In memory it seemed to involve changes to patch files patchNN.c, changes to the patch-recommendation system, and maybe even other stuff I don't remember.  It was really all of the patch stuff that drove me crazy, because patches are so hard to get working, and so hard to debug when they're not working right; and for that matter, so hard to even figure out that they might not be working in the first place.

rburkey2005:
> I can shoulder a little of the burden by handling creation and maintenance of an ibmFloat.py that mirrors ibmFloat.c.

Yikes!  I sent ibmFloat.c through [CodeConvert](https://app.codeconvert.ai/home) (which I had never heard of before) to create ibmFloat.py.  It's actually the first time I've used A.I. for any substantive (beyond Google A.I. summaries, which are baloney half the time).  If the notes it produced during the process are any indication, it did a better job than I would have, and has left me almost nothing to do other than to test it.

I guess what I'll do is create main programs for both ibmFloat.c and ibmFloat.py, feed them in a bunch of test cases, and check that they produce the same answers.

The result is basically just a Python module (which I'm anticipating will be located in XCOM-I/) in which all of the functions and variables have the same names as in ibmFloat.c.

ColanderCombo:
Could you drop a copy of the converted py file?  I can take a look at it. Sometimes AI does a great job (especially on mechanical things like this), but I still like to examine it carefully before using it. :)

Most of what's in nsts-sdl-dps is a catalog of tests, and I already have some for the ibmFloat.c implementation.  I could add a parallel test for the python.  (Happily, AI is really good at generating test data, if you're specific enough!)

> The result is basically just a Python module (which I'm anticipating will be located in XCOM-I/) in which all of the functions and variables have the same names as in ibmFloat.c.

Yeah, honestly it should all fit in a single file.

ColanderCombo:
> On a separate matter, though, I realize I've badly overreached in insisting on a narrow specialization to MONITOR(10) and MONITOR(12). MONITOR(9) is still going to be needed if we expect compile-time stuff like INITIAL(2 (1 + 3)) to produce correct literals. That still seems to me like quite a "small" pull request rather than the big one that freaked me out originally.
> 
> Why was the original pull request so big? In memory it seemed to involve changes to patch files patchNN.c, changes to the patch-recommendation system, and maybe even other stuff I don't remember. It was really all of the patch stuff that drove me crazy, because patches are so hard to get working, and so hard to debug when they're not working right; and for that matter, so hard to even figure out that they might not be working in the first place.

The original PR is still available here: https://github.com/virtualagc/virtualagc/compare/master...ColanderCombo:virtualagc:xcomi-ibm-float

The most impactful diff was changing the FR[] register set from a IEEE754 double to raw 64-bit IBM Hex values. We can group the diffs into three categories:

 - changes to internal MONITOR arithmetic
 - change to the "automatic" INLINE() instruction generation
 - changes to the 'patched' INLINE() locations

The MONITOR calls are isolatable, and adding MONITOR(9) back is no problem.  It drops a value into FR[0], but we already do a `FR[0] = ibm_dp_to_double(msw, lsw);` elsewhere.

The INLINE() calls are the bit that probably has to all change at once.  To get any benefit from the conversion we need FR[] to be IBM hex, uint64_t, and if we do that any place that touches FR needs to be updated.

rburkey2005:
> Could you drop a copy of the converted py file? I can take a look at it. Sometimes AI does a great job (especially on mechanical things like this), but I still like to examine it carefully before using it. :)

Well, I haven't yet looked at it beyond slapping a header at the top and at the end a skeleton of a main program that I copied from ibmHex.py but haven't modified yet, so I didn't intend to imply that it was just going to be used unquestioned.  But I'll admit that if tested thoroughly with a large dataset, I'm not going to quibble too much about the internals.

[ibmFloat.py](https://github.com/user-attachments/files/27456955/ibmFloat.py)

rburkey2005:
> The original PR is still available here: [master...ColanderCombo:virtualagc:xcomi-ibm-float](https://github.com/virtualagc/virtualagc/compare/master...ColanderCombo:virtualagc:xcomi-ibm-float)

I suppose I overreacted.  I like very much that my unnormalized-float hack has been fixed.  It goes without saying, but I suppose that with your original pull reguest you rebuilt the compiler and did a number of trial compilations?

I guess, considering that I'm the one who has all of the PASS source code and what few BFS files there are, I should have done test compilations in the first place rather than complain.  I could then directly compare the results gotten from the "old" compiler vs the "new", and should see only differences attributed to literals or other compile-time  expression evaluations.

Okay, I'll think for a bit about how to go about doing that.

rburkey2005:
> I guess, considering that I'm the one who has all of the PASS source code and what few BFS files there are, I should have done test compilations in the first place rather than complain. I could then directly compare the results gotten from the "old" compiler vs the "new", and should see only differences attributed to literals or other compile-time expression evaluations.
> 
> Okay, I'll think for a bit about how to go about doing that.

I admit that my understanding of github is primitive, but it doesn't look like I can fetch the original PR to my local machine since it doesn't have a PR number any longer.  However, I can apparently create a new pull request from it, and then fetch that new pull request.  Does that sound right to you?

ColanderCombo:
> I admit that my understanding of github is primitive, but it doesn't look like I can fetch the original PR to my local machine since it doesn't have a PR number any longer. However, I can apparently create a new pull request from it, and then fetch that new pull request. Does that sound right to you?  

Yeah, I think that would work.. I'm not on my machine with a tree right now, but I also believe you can add my branch as a 'remote' and then check out the branch:

```
git fetch https://github.com/ColanderCombo/virtualagc.git xcomi-ibm-float
git checkout -b xcomi-ibm-float FETCH_HEAD
```

ColanderCombo:
> I suppose I overreacted.

Nah, from my perspective it was reasonable caution.  the MONITOR() changes are more self contained.  The INLINE() stuff is spooky.

>  I like very much that my unnormalized-float hack has been fixed. It goes without saying, but I suppose that with your original pull reguest you rebuilt the compiler and did a number of trial compilations?

Yeah, I compile every test .hal file (e.g., from the Programming Manual, BENCH, etc.) and a number of additional unit/regression tests here:

https://github.com/ColanderCombo/nsts-sdl-dps/tree/master/test

I don't currently directly compare the compilation (e.g. by looking at the PASS2.rpt or obj output), but every one is executed using the gpc emulator and the outputs compared against baselines.  The only mismatches I saw were the expected floating point changes.

It sounds like a comparison test between the XPL and python HAL/S-FC implementations would be useful; I'll add that.  Having a large suite of automated tests has made my life much easier.  I'm happy to add whatever you think might be handy!

> I guess, considering that I'm the one who has all of the PASS source code and what few BFS files there are, I should have done test compilations in the first place rather than complain. I could then directly compare the results gotten from the "old" compiler vs the "new", and should see only differences attributed to literals or other compile-time expression evaluations.

I've been using cmake to scan for source, run the compile, run the gpc sim and then compare against baselines.  It would need some adaptation for your case, but it might be informative:

https://github.com/ColanderCombo/nsts-sdl-dps/blob/master/cmake/TestHalProgram.cmake

ColanderCombo:
@rburkey2005 I've updated this PR with a commit to port in the MONITOR(9) work from the other branch.

rburkey2005:
I got distracted evaluating the original PR (now PR 1303), so only now have I gotten back to trying to evaluate ibmFloat.py vs ibmFloat.c.  I haven't yet created the test rig for ibmFloat.c, but I have done so for ibmFloat.py, and [I've added ibmFloat.py to the repo under the XCOM-I directory](https://github.com/virtualagc/virtualagc/blob/master/XCOM-I/ibmFloat.py).  An identical stand-alone mode for ibmFloat.c, plus a large dataset for cross-comparison, will be the next step.  Rather than describe the various tests that can be performed, I'll just note the `--help` screen, with a few things <b>bolded</b> to draw special attention:
<pre>
rburkey@ron-1T XCOM-I: ./ibmFloat.py --help

Utility for exercising the ibmFloat Python module Usage:

	ibmFloat.py arg1 arg2 arg3 ...

The arguments can take any of the forms listed below.

--help
	Prints this message and exits.
HEX,HEX
	Converts an IBM DP floating-point number, expressed
	as a pair of comma-delimited 8-digit hexadecimals, to a
	human-readable, floating-point number.
NUMBER
NUMBER+NUMBER
<b>NUMBERsNUMBER</b>
NUMBER*NUMBER
NUMBER/NUMBER
	Any integer or floating-point number or simple expression
	is converted to an IBM DP floating-point number.  <b>Note
	that NUMBERsNUMBER is used in place of NUMBER-NUMBER</b>,
	which is not accepted.  The reason for that is that in
	HAL/S, -NUMBER is not a numeric literal, but rather an
	expression in which the operator '-' operates on NUMBER,
	therefore a HAL/S parser could not accept expressions like
	-3+12 or 5+-2.  Using 's' in place of '-' is a workaround
	for that.
<b>NUMBERpNUMBER</b>
<b>NUMBERmNUMBER</b>
	Same as number+number and number-number, except that the
	result is not normalized.
hcNUMBER
chNUMBER
	Convert HAL/S floating-point to native C/Python floating
	point or vice versa, respectively.
</pre>
I have noted a few peculiarities in ibmFloat.py, though as to whether they're present in ibmFloat.c I don't know yet.  At any rate, here's the list:

- In `ibm_dp_to_string` and `ibm_dp_to_hal_string`, 0.0 is truncated to " 0.0" rather than being padded to the correct number of spaces (i.e., matching the widths of things like " N.NNN...E+NN").  I've fixed that in both functions.
- `ibm_dp_from_string` does not detect illegal strings, and merely transparently bails under some circumstances, leaving the impression that the conversion was successful.  For example, feeding it the string "123hello4" returns 123.  Whether or not this has any actual bad result in XCOM-I or HAL_S_FC, I don't know yet.  However, it can lead to misleading results during testing, if you get confused and try to evaluate expressions like "123-4" rather than "123s4".  I've not made any attempt to fix it.

Here's a sample run:
<pre>
rburkey@ron-1T XCOM-I: ./ibmFloat.py 41619999,99999999 0 2+3 2s3 2*3 2/3 2p3 2m3 1.0001s1 1.0001m1 hc12345.67890 ch12345.67890
41619999,99999999   <->   DP=' 6.099999999999999E+00'   SP=' 6.099999E+00'   (41619999,99999999)
00000000,00000000   <->   DP=' 0.0                  '   SP=' 0.0         '   (0)
41500000,00000000   <->   DP=' 5.000000000000000E+00'   SP=' 5.000000E+00'   (2+3)
C1100000,00000000   <->   DP='-1.000000000000000E+00'   SP='-1.000000E+00'   (2s3)
41600000,00000000   <->   DP=' 6.000000000000000E+00'   SP=' 6.000000E+00'   (2*3)
40AAAAAA,AAAAAAAA   <->   DP=' 6.666666666666666E-01'   SP=' 6.666666E-01'   (2/3)
41500000,00000000   <->   DP=' 5.000000000000000E+00'   SP=' 5.000000E+00'   (2p3)
C1100000,00000000   <->   DP='-1.000000000000000E+00'   SP='-1.000000E+00'   (2m3)
3D68DB8B,AC710000   <->   DP=' 9.999999999998898E-05'   SP=' 9.999999E-05'   (1.0001s1)
41000068,DB8BAC71   <->   DP=' 9.999999999998897E-05'   SP=' 9.999999E-05'   (1.0001m1)
12345.678899999999   (hc12345.67890)
443039AD,CC63F142   <->   DP=' 1.234567889999999E+04'   SP=' 1.234567E+04'   (ch12345.67890)
</pre>

rburkey2005:
@ColanderCombo The test rig for ibmFloat.c turned out to be easier than I supposed.  I just ran ibmFloat.py back through CodeConvert, this time in the Python-to-C direction, pulled out the main program, and made a couple of minor corrections, calling it ibmFloat.Rig.c.  

I did have to push ibmFloat.c and ibmFloat.h, along with ibmFloatRig.c, into XCOM-I in Master.  And I did have to fix `ibm_dp_to_string` and `ibm_dp_to_hal_string` in ibmFloat.c.  So I quess that PR 1302 is basically wrecked now, and your ibmFloat.c is out of sync with mine.  :frowning_face: 

With the example usage I gave earlier, the C version produces identical results to the Python version, so I guess it's on to creating a bigger test dataset now.

rburkey2005:
@ColanderCombo I've added a feature to ibmFloat.py, `--make-test`, which outputs a file of test data that exercises every feature of ibmFloat to some extent.  The test data is then fed into both ibmFloat.py and ibmFloat.c+ibmFloatRig.c, and the output files of those two runs are compared.  Out of about 3800 cases tested, I find a single discrepancy (well, _two_ or _four_, depending on your outlook), namely converting 1E-6 (and -1E-6) to a HAL/S DP/SP.  In ibmFloat.py,
<pre>
3C10C6F7,A0B5ED8E   <->   DP=' 1.000000000000000E-06'   SP=' 1.000000E-06'   (ch1e-06)
BC10C6F7,A0B5ED8E   <->   DP='-1.000000000000000E-06'   SP='-1.000000E-06'   (ch-1e-06)
</pre>
whereas in ibmFloat.c+ibmFloatRig.c,
<pre>
3C10C6F7,A0B5ED8D   <->   DP=' 9.999999999999999E-07'   SP=' 9.999999E-07'   (ch1e-06)
BC10C6F7,A0B5ED8D   <->   DP='-9.999999999999999E-07'   SP='-9.999999E-07'   (ch-1e-06)
</pre>
This particular test uses native C or Python capability to convert the string "1e-06' to IEEE 754, then uses `ibm_dp_from_double` to get the results printed above.  I've added a feature to optionally print the IEEE 754 value in hexadecimal form, and have verified in both cases (i.e., in ibmFloat.py and ibmFloat.c+ibmFloatRig.c) that identical values of 3EB0C6F7A0B5ED8D are produced, so this is _not_ a problem of how the string "1e-06" is converted to IEEE 754.  (Interestingly, the IEEE 754 hex is very, very close to the IBM hex, which I hadn't expected to see.)

Comparing the code for `ibm_dp_from_double` in ibmFloat.py vs ibmFloat.c, there is nothing that seems different.  The sole thing I can point to, and which I conclude must be the difference, is that `llround(d)` in C must not be _quite_ identical to `int(d+0.5)` in Python.  But at the moment I don't know how it might be different or how to fix it if it is.

Any thoughts on that would be welcome.  Or if you want to look at the test cases generated by `--make-test` and suggest improvements, more test cases are always nicer than less test cases.

rburkey2005:
I've added debugging statements to ibmFloat.py and ibmFloat.c at each step in `ibm_dp_from_double(1E-6)` that modifies the value of `d`.  These statements show the IEEE 754 representation of `d`.  They are are identical right up to `llround(d)` or `int(d+0.5)`, after which the IEEE 754 representations differ.  So the problem must be in the rounding.

rburkey2005:
@ColanderCombo When asked for suggestions for replacements for `llround(d)`, Google A.I. suggested that if using only positive numbers, then `(long long)(d+0.5)` would be a higher-performance alternative.  Whether it's higher-performance, I can't say, but I've inserted it as a replacment in `ibm_dp_from_double` in ibmFloat.c, and now ibmFloat.py and ibmFloat.c produce identical results with my test cases.  I've updated the copy if XCOM-I/ibmFloat.c in Master accordingly.

Here's my summary of the current status of the pending PR's:

- ibmFloat.[ch] are now in the repository in XCOM-I/ independently of this pull request (1302), and have been modified relative to the pull request.
- The only modifications ("0.0" padding and `llround`), I think, are necessary corrections.
- ibmFloat.py is also in Master in XCOM-I/, and with the current test cases is functionally identical to ibmFloat.c.
- PR 1302, <strike>because of several of the above factors, is no longer good as-is, and will need some alterations before the remaining useful parts can be merged</strike> had some conflicts due to the changes mentioned above, but I was able to resolve them and so it could be merged.
- PR 1303, unmodified, is now satisfying to me, but has conflicts with Master that will need to be taken care of somehow before merging.  I don't have permissions to resolve the conflicts myself.

Issue #177, opened 2016-10-07
Transcribe RADAR LEADIN ROUTINES (pp. 168-197)


thewonderidiot:
Ron has already done this, just generating this issue for tracking purposes.

PR #1303, opened 2026-05-06
Original xcomi ibm float

rburkey2005:
(Now superseded by PR #1304, and soon to be deleted.)

Temporary recreation of older pull request for the purpose of fetching it to a local machine.

### DIFF
--- a/XCOM-I/XCOM-I.py
+++ b/XCOM-I/XCOM-I.py
@@ -711,6 +711,7 @@ if targetLanguage == "C":
         except:
             pass
         for name in ["runtimeC.c", "runtimeC.h", "inline360.c", "inline360.h",
+                  "ibmFloat.c", "ibmFloat.h",
                   "debuggingAid.c", "Makefile.template"]:
             shutil.copy(basePath + name, outputFolder + os.sep + \
                         name.replace(".template", ""))
--- a/XCOM-I/guessINLINE.py
+++ b/XCOM-I/guessINLINE.py
@@ -454,23 +454,15 @@ def guessINLINE(scope, functionName, parameters, inlineCounter, errxitRef,
         thisLine.append(indent + "setCC();")
         thisLine.append(indent + "GR[%d] = (int32_t) scratch;" % R1)
     elif opcode == 0x20: # LPDR p. 18-17
-        thisLine.append(indent + "scratchd = fabs(FR[%d]);" % R2)
-        thisLine.append(indent + "setCCd();")
-        thisLine.append(indent + "FR[%d] = scratchd;" % R1)
+        thisLine.append(indent + "lpdr(%d, %d);" % (R1, R2))
     elif opcode == 0x28: # LDR p. 9-10
-        thisLine.append(indent + "FR[%d] = FR[%d];" % (R1, R2))
-        # Note: No CC effect.
+        thisLine.append(indent + "ldr(%d, %d);" % (R1, R2))
     elif opcode == 0x29: # CDR p. 18-10
-        thisLine.append(indent + "scratchd = FR[%d] - FR[%d];" % (R1, R2))
-        thisLine.append(indent + "setCCd();")
+        thisLine.append(indent + "cdr(%d, %d);" % (R1, R2))
     elif opcode == 0x2A: # ADR p. 18-8
-        thisLine.append(indent + "scratchd = FR[%d] + FR[%d];" % (R1, R2))
-        thisLine.append(indent + "setCCd();")
-        thisLine.append(indent + "FR[%d] = scratchd;" % R1)
+        thisLine.append(indent + "adr(%d, %d);" % (R1, R2))
     elif opcode == 0x2B: # SDR p. 18-23
-        thisLine.append(indent + "scratchd = FR[%d] - FR[%d];" % (R1, R2))
-        thisLine.append(indent + "setCCd();")
-        thisLine.append(indent + "FR[%d] = scratchd;" % R1)
+        thisLine.append(indent + "sdr(%d, %d);" % (R1, R2))
     elif opcode == 0x41: # LA p. 7-78
         thisLine.append(indent + "GR[%d] = address360B & 0xFFFFFF;" % R1)
     elif opcode == 0x43: # IC p. 7-76
@@ -513,33 +505,19 @@ def guessINLINE(scope, functionName, parameters, inlineCounter, errxitRef,
     elif opcode == 0x60: # STD p. 9-11
         thisLine.append(indent + "std(%d, address360B);" % R1)
     elif opcode == 0x68: # LD p. 9-10
-        thisLine.append(indent + \
-            "FR[%d] = fromFloatIBM(COREWORD(address360B), COREWORD(address360B + 4));"\
-            % R1)
+        thisLine.append(indent + "ld(%d, address360B);" % R1)
     elif opcode == 0x69: # CD p. 18-10
-        thisLine.append(indent + "scratchd = FR[%d];" % R1)
-        thisLine.append(indent + \
-              "scratchd -= fromFloatIBM(COREWORD(address360B), COREWORD(address360B + 4));")
-        thisLine.append(indent + "setCCd();")
+        thisLine.append(indent + "cd(%d, address360B);" % R1)
     elif opcode == 0x6A: # AD p. 18-8
-        thisLine.append(indent + "scratchd = FR[%d];" % R1)
-        thisLine.append(indent + \
-              "scratchd += fromFloatIBM(COREWORD(address360B), COREWORD(address360B + 4));")
-        thisLine.append(indent + "setCCd();")
-        thisLine.append(indent + "FR[%d] = scratchd;" % R1)
+        thisLine.append(indent + "ad(%d, address360B);" % R1)
     elif opcode == 0x6B: # SD p. 18-23
-        thisLine.append(indent + "scratchd = FR[%d];" % R1)
-        thisLine.append(indent + \
-              "scratchd -= fromFloatIBM(COREWORD(address360B), COREWORD(address360B + 4));")
-        thisLine.append(indent + "setCCd();")
-        thisLine.append(indent + "FR[%d] = scratchd;" % R1)
+        thisLine.append(indent + "sd(%d, address360B);" % R1)
     elif opcode == 0x6E: # AW p. 18-10
         thisLine.append(indent + "aw(%d, address360B);" % R1)
     elif opcode == 0x70: # STE p. 9-11
-        thisLine.append(indent + "toFloatIBM(&msw360, &lsw360, FR[%d]);" % R1)
-        thisLine.append(indent + "COREWORD2(address360B, msw360);")
+        thisLine.append(indent + "ste(%d, address360B);" % R1)
     elif opcode == 0x78: # LE p. 9-10
-        thisLine.append(indent + "FR[%d] = fromFloatIBM(COREWORD(address360B), 0);" % R1)
+        thisLine.append(indent + "le(%d, address360B);" % R1)
     elif opcode == 0xD2: # MVC p. 7-83
         thisLine.append(indent + "mvc(address360A, address360B, %d);" % L)
     elif opcode == 0xDC: # TR p. 7-131
--- a/XCOM-I/ibmFloat.c
+++ b/XCOM-I/ibmFloat.c
@@ -12,7 +12,8 @@
 
 #include <assert.h>
 #include <math.h>
-#include <stddef.h>  // NULL
+#include <stddef.h>
+#include <stdio.h>
 #include "ibmFloat.h"
 
 #define twoTo56 (1LL << IBM_DP_MANT_BITS)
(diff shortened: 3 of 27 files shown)


rburkey2005:
Notes to self:  The idea here is to run `compilePASS` to perform as complete a build as possible right now of OI340600, and to compare it to an earlier build made with the current (unpatched) version of the compiler.  The latter results had been archived along with compilePASS.log in OI340600/old/.  Several immediate difficulties related to recent changes in `compilePASS` were encountered and had to be worked around.

- PASS2.PROCS/OBJECTGE.xpl must be modified to add `/?V  DECLARE MAX_SEVERITY FIXED INITIAL(0); ?/` to `OBJECT_GENERATOR`.
- `--preprocess` has to be added to the command line when invoking `compilePASS`.
- For compatibility, `parmBase` needed to be changed to "SREF,LIST,LISTING2,SRN,TEMPLATE,NOLFXI,REGOPT,LITSTRINGS=3000,CARDTYPE=".
- `defaultR` had to be changed from "C" to "M" to prevent compilation from stalling at SPPPRECO.hal.
- Prior to running `compilePASS`, TEMPLIB and INCLIB had to be cleared and recreated.

The results of the test build (and compilePASS.log) were stored in the new/ directory.

rburkey2005:
With the modification mentioned in the previous comment, I find that at least as far as compilePASS.log is concerned, the builds proceed identically, finally stalling at CV5SLCOM.hal (item 782).

The next step is to compare corresponding files generated by the compilation in new/ vs old/.  This was made tricky by the fact that the results for source file XXXXXXXX.hal are stored in a subdirectory with a name that includes a timestamp, such as 'HALSFC XXXXXXXX.hal 2026-04-29 09-34-10.results'.  I facilitated the comparison by changing the names simply to 'XXXXXXXX.results', which seemed harmless, so that there are identically-named subdirectories in new/ vs old/.

Comparisons of file-pairs can mostly be performed by simple `diff`'s, except where timestamps or other variable but irrelevant data is embedded with the files themselves.

- pass1A.rpt:  I wouldn't expect to see differences here unless `DEBUG ¢E` were used, which causes a literal table to be embedded in the report, but which is not the case here.  Extraneous data was already removed when `HALSFC --test ...` converted pass1.rpt into pass1A.rpt.  Actual results of the comparisons:  100% match.
- halmat.bin:  All pairs matched except for GPBGPS.  (See next comment for discussion.)
- optmat.bin:  All pairs matched except for GPBGPS.
- litfile0.bin, COMMON0.out.bin.gz: I would expect scattered discrepancies in the litfiles, corresponding to the least-significant bytes of floating-point values.  The most-direct way to see these would seem to be using `unLitfile.py` to turn the files into human-readable form, and visually comparing every pair of files that were not 100% matches.  Actual results:  Out of 782 pairs total, a simple `diff` of litfile0.bin reveals that 472 pairs do not match.  Curiously, processing with `unLitfile.py` first reveals only 179 mismatches, presumably due to the limits of precision of the IBM Hexadecimal to IEEE floating point performed by `unLitfile.py`.  Unfortunately, `unLitfile.py` displays a human-readable form of floating-point literals rather than the IBM Hex form, so determining what's really a change in the least-significant bit(s) is quite difficult.  I've merely used the criterion that a "very small number" of mismatched floats occur and that the discrepancies are "very small", which are subjective judgements.  And rather than manually check all 179 pairs, I've only looked at a sampling of the mismatched pairs:
    1. CGA1MC:  Yes, very few, very low-significance.
    2. CGA2MC:  Same.
    3. CGA3MC:  Same.
    4. CGAGAX:  Same.
    5. CGAMC1:  Same.
    6. CGCUN1:  Same.
    7. CGEIPC:  Same.
    8. CGIGNC:  Same.
    9. CGMMC9:  Same.
    10. CGNCOM:  Same.
    11. CGSAS1:  Same.
    12. GADHFE:  Same.
    13. GC5ORB:  Same.
    14. GCHGRT:  Same.
    15. GG0THR:  Same.
    16. GGFAUT:  Same.
    17. GHLRTL:  Same.
    18. GLWRNP:  Same.
    19. GY7STC:  Same.
    20. SSCFUELC:  Same.
    21. VNECYCLI:  Same.
- cards.bin (object code):  These files have embedded-timestamp data in these files, and all 715 pairs do not match according to `diff`.  Instead, I used the technique of turning the files into human-readable form using `ibmobjdump`, eliminating timestamp and other extraneous messages using `egrep`, and then `diff`'ing that form of the data.  42 pairs of  715 total were found to differ.  There is only occasional overlap between the source files with litfile mismatches (see above) and the source files with object-code mismatches.
- pass2.rpt:  There's embedded timestamp info in these reports.  Using `egrep` to eliminate that info, and `diff`'ing, there are 42 mismatched pairs of  715 total.  Those 42 are same files identified in the cards.bin test.  Visually checking the `diff`'s in a sampling of the files, I find:
    1. ASGCYCLI:  Apparent differences in the least-significant byte of several doubleword constants.
    2. CGCUN1: Same.
    3. CGEIPC:  Same.
    4. CGGC01:  Same.
    5. CGGCOM:  Same.
    6. CGYMC3:  Same, but in some word constants rather than doubleword.
    7. GCKGCS:  Same (doubleword).
    8. GHRTAC:  Same (doubleword).
    9. SPSPSP:  Same (doubleword).
    10. VMELOAD:  Same (doubleword).

rburkey2005:
I'd evaluate the preceding comment as containing only expected findings, with the exception of the module GPBGPS.  To recap the GPBGPS finding:  The pass1.rpt report is identical to what it was before, but the halmat.bin file generated by pass 1 differs, therefore so does optmat.bin, cards.bin, and pass2.rpt.   What's up with that?

You can see the difference immediately in the discrepancies of the pass2.rpt.  Specifically, in the unpatched current build, in p. 44 of the report, you find
<pre>
000082A                             ST#3469  EQU    *
0082A 9A05           0001                    LH     R2,1(R1)            TIME: 0.25; BASE_REG#25
0082B 9EF6 E001      0001                    LH     R6,1(R7,R2)         TIME: 7.0
<b>0082D 9CF3 0590      0590                    LH     R4,1424()           TIME: 0.25; CZ2B_OFF_MASK+1</b>
0082F 26E4                                   NR     R6,R4               TIME: 0.25
00830 BEF6 E001      0001                    STH    R6,1(R7,R2)         TIME: 8.5
0000832                             ST#3470  EQU    *
<b>00832 9B15           0005                    LH     R3,5(R1)            TIME: 0.25; BASE_REG#32</b>
00833 9C0F           0003                    LH     R4,3(R3)            TIME: 0.25
00834 2EE4                                   OR     R6,R4               TIME: 0.25
</pre>
whereas in the build from this PR you find instead
<pre>
000082A                             ST#3469  EQU    *
0082A 9A05           0001                    LH     R2,1(R1)            TIME: 0.25; BASE_REG#25
0082B 9EF6 E001      0001                    LH     R6,1(R7,R2)         TIME: 7.0
<b>0082D 9B15           0005                    LH     R3,5(R1)            TIME: 0.25; BASE_REG#32</b>
<b>0082E 9C4B           0012                    LH     R4,18(R3)           TIME: 0.25</b>
0082F 26E4                                   NR     R6,R4               TIME: 0.25
00830 BEF6 E001      0001                    STH    R6,1(R7,R2)         TIME: 8.5
0000832                             ST#3470  EQU    *
00832 9C0F           0003                    LH     R4,3(R3)            TIME: 0.25
00833 2EE4                                   OR     R6,R4               TIME: 0.25
</pre>
The net effect of these differences is that in the PR build there's one less halfword, which pulls everything following them downward in memory by one halfword, but those seem to be the _only_ differences.

(It does appear to me that at 0082B, the unpatched build is loading the address of `CZ2B_OFF_MASK` into `R4`, and not `CZ2B_OFF_MASK+1`, whereas at 0082D-0082E the PR build is loading `CZ2B_OFF_MASK+1` into `R4`.  Perhaps I'm reading the report wrong.)

rburkey2005:
If instead we use `unHALMAT` to turn the pair of halmat.bin files into human-readable form, and `diff` those human-readable forms, we find that our troubles begin at HAL/S statement 3388, with the current build having
<pre>
  HALMAT #844 (0x007, Start of IF-statement): IFHD(0) 00 1 0
  HALMAT #845 (0x5C1, INTEGER-to-SCALAR conversion): ITOS(1) 00 0 0
  HALMAT #846	  | 1703 00 05(LIT) 00 00 00 01 00 00   (Literal #5891: Type FIXED, 2147483648.0)
  HALMAT #847 (0x7AA, Test SCALAR less than): SLT(2) 00 1 0
  HALMAT #848	  | D59 00 01(SYT) 00 00 00 01 00 00   (Symbol #3417: AD_BARO_CONV2, SCALAR DOUBLE TEMPORARY)
  HALMAT #849	  | 34D 00 03(VAC) 00 00 00 01 00 00   (Result from operation at HALMAT #845)
  HALMAT #850 (0x00A, Branch on false): FBRA(2) 00 1 0
  HALMAT #851	  | 0B1 00 02(INL) 00 00 00 01 00 00   (Bookkeeping label #177)
  HALMAT #852	  | 34F 00 03(VAC) 00 00 00 01 00 00   (Result from operation at HALMAT #847)
  HALMAT #853 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #854	  | D3C 00 00(  0) 00 00 01 01 00 01   (HAL/S statement #3388)

108500 3388 M| 4               IF AD_BARO_CONV2 < 2   THEN DO;                                                     |
</pre>
and the PR build having
<pre>
  HALMAT #844 (0x007, Start of IF-statement): IFHD(0) 00 1 0
  HALMAT #845 (0x7AA, Test SCALAR less than): SLT(2) 00 1 0
  HALMAT #846	  | D59 00 01(SYT) 00 00 00 01 00 00   (Symbol #3417: AD_BARO_CONV2, SCALAR DOUBLE TEMPORARY)
  HALMAT #847	  | 1703 00 05(LIT) 00 00 00 01 00 00   (Literal #5891: Type FIXED, 2147483648.0)
  HALMAT #848 (0x00A, Branch on false): FBRA(2) 00 1 0
  HALMAT #849	  | 0B1 00 02(INL) 00 00 00 01 00 00   (Bookkeeping label #177)
  HALMAT #850	  | 34D 00 03(VAC) 00 00 00 01 00 00   (Result from operation at HALMAT #845)
  HALMAT #851 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #852	  | D3C 00 00(  0) 00 00 01 01 00 01   (HAL/S statement #3388)

108500 3388 M| 4               IF AD_BARO_CONV2 < 2   THEN DO;                                                     |
</pre>
Note that the text of the HAL/S statement, "`IF AD_BARO_CONV2 < 2   THEN DO;`" is really a bug in `unHALMAT`, because the real statement is:
<pre>
            E|                                     31                                                              | 
108500 3388 M| 4               IF AD_BARO_CONV2 < 2   THEN DO;                                                     |DO=ST#3389
</pre>
Now, the legal range of `INTEGER DOUBLE` values is -2<sup>31</sup> through 2<sup>31</sup>-1, so the `INTEGER`-to-`SCALAR` conversion which the current build is performing on the literal 2147483648 _cannot_ be correct, since 2147483648 is a `SCALAR` rather than an `INTEGER` to begin with.  The patched build, in other words, has the _correct_ behavior, and the current build does not.

Here's a curious factoid:  HAL_S_FC.py's HALMAT (FILE1.bin) agrees with the current build, and compilePASS in the PR build doesn't notice that halmat.bin doesn't match FILE1.bin.  Or more accurately, failure of the cross-comparison between HALSFC-PASS1 and HAL_S_FC.py is not detected as a failure condition by compilePASS, and it only prints the results of the cross-comparison upon failure, so the messages never show up in the log.  That sounds like a problem that may require a fix.  But at any rate, the discrepancy between HALSFC-PASS1 and HAL_S_FC.py is still something that needs fixing, whether or not reported by compilePASS.

<strike>Question:  Should I continue trying to figure the exact manner in which the current build and the PR build produced differing results for GPBGPS, or should I just accept the PR?  I think I should continue.</strike>

rburkey2005:
@ColanderCombo After a lot of messing around with it, I guess I have no qualms about merging the original PR after all.  I found goofy stuff, not in the PR itself, I think, but goofiness in what I was already doing but hadn't yet detected.  At any rate, I don't find it breaking anything.  I don't want to just go ahead and do it without feedback, though, since I don't necessarily know how it affects your scaled-down PR.  I only tested the original PR, and didn't try to pull in any of your scaled-down PR, even though you said you thought I could do it.

rburkey2005:
@ColanderCombo As near as I can tell by looking, it appears to me that if I merge PR #1303 (your original PR, as recreated by me, but otherwise unchanged), there will be conflicts if I were to try to merge PR #1302 after that, specifically in XCOM-I/runtimeC.c.  I admit that it's hard to tell just by looking at the automated comparison.  I'm thinking that I should merge #1303, but that #1302 would be junk after that and that you'd have to do something in its place.  Should I go ahead with #1303?

Or perhaps it would be easier if both #1302 and #1303 were discarded and you just made a new PR with all of your changes in it, perhaps pulling the latest from master before doing so.

rburkey2005:
@ColanderCombo I've gone ahead and merged PR 1302, after resolving some conflicts.  That leaves only 1303, which has merge conflicts but for some reason I don't have permissions to resolve them.  (Literally, "You do not have permission to push to the head branch".)  It's puzzling.

PR #32, opened 2016-03-16
Updated readme

shanecoughlan:
Added context and detailed build instructions from the project main
website, cleaning up language as I went.

Issue #29, opened 2015-04-07
agc_engine_init return value wrong


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Call the function:
int res = agc_engine_init(&state, "filethatdoesnotexist", NULL, 0);

2. res should be 1, according to:
http://www.ibiblio.org/apollo/developer.html#AGC_CPU-Engine_API

3. BUT: res is equal 0 (value is overwritten in agc_engine_init.c:210  (see 
also line 193 for ROM-file-load assignment)

What version of the product are you using? On what operating system?
Last change in header: 03/30/09 RSB	Added the Downlink variable to the core 
dumps.

yaAGC-dev-20100220.tar.bz2
(same as subversion rev 609)

Also found in:
https://code.google.com/p/virtualagc/source/browse/trunk/yaAGC/agc_engine_init.c
```

Original issue reported on code.google.com by `flo.reut...@gmail.com` on 24 Sep 2014 at 8:18

ohommes:
I see the overwrite of the value at line agc_engine_init.c:210 but I believe the reason why the return value of agc_load_binfille is ignored is because it could never return 0 as success the way it is written. I will fix agc_load_binfile and then fix the agc_engine_init to start using the return value of the agc_load_binfile.

ohommes:
I have fixed this issue on Aug. 14 2016 per previous comment. The agc_engine_init now returns the correct values.

Issue #28, opened 2015-04-07
Floating point exception in yaAGC.


GoogleCodeExporter:
```
Hello, I'm zxMarce, and I have a problem I think you may help me overcome.
I installed VirtualAGC version 20100220 under GNU Ubuntu Linux 12.04.1 in my 
home folder. I made two installs on two different machines, one is a physical 
laptop (2GB RAM), the other is a VirtualBox virtual machine (768MB RAM).

Both GUIs fail with the same symptoms: No matter the modules/modes chosen, when 
the Run button is clicked, the selected support modules (DSKY, Status, 
Telemetry, etc) start up and appear for 3-5 seconds and then everything shuts 
down and I get the main Virtual AGC screen back to select other emulation 
parameters.

When resorting to the terminal, it seems yaAGC has a problem. In order to see 
if the Luminary binaries are to be blamed I tried running with both binaries 
supplied (099 and 131) and even a fake, non-existing version 999 (not even the 
directory for Luminary 999 exists in my environment).

I get the following "Floating point exception (core dumped)", on all three 
Luminary versions I tried (virtual machine tests shown, physical laptop throws 
the same exceptions):

user@vmachine:~/VirtualAGC/Resources$ ../bin/yaAGC 
--core="source/Luminary131/Luminary131.bin" --port=19797 --cfg=LM.ini
Apollo Guidance Computer simulation, ver. 20100220, built Feb 20 2010 11:03:52
Copyright (C) 2003-2009 Ronald S. Burkey, Onno Hommes.
yaAGC is free software, covered by the GNU General Public License, and you are
welcome to change it and/or distribute copies of it under certain conditions.
Refer to http://www.ibiblio.org/apollo/index.html for additional information.
Floating point exception (core dumped)

user@vmachine:~/VirtualAGC/Resources$ ../bin/yaAGC 
--core="source/Luminary099/Luminary099.bin" --port=19797 --cfg=LM.ini
Apollo Guidance Computer simulation, ver. 20100220, built Feb 20 2010 11:03:52
Copyright (C) 2003-2009 Ronald S. Burkey, Onno Hommes.
yaAGC is free software, covered by the GNU General Public License, and you are
welcome to change it and/or distribute copies of it under certain conditions.
Refer to http://www.ibiblio.org/apollo/index.html for additional information.
Floating point exception (core dumped)

user@vmachine:~/VirtualAGC/Resources$ ../bin/yaAGC 
[...]

PR #1304, opened 2026-05-10
Replacement xcomi ibm float

rburkey2005:
Updates XCOM-I for IBM hex computations.

@ColanderCombo Since I didn't have permissions to resolve the merge conflicts directly online in PR #1303, I did it locally and created this replacement PR.  If it's possible to confirm that this PR changes the files you wanted changed in the way you wanted them changed, and if it passes whatever tests I can throw at it, then we can forget PR #1303 and merge this one instead.


rburkey2005:
A problem appears.  Consider the following HAL/S sample program (quoted from its PASS1 compilation report rather than its source code):
<pre>
   1 M| TEST:                                                                                               |TEST
   1 M| PROGRAM;                                                                                            |TEST
   2 M|   DECLARE N1 INTEGER INITIAL(.625 6400);                                                            |TEST
   3 M|   <b>DECLARE N2 INTEGER INITIAL(.425 6400);</b>                                                            |TEST
   4 M|   WRITE(6) N1, N2;                                                                                  |TEST
   5 M| CLOSE TEST;                                                                                         |TEST
</pre>
When compiled, this seemingly succeeds in PASS1 (ignoring issue #1305, which doesn't cause us any problem here), then succeeds in OPT, but ultimately fails PASS2 with the error:
<pre>
***** BS108    ERROR #1 OF SEVERITY 2 OCCURRED ***** 
*****  DURING CONVERSION OF <b>HAL/S STATEMENT 3</b>.***** 
***** LITERAL PROCESSING FAILURE                                                      
***** CONVERSION ABANDONED
</pre>

The problem is that `.425 6400`, as computed in IBM hexadecimal floating point, is not in fact a pure integer, whereas `.625 6400` is.  Here's a demonstration of that assertion using the ibmFloat program (in either its C or Python forms) to tell use how these various computations are actually carried out in IBM hex format:
<pre>
> ibmFloat.py .625*6400 <b>.425*6400</b> .425
43FA0000,00000000   <->   DP=' 4.000000000000000E+03'   SP=' 4.000000E+03'   (.625*6400)
<b>43A9FFFF,FFFFFFFE</b>   <->   DP=' 2.719999999999999E+03'   SP=' 2.719999E+03'   (.425*6400)
406CCCCC,CCCCCCCC   <->   DP=' 4.249999999999999E-01'   SP=' 4.249999E-01'   (.425)
</pre>
Where in the compilation chain does this problem occur?  It occurs already in PASS1, even though not recognized until PASS2.  Here's the full HALMAT produced by PASS1, as parsed by the program unHALMAT.py, with the relevant part **bolded**, which communicates the gist of the problem:
<pre>
  HALMAT #0 (0x005, Start of HALMAT block): PXRC(1) 00 0 0
  HALMAT #1	  | 020 00 00(  0) 00 00 00 01 00 00   (Matching XREC is at HALMAT #32)
  HALMAT #2 (0x02B, Start of PROGRAM definition): MDEF(1) 00 0 0
  HALMAT #3	  | 001 00 01(SYT) 00 00 00 01 00 00   (Symbol #1: TEST, PROGRAM LABEL)
  HALMAT #4 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #5	  | 001 00 00(  0) 00 00 01 01 00 01   (HAL/S statement #1)
  HALMAT #6 (0x8C1, INTEGER INITIAL value): IINT(2) 05 0 0
  HALMAT #7	  | 002 00 01(SYT) 00 00 00 01 00 00   <b>(Symbol #2: N1, INTEGER SINGLE ALIGNED STATIC INITIAL)</b>
  HALMAT #8	  | 003 00 05(LIT) 00 00 00 01 00 00   <b>(Literal #3: Type FIXED, 4000.0)</b>
  HALMAT #9 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #10	  | 002 00 00(  0) 00 00 00 01 00 00   (HAL/S statement #2)
  HALMAT #11 (0x8C1, INTEGER INITIAL value): IINT(2) 05 0 0
  HALMAT #12	  | 003 00 01(SYT) 00 00 00 01 00 00   <b>(Symbol #3: N2, INTEGER SINGLE ALIGNED STATIC INITIAL)</b>
  HALMAT #13	  | 006 00 05(LIT) 00 00 00 01 00 00   <b>(Literal #6: Type FIXED, -0.0)</b>
  HALMAT #14 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #15	  | 003 00 00(  0) 00 00 00 01 00 00   (HAL/S statement #3)
  HALMAT #16 (0x031, End of DECLAREs): EDCL(0) 01 1 0
  HALMAT #17 (0x025, Start of argument list): XXST(1) 00 1 0
  HALMAT #18	  | 002 00 06(IMD) 00 00 00 01 00 00   (Immediate integer data 2 0x2)
  HALMAT #19 (0x027, Argument): XXAR(1) 00 1 0
  HALMAT #20	  | 002 06 01(SYT) 00 00 00 01 00 00   (Symbol #2: N1, INTEGER SINGLE ALIGNED STATIC INITIAL)
  HALMAT #21 (0x027, Argument): XXAR(1) 00 1 0
  HALMAT #22	  | 003 06 01(SYT) 00 00 00 01 00 00   (Symbol #3: N2, INTEGER SINGLE ALIGNED STATIC INITIAL)
  HALMAT #23 (0x021, WRITE statement): WRIT(1) 00 0 0
  HALMAT #24	  | 006 00 06(IMD) 00 00 00 01 00 00   (Immediate integer data 6 0x6)
  HALMAT #25 (0x026, End of argument list): XXND(0) 00 0 0
  HALMAT #26 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #27	  | 004 00 00(  0) 00 00 01 01 00 01   (HAL/S statement #4)
  HALMAT #28 (0x030): CLOS(1) 00 0 0
  HALMAT #29	  | 001 00 01(SYT) 00 00 00 01 00 00   (Symbol #1: TEST, PROGRAM LABEL)
  HALMAT #30 (0x004, HAL/S statement marker): SMRK(1) 00 1 0
  HALMAT #31	  | 005 00 00(  0) 00 00 01 01 00 01   (HAL/S statement #5)
  HALMAT #32 (0x002, End of HALMAT block): XREC final
</pre>
As you can see, the `INITIAL` value for `N1` is correctly recognized as having the value 4000.0, while the `INITIAL` value for `N2` is found to be -0.0.  (**Note**:  The specific values listed are indirect references to the literal file, and are not embedded within the HALMAT itself, which is why `HALSFC --test ...` does not detect any difference in the HALMAT files produced by HALSFC-PASS1 and HAL_S_FC.py, and any differences in the literal files they produce are obscured by issue #1305.)

The problem seems to be traceable to the HAL/S-FC compiler's `PREP_LITERAL` function, which when translated from XPL+BAL (with "BAL" being System/360 Basic Assembly Language) to C by the XPL/I compiler, becomes PREPuLITERAL.c+patch57.c+patch59.c+patch76.c.  Both patch57.c and patch59.c are changed by the PR, but in particular, it is patch59.c which tries to determine whether or not the literal being prepared is an "exact" integer or not, so I'd infer that the problem is in patch59.c.

Here's the XPL/I source code for `PREP_LITERAL`, for readability rather than in the C translation, _sans_ comments:
<pre>
PREP_LITERAL:                                                                   00671800
   PROCEDURE;                                                                   00671900
      DECLARE TEMP1 FIXED;                                                      00672000
      IF EXP_OVERFLOW THEN DO;                                                  00672100
         CALL INLINE("58",1,0,ADDR_FIXED_LIMIT); /* L 1,ADDR_FIXED_LIMIT*/      00672200
         CALL INLINE("68",6,0,1,0);       /* LD 6,0(0,1)   */                   00672300
NOT_EXACT:                                                                      00672400
         VALUE = -1;                                                            00672500
         TOKEN = CPD_NUMBER;                                                    00672600
         GO TO SAVE_NUMBER;                                                     00672700
      END;                                                                      00672800
      TEMP1 = ADDR(NOT_EXACT);                                                  00672900
      CALL INLINE("58", 1, 0, ADDR_FIXED_LIMIT);       /* L   1,ADDR_LIMIT */   00673000
      CALL INLINE("58", 2, 0, TEMP1);                  /* L   2,TEMP1      */   00673100
      CALL INLINE("28", 6, 0);                         /* LDR 6,0          */   00673200
      CALL INLINE("20", 0, 0);                         /* LPDR 0,0         */   00673300
      CALL INLINE("69", 0, 0, 1, 0);                   /* CD  0,0(,1)      */   00673400
      <b>CALL INLINE("07", 2, 2);                         /* BHR 2            */   00673500</b>
      CALL INLINE("2B", 4, 4);                         /* SDR 4,4          */   00673600
      CALL INLINE("28", 2, 0);                         /* LDR 2,0          */   00673700
      CALL INLINE("58", 1, 0, ADDR_FIXER);             /* L   1,ADDR_FIXER */   00673800
      CALL INLINE("6E", 0, 0, 1, 0);                   /* AW  0,0(,1)      */   00673900
      CALL INLINE("58",1,0,TABLE_ADDR);                                         00674000
      CALL INLINE("60", 0, 0, 1, 0);                   /* STD 0,0(,1)      */   00674100
      CALL INLINE("2A", 0, 4);                         /* ADR 0,4          */   00674200
      CALL INLINE("2B", 2, 0);                         /* SDR 2,0          */   00674300
      <b>CALL INLINE("07", 7, 2);                         /* BNER 2           */   00674400</b>
      CALL INLINE("58", 2, 0, 1, 4);                   /* L   2,4(,1)      */   00674500
      CALL INLINE("50", 2, 0, VALUE);                  /* ST  2,VALUE      */   00674600
SAVE_NUMBER:                                                                    00674700
      CALL INLINE("58",1,0,TABLE_ADDR);                                         00674800
      CALL INLINE("60",6,0,1,0);          /* STD 6,0(0,1)   */                  00674900
      SYT_INDEX=SAVE_LITERAL(1,TABLE_ADDR);                                     00675000
   END PREP_LITERAL;                                                            00675100
</pre>
The parts I've **bolded** are error branches to the label `NOT_EXACT` when the value being tested is found not to be representable as an integer, and the problem must occur in the lead-ups to these branches.  The branch instructions are tests of the CPU's `CC` register, which holds the status of preceding arithmetical or comparison operations, and branches are to the address held in CPU general-register `R2`, which has previously been loaded with the address of the label `NOT_EXACT`.

The lead-up to the first branch (SRN 00673500) compares the test value (which is an absolute value stored in floating-point register `FR0`) with the maximum positive integer, 2<sup>31</sup>-1.  Note that the comparison is entirely in DP floating point, and hence is not a boundary case as it would be if it were an integer calculation, so it is unlikely to be in error.  Not impossible, just unlikely.  So let's look beyond that.

The lead-up to the second branch (SRN 00674400) is a kind of trick calculation that could only work in IBM hex floating point, in which the value being tested is combined with a special value, the "fixer" (IBM hex value 4E000000,00000000, or 0&times;16<sup>14</sup>), which technically has the value 0.0 though not encoded the way 0.0 normally would be (as 00000000,00000000, or 0&times;16<sup>0</sup>), but has the special power of converting floats to integers.  So that seems suspicious.  Let's step through the code leading up to the 2nd branch:

- `SDR 4,4`:  Subtracts float-register `FR4` from itself, leaving it loaded with 0.0.
- `LDR 2,0`:  Loads float-register `FR0` (currently the absolute value of the float being tested) into float-register `FR2`.
- `L 1,ADDR_FIXER`:  Loads general register `R1` with the address of the fixer constant.
- `AW 0,0(,1)`:  Adds the fixer constant (0&times;16<sup>14</sup>), _unnormalized_, to the value in float-register `FR0`.  Not being "normalized" means that prior to the addition, the operand with the smaller exponent is going to have its mantissa shifted rightward, at the same time having its exponent increased, until the two operands have the same exponent; and then after the addition, the mantissa is left in that unnaturally-shifted position.  So if the value being tested is &ge;16<sup>14</sup>=2<sup>56</sup> it will remain unchanged, whereas if it is &lt;2<sup>56</sup> it will be shifted downward until its least-significant bit is aligned with the least-significant bit of the entire doubleword.  The exponent can then be discarded and the doubleword treated as a very large integer. 
- _etc_.

So ... does the `AW` operation in this PR treat this special case properly?  Answer:  It does not; rather, it simply treats the fixer as if it were like any other 0.  The problem is not in `AW` itself, but rather in the call it makes to `ibm_dp_addsub` in ibmFloat.c or ibmFloat.py.

**Fix**:  Fixes for ibmFloat.c and .py pushed into the TestPR1303 branch.  Lots of fixer test cases were added for cross tests of ibmFloat.py vs ibmFloat.c.  A `--test-fixer` mode was also added (to both ibmFloat.c and .py) for testing computations with the fixer internally.

rburkey2005:
Alas, this does not fix the problem described at the top of the preceding comment, namely that with the sample HAL/S program given there, compilation fails at PASS2.  This continues to happen! The error reported is identical, and the diagnosis via HALMAT that the problem was in `PREP_LITERAL` remains the same.  There seem to be two possibilities:

1. The problem occurs _after_ the `AW` instruction (in which the fixes described above occur).  Or else,
2. The fixes so far for ibmFloat, which are undoubtedly an improvement, still do not capture the full effect of unnormalized computations involving unnormalized zeroes.

Let's start with the first, and resume plodding through the BAL instructions in `PREP_LITERAL`:

- SRN 00674000, `L 1,TABLE_ADDR`:  Loads the address of `DW(6),DW(7)` in the floating-point work-area into general register `R1`.  I'm at a loss as to why it might want to do that, since I cannot discern that values stored in `DW(6),DW(7)` are used for anything, anywhere.
- `STD 0,0(,1)`:  Stores the value from float-register `FR0` (which, recall, contains the result of the earlier `AW` instruction) into `DW(6),DW(7)`.
- `ADR 0,4`:  Adds float-register `FR4` (which, recall, still contains 0.0) to float-register `FR0`.  The result is to normalize `FR0`.
- `SDR 2,0`:  Subtract float-register `FR0` from float-register `FR2`, which (recall) contains the absolute value of the value being tested.  The condition codes in the `CC` register will be set at this point to indicate whether the now-normalize value computed by `AW` is equal to, greater than, or less than the absolute test value.  In our problematic case, 0.425*6400, the test value is 2719.999..., the value computed by `AW` was 2719, and thus the result left in `FR2` should be 0.999..., which is positive.
- `BNER 7,2`:  Branch to the address held in general register `R2` (recall, the label `NOT_EXACT`) if the `CC` register indicates a result that's either positive or negative (no branch if zero).  Consequently, the branch to `NOT_EXACT` is taken.

It simply cannot be intended to work like this, since if it did, the calculations being performed would be nonsense without any hope of detecting whether the value being tested was an integer or not.  I would infer that _both_ of the potential types of errors postulated earlier are occurring:

1. `AW` via `ibm_dp_addsub` is still producing an incorrect result, in that it should produce 2720 in our problematic case rather than 2719; and
2. The method used by `SDR` to set the zero-detection bit in the `CC` register is wrong, because it requires some tolerance, which I would imagine involves ignoring differences in the least-significant nibble ("guard digit") of the result.  I would suppose that this hypothetical incorrect treatment extends to other BAL instructions we're imitating as well, such as `ADR`, though whether any of them is as critical to the XPL/I compiler as `SDR` is questionable.

To illustrate what I mean, if `ibm_dp_addsub` worked as I say it should have, then `AW` would have left unnormalized 2720, namely 4E000000,00000AA0, in float-register `FR0`.  After normalization by `ADR 0,4`, this would have become 43AA0000,00000000.   All along in float-register `FR2` would have been 2.17199...E+03, namely 43A9FFFF,FFFFFFFE.  What is the difference between these two?  Subtracting them _unnormalized_ we find the IBM hex result C3000000,00000002.  I.e., the difference is only in the guard digit, and if that had been ignored by `SDR` in setting `CC`, the result would have been found to be 0, and the `BNER 7,2` branch would not have been taken.

I'd note that in the ESA/390 Principles of Operation description of `AW`, it specifically mentions that "the  guard digit does not participate in the recognition of a zero result fraction. A zero result fraction is recognized when the fraction (that is, the intermediate-sum fraction, excluding the guard digit) is zero".  This comment was intended specifically to contrast to `AD` instruction (the normalized counterpart to `AW`), in which the guard digit does participate in determination of the `CC` bits.  Which doesn't immediately help us, but does indicate that in saying that `CC` bit 0 means the result is "zero", that doesn't necessarily mean that it's literally 0.  Whereas in ibmFloat, we invariably find that `CC==0` _does_ literally mean zero.  So the way ibmFloat is setting `CC` is not entirely correct.

rburkey2005:
Regarding the two notions proposed in the preceding comment:

1. I've made a change in `ibm_dp_addsub` in both ibmFloat.c and .py in which in the unnormalized case _only_, the mantissa is rounded on the basis of the guard digit:  rounding up if the guard is &ge;8, and rounding down if the guard is &lt;8.
2. Regarding the setting of the `CC` register to "result is zero" on the basis of a preceding arithmetical operation, ibmFloat is structured wrong to provide a tolerance based on the guard digit:  The code assumes that an arithmetical operation can be performed, and then independently, the `CC` register can be be set entirely on the basis of the resulting value itself.  Information such as the guard digit or even the original operands that could provide a tolerance have been discarded prior to that, and are no longer available for decision making.  Whereas according to my assertions, `CC` should instead be set _during_ the arithmetical operation while intermediate data like the guard digit is still available.  That means that a "fix" to ibmFloat, if advisable, would mean restructuring.

Since fixing how ibmFloat deals with the `CC` register seems not to be feasible &mdash; and since honestly, it's a big effort based on very little evidence &mdash;, I'm not going to attempt that.  I think it's wrong, but it will just have to stay wrong.

Instead, I notice that if the developer who wrote `PREP_LITERAL` had used an `SWR` instruction (unnormalized subtract) rather than `SDR` (normalized subtract) at SRN 00674300 in `PREP_LITERAL`, and then it should work fine and produce exactly the result needed.  Therefore, I'm changing patch59.c on that basis.

These changes are so far local only, and have not been pushed to the repository.  I am checking their effect on the HAL/S compiler first.

rburkey2005:
Those changes didn't immediately produce the results I was hoping for, so I presume there's some other problem I'm not seeing, and will have to resort to actually running HALSFC-PASS1 in gdb to see how `PREP_LITERAL` is going wrong.  (If it is, and the problem isn't elsewhere.)  Some helpful notes to self:

- `PREP_LITERAL` is called only by `SCAN`.
- It is called only after `EXP_OVERFLOW = MONITOR(10, INTERNAL_BCD)`, which stores the HFP value in `DW(0),DW(1)`.
- The string being processed is `INTERNAL_BCD` (or as mangled for debugging, `SCANxINTERNALuBCD`).
- In debugging, the work-area entries `DW(N)` would be found in mangled form as `FORuDW(N).CONST_DW`.

But wait ... if that's true, then `PREP_LITERAL` is only going to be called for a _number_ (like `.425` or `6400`), but not for an expression like `.425 6400`.  So the culprit in our case isn't `PREP_LITERAL` after all.  I reckon it would have to be some other code that also calls `SAVE_LITERAL`, but which prior to doing so performs arithmetical operations on the value to be saved.  The two candidates for that would seem to be `MULTIPLY_SYNTHESIZE` (in our case) and `ADD_AND_SUBTRACT` (in other contexts).  Both use `MONITOR(9,...)` to perform the actual arithmetic, expecting the result of the operation in `DW(0),DW(1)`.

In the debugger, I find that if `DW(0),DW(1)` is loaded with 406CCCCC,CCCCCCCC (.425 to the accuracy of HFP) and `DW(2),DW(3)` is loaded with 441900000,00000000 (6400), then `MONITOR(9,3)` ("multiply") returns `DW(0),DW(1)` of FF000000,00000000 (-0.0).  (Whereas in contrast, loading `DW(0),DW(1)` with .625 instead produces a correct result.)  But `MONITOR(9,3)` merely calls `ibm_dp_sub`, which produces the correct answer in the context when we do something like
<pre>
ibmFloat .425*6400
</pre>

The answer is found in the following new source code added by this PR, for `MONITOR9` in the XCOM-I runtime library runtimeC.c:
<pre>
  uint64_t op0, op1;
  ...
  op0 = ((uint64_t)getFIXED(address) << 32) | getFIXED(address + 4);
  if (op < 6)
    op1 = ((uint64_t)getFIXED(address + 8) << 32) | getFIXED(address + 12);
</pre>
What this code is doing is to construct 64-bit values from the 32-bit values `DW(0),DW(1)` for `op0`, and `DW(2),DW(3)` for `op1`.  When multiplying .425 and 6400, `DW(0),DW(1)` is 406CCCCC,CCCCCC, and `op0` is computed as FFFFFFFFCCCCCCCC.  It is as if `DW(1)` is being _sign-extended_ to 64 bits before performing the bitwise-OR operation.  

The reason this occurs would seem to be that `getFIXED()` is declared as `int32_t` rather than `uint32_t`, and hence in extending it to 64 bits, it's naturally sign-extended.

**Fix**:  The calls to `getFIXED` in the code mentioned above are now qualified with `(uint32_t)`.

rburkey2005:
Post-merge observation:  As [this comment says](https://github.com/virtualagc/virtualagc/pull/1304#issuecomment-4435732037), I changed one `SDR` instruction to `SWR` in the patch59.c file for `PREP_LITERAL`.  I realize now, though, that my comments about the `CC` register for `SDR` were misguided, and the use of `SWR` was probably unnecessary.

Why?  Well, `PREP_LITERAL` only processes actual literals, and not the results of computations.  All legal HAL/S integers can be exactly expressed as DP HFP.  Therefore, the `SDR` in `PREP_LITERAL` is perfectly capable of dealing with them, because they'll all be exact.  They'll only be non-exact if their original strings were inexact, such as `DECLARE N INTEGER INITIAL(1.9999999999)`.  Silly, yes, but if the user explicitly did that, they must have done it for a reason.  The results of compile-time _calculations_ may be inexact integers, but they're processed by subroutines other than `PREP_LITERAL`, by other means.

My excuse is that at the time I originally commented, I was still under the impression that the results of compile-time calculations like `... INITIAL(.425 6400)` were passing though `PREP_LITERAL`.

rburkey2005:
Post-merge comment:  [My iron-clad conclusion earlier](https://github.com/virtualagc/virtualagc/pull/1304#issuecomment-4430722223) that the `AW` BAL instruction needs to round rather than truncate was bogus.  My conclusion was based on how the `PREP_LITERAL` function works and on what I thought it needed to achieve.  What I didn't grasp was that `PREP_LITERAL` is called only for _actual_ literals (as opposed, say, to "literals" created by compile-time evaluation of expressions), and further, that all integer literals can be _exactly_ expressed in HFP without truncation or rounding.  That being the case, there's no need for rounding in `PREP_LITERAL`.  There are at least two other functions with essentially the same logic, namely `MAKE_FIXED_LIT` and `ROUND_SCALAR` which do need to round the of `AW`, but they explicitly handle it by adding 0.5 prior to the `AW`.

So, to sum up:  `ibm_dp_addsub` in ibmFloat.c and .py _should not_ round the result when performing unnormalized additions or subtractions, and I'm reverting that change.

PR #1322, opened 2026-05-22
yaShuttle: HAL/S examples: changes for testability

ColanderCombo:
Some additions to the HAL/S example programs in `yaShuttle/Source Code/Programming in HAL-S/` that makes them more useful as test cases:
  - Add driver programs to call bare functions
  - Add suitable test inputs where there weren't any supplied
  - Write results to default output
  - Implement external COMPOOL and function for `176-P.hal`
    - Exercises compool->template functionality and external references
  - Comment out `FILE()` call in `176-P.hal`
    - per USA-003090/p.80 sect.6.4 "File I/O" this is illegal
  - `037-ROOTS.hal`: convert to labeled `EXIT ROOTLOOP`
    - A bare `EXIT` breaks out of the innermost loop only.
    - An`EXIT LABEL` breaks out of all loops up to `LABEL:` and is required here. 

I pull these tests into [nsts-sdl-dps](https://github.com/ColanderCombo/nsts-sdl-dps) and compare against baselines that live here: [baselines](https://github.com/ColanderCombo/nsts-sdl-dps/tree/master/test/baselines)

### DIFF
--- a/yaShuttle/Source Code/Programming in HAL-S/176.1-READ_ACC.hal
+++ b/yaShuttle/Source Code/Programming in HAL-S/176.1-READ_ACC.hal
@@ -1,0 +1,14 @@
+C/ External function for 176-P: simulates an accelerometer reading.
+C/ Returns a SUPER_VECTOR-STRUCTURE with a sample acceleration
+C/ vector, STATUS=TRUE, and a timetag derived from the selector N.
+D INCLUDE TEMPLATE DUMMY_COMPOOL
+
+  READ_ACC:
+  FUNCTION(N) SUPER_VECTOR-STRUCTURE;
+     DECLARE N INTEGER;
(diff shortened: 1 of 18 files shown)


yaShuttle/Source Code/Programming in HAL-S/176.1-READ_ACC.hal
rburkey2005, line 1: @ColanderCombo I think you've left out one final step with 176.0-SUPER_VECTOR.hal and 176.2-READ_ACC.hal.  I'd suggest:

1. Compiling 176.1-READ_ACC.hal with the TEMPLATE parameter.
2. Creating 176.2-TEST_SUPER_VECTOR.hal, that has a `D INCLUDE TEMPLATE READ_ACC` and provides a test `PROGRAM` for `READ_ACC`.

PR #1342, opened 2026-08-04
XCOM-I + patch optimization related fixes

ColanderCombo:
This PR rolls up fixes for four HAL/S-FC optimization/code-gen failure cases I hit while running CON80 builds along with using SDFPKG template loading (which gets much further than textual inclusion!):

- XCOM-I: honor SET_RECORD_WIDTH when indexing BASED RECORD DYNAMIC variables.
- XCOM-I: ADDR(DESCRIPTOR) is freeBase, not the end of XPL-COMMON
- HAL/S-FC PASS2: patch42 must return GR[3], INTEGER_VALUED's result
- XCOM-I: fix INITIAL parser misrouting inner parens into array-dimension parsing

Details are in the commit messages.  "XCOM-I: ADDR(DESCRIPTOR) is freeBase" is a failure case where some source modules will simply fail to compile, the others all produce valid code but miss an optimization.  With these fixes in place the output object are closer to the know "correct" output.

### DIFF
--- a/XCOM-I/DECLARE.py
+++ b/XCOM-I/DECLARE.py
@@ -8,6 +8,7 @@ Purpose:    This is the module of XCOM-I.py which processes XPL
             DECLARE, ARRAY, or BASED statements.
 Reference:  http://www.ibibio.org/apollo/Shuttle.html
 Mods:       2024-03-07 RSB  Began experimenting with this concept.
+            2026-08-03 DAS  Allow '(' in INITIAL expressions.
 '''
 
 import copy
@@ -247,7 +248,7 @@ def DECLARE(pseudoStatement, scope, library, inRecord = False):
                 for token in attributes:
                     if skip > 0:
                         skip -= 1
-                    elif token == "(": # and inFirst:
+                    elif token == "(" and not inInitial:
                         inTop = True
                         topString = ''
                     elif inTop:
--- a/XCOM-I/generateC.py
+++ b/XCOM-I/generateC.py
@@ -15,7 +15,9 @@ Mods:       2024-03-27 RSB  Began.
             2026-02-19 RSB  Switched from left-to-right processing order in
                             in multipl assignments to right-to-left.
             2026-08-03 DAS  use runtime record widths set by SET_RECORD_WIDTH
-                            on RECORD DYNAMIC tables
+                            on RECORD DYNAMIC tables.
+                            Fix "BI002 ... NON COMMON RECORD AT LINK" error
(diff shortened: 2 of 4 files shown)


ColanderCombo:
@rburkey2005 FYI!

rburkey2005:
@ColanderCombo From your comment, I gather that you've been using the SDF-inclusion feature. We've switched to getting OI340600 and OI301700 HAL/S to build completely.  As of this moment, 100% of OI340600 builds (with the proviso that I only have .dfg files for a lot of them, and so they're excluded) but there are a few (<10) OI301700 files we're still working through.  As for the latter, some of the OI301700 source code itself was messed up when I extracted it from the reports it was living in, and I haven't posted the fixes for them yet.  My faithful assistant is now trying to decide if this PR will mess us up in any way, because we've made some changes that look related to me ... and we conclude it's okay, but unfortunately doesn't fix any of the few remaining problems we're currently working.  Our next step will be a complete recompile of both PASS versions, so there's a possibility, however, unlikely, of discovering a conflict at that point.

rburkey2005:
> FYI!

I saw it.  I just wasn't near the computer.

ColanderCombo:
Cool!  Yes, I'm using SDF inclusion, and I am also able to build all of OI340600. (I can build OI301700 but it's not in my primary test loop)  I'm using my con80 build script in this branch to drive the compile:

https://github.com/ColanderCombo/nsts-sdl-dps/tree/dev/squashed-2026-07-04

My working version does have a large number of unpushed fixes to the tools (lnk101, asm101 and all the new ones) that I'm working through to get cleaned up and pushed now.

For dfg files, you should be able to generate the hal using the current DFG version here:
https://github.com/ColanderCombo/nsts-sdl-dps/tree/dev/squashed-2026-07-04/src/dfg

It's still using textual inclusion and some hacky parsing to get COMPOOL member info, but it should mostly work.  It uses sdf in the local version, I'll get all that pushed as soon as I can!

rburkey2005:
We still have stuff to do _sans_ dfg, but I'll undoubtedly need to come back to it real soon now.

rburkey2005:
With that said, I find no way to get it out of the dev/squashed-2026-07-04 branch and to somewhere else on my computer so that I can run it once I've done a `git checkout master`.  I don't want to keep that other branch checked out forever.

ColanderCombo:
You can use git worktrees to keep multiple branches checked out at the same
time: https://git-scm.com/docs/git-worktree

PR #1344, opened 2026-08-12
XCOM-I: fix for EXPON & make writing the COMMON image optional

ColanderCombo:
Fixes for XCOM-I and the HAL/S-FC compiler:

- use the original MONITOR.bal EXPON algorithm in MONITOR(9) op 5
  - it has a special case path for integer exponents that produces slightly different results than libm pow()
  - still falls back to libm for the non-integer case
  - resolves literal diffs in SSMANTMG & GTBUPL

- add --no-common-image option that skips the output and gzip'ing of the COMMON output file
  - no one seems to consume it by default, and the operation is expensive enough that in some cases it take ~40% of the total runtime of the process.
  - defaults to disabled
  - I enable it in my build system and it results in significant reduction in compile times.

### DIFF
--- a/XCOM-I/runtimeC.c
+++ b/XCOM-I/runtimeC.c
@@ -2867,13 +2867,72 @@ MONITOR9(uint32_t op) {
     r = ibm_dp_div(op0, op1);
   }
   else if (op == 5) {
-    /* a**b — fall back to libm. Convert IBM->IEEE, pow, IEEE->IBM. */
-    double a = ibm_dp_to_double(op0 >> 32, (uint32_t)op0);
-    double b = ibm_dp_to_double(op1 >> 32, (uint32_t)op1);
-    if (a == 0 && b == 0)
-      abend("Zero to the power zero");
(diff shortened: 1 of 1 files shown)


rburkey2005:
@ColanderCombo If you're referring to the COMMONx.out.bin.gz, that is not a gzipped form of COMMONx.out &mdash; they are two entirely different things, in spite of the unfortunate similarity of naming &mdash;, and it is used.

rburkey2005:
@ColanderCombo But if you have no need for it, I suppose there's no reason to keep you from disabling it.  In regard to `pow`, I'm surprised you stopped there.  In yaGPC2 (or is it yaHALMAT2, or both) all of the floating point operations have been fixed to be native IBM Hex, including `sqrt` and transcendentals.

ColanderCombo:
@rburkey2005 the pow/EXPON change was all that was required to match the compiled references; everything else (so far) matches.  Figured minimal changes!

PR #1348, opened 2026-09-19
Preliminary support for simplex code in yaASM.py and yaLVDC

thewonderidiot:
@indy91 and I have started work on reimplementing the LVDC preflight program from the available documentation and flowcharts. Code for it lives here for now: https://github.com/thewonderidiot/lvdc_simulation/tree/main/programs/preflight

This program, unlike the flight programs, was fully simplex. To build and run it, therefore, I've started work on supporting simplex software in yaASM.py and yaLVDC.

At the very least, these changes don't break anything -- AS-512, AS-513, AS-206RAM, and the PTC-ADAPT self-test program all assemble with 0 octal mismatches, and the yaLVDC AS-512 simulated repeatable flight produces identical yaLVDC.pio outputs to before. Once loaded, simplex software seems to execute basically correctly in yaLVDC, but we have some more learning to do here as we flesh out the LVDC<->RCA-110 interfaces.


rburkey2005:
You had me at "don't break anything". :smile:

